PR #32, opened 2026-08-01
Deploy health check must verify real HTTPS, not localhost:80

### LINKED ISSUE #25: Deploy health check must verify real HTTPS, not localhost:80
## Parent

#11 — Phase 5: Autonomous Ticket-Loop + Fleet Observability (spec)

## What to build

Make the deploy tell the truth about whether the site is actually reachable.

**This ticket exists because of a real 24-hour outage.** The `DOMAIN` variable was removed from the VPS `.env` during the Doppler migration. Caddy is the one service that does not run under `doppler run` — it reads that file directly — so it fell back to serving `localhost`, and every real hostname got a TLS error. Port 80 still redirected, so browsers failed outright. The deploy's health check probes `http://localhost:80` from inside the VPS, bypassing both TLS and hostname matching, and reported success throughout.

Two fixes: make the check exercise what a browser exercises, and remove the split-brain that let one consumer of `DOMAIN` go blind while the other was fine.

## Acceptance criteria

- [ ] The deploy health check verifies HTTPS on the real domain with real SNI, not `http://localhost:80`
- [ ] A deploy whose TLS front door is broken **fails**, rather than reporting success
- [ ] `DOMAIN` has a single source of truth, so the reverse proxy and the app cannot diverge again
- [ ] Regression test or check: a deploy with `DOMAIN` missing from the proxy's environment is caught
- [ ] The `.env` comment recording why the proxy needs the value there is preserved or superseded by the single-source fix

## Blocked by

None — can start immediately.

lennons301:
Closes #25

## What changed

**Health check** (`.github/workflows/deploy.yml`): the deploy now probes `https://$DOMAIN/` with `curl -sf --max-time 10 --resolve "$DOMAIN:443:127.0.0.1"` — real SNI, real certificate + hostname verification, through this box's actual front door, with DNS out of the loop. The old `http://localhost:80` probe passed for 24 hours while every real hostname failed with a TLS error; this probe fails a deploy whose TLS front door is broken. Window extended 60s → 120s to cover first-time ACME issuance. A broken deploy also dumps a verbose TLS attempt + compose logs.

**Single source of truth for `DOMAIN`**: Doppler `interlude/prd` — the config the app already boots with (`doppler run` in the Dockerfile CMD). The deploy renders a generated `caddy.env` from that same config (via the just-built app image, which ships the Doppler CLI) and Caddy's `env_file` switches from `.env` to `caddy.env` (`required: false`, so local `docker compose up` still falls back to `localhost`, and `docker compose down` can parse the model on a fresh box before the first render). No hand-maintained copy left to drift — and Caddy stops receiving unrelated secrets (GitHub App key, etc.) from `.env`.

**Regression checks** (AC 4):
- deploy aborts if Doppler yields no `DOMAIN` ("refusing to deploy the proxy blind")
- after `up`, the deploy asserts the caddy container's `DOMAIN` **equals** the freshly rendered value (catches dropped env_file wiring, stale renders, or a crash-looping caddy)
- the HTTPS probe catches the end-to-end symptom regardless of cause

**AC 5 (the `.env` comment)**: superseded. The proxy no longer reads `.env`, and the rationale now lives in the `docker-compose.yml` caddy comment, `.env.example`, and CLAUDE.md. The deploy prints a NOTE while a stale `DOMAIN=` remains in the VPS `.env` so it gets cleaned up; I deliberately did **not** auto-edit that file — it's an operator-owned secrets file and a stale line no service reads is safer than a deploy script mutating it.

## Self-review findings I'm not acting on (and why)

- **Deploy-time snapshot vs boot-time read**: the app re-reads Doppler on every container boot; Caddy gets a snapshot rendered at deploy. If `DOMAIN` changed in Doppler and the app restarted *without* a deploy, they'd diverge until the next deploy. Closing that fully means a custom Caddy image running the Doppler CLI in the TLS path — a poor trade for a value that changes approximately never and whose change would warrant a redeploy anyway. The equality check catches any drift at every deploy.
- **Transitional `grep` NOTE in the deploy log**: flagged as a temporary nag — it is, intentionally; it's the AC-5 supersession nudge and can be deleted once the VPS `.env` is cleaned.

## Notes for the reviewer

- `env_file` long form with `required: false` needs Docker Compose ≥ 2.24 (Jan 2024). Worth a one-off `docker compose version` on the VPS before merging; an older Compose would fail loudly at `docker compose down` (parse error), not silently.
- The new probe traverses Caddy **to the app**, so an app-level 5xx on `/` now fails a deploy. The old port-80 probe only ever saw Caddy's HTTP→HTTPS redirect (`curl -f` passes 3xx), so it never actually checked the app. Stricter on purpose.
- Touches `docker-compose.yml` and `.github/workflows/**` — both `human-signoff` gated paths, as expected for this ticket.
- Validated: `pnpm test` 27/27 green; YAML of both edited files parses; embedded shell reviewed for quoting/fail-fast behavior. `pnpm lint` reports 19 pre-existing problems, all in files this PR doesn't touch. Couldn't run `docker compose config` locally (no Docker access in the agent sandbox); first real verification is the next deploy.


lennons301:

| gate | glob | path |
|---|---|---|
| ci-secrets | `.github/workflows/**` | `.github/workflows/deploy.yml` |
| infrastructure | `docker-compose.yml` | `docker-compose.yml` |

lennons301-reviewer (approved):
Reviewed against issue #25 with fresh context. **Approve** (this PR is gated: `human-signoff` is set and auto-merge is off, so a human merges after looking — correctly so, since it touches `.github/workflows/**` (estate gate) and `docker-compose.yml` (repo infrastructure gate)).

## What I verified

**Acceptance criteria — all five met:**

1. *HTTPS with real SNI* — `curl -sf --max-time 10 --resolve "$DOMAIN:443:127.0.0.1" "https://$DOMAIN/"` does real SNI, real certificate + hostname verification, through the actual front door, DNS out of the loop. Correct `--resolve` syntax.
2. *Broken TLS front door fails the deploy* — the exact failure mode from the outage (Caddy serving `localhost`, TLS error on the real hostname) now fails the probe; failure path dumps a verbose TLS attempt + compose logs.
3. *Single source of truth* — Doppler `interlude/prd`, the same config the app already boots with (Dockerfile CMD). Deploy renders `caddy.env` via the just-built app image (which ships the Doppler CLI — verified); Caddy's `env_file` switches from `.env` to the generated file. I confirmed the mechanics end-to-end: `docker compose run --rm --no-deps -T app doppler secrets get` gets `DOPPLER_TOKEN` from the app's `.env`, stdout is clean (no ENTRYPOINT wrapper), command substitution strips the newline. Side benefit: Caddy stops receiving unrelated secrets from `.env`.
4. *Regression checks* — fail-closed on empty Doppler read (works even if `DOPPLER_TOKEN` is missing/invalid); post-`up` equality check reads the env **inside** the caddy container (quoting verified: `sh -c 'echo "$DOMAIN"'` expands in-container), and a crash-looping/absent caddy fails it via `|| true` → empty → mismatch. Ordering is sound: `down` → build → render → `up`, so the container can never be running a stale render.
5. *`.env` comment* — superseded, with the rationale relocated to the compose caddy comment, `.env.example`, and CLAUDE.md, plus a transitional deploy NOTE while the stale line remains. Not auto-editing the operator-owned `.env` is the right call.

**Objective signals** (no CI runs on PRs in this repo, so run locally at PR head c55e33f): `pnpm test` 27/27 green; both edited YAML files parse; `pnpm lint` shows 19 problems, all in TS files this PR does not touch (the diff contains no TS) — pre-existing, matching the PR body. Scope is tight: every touched file traces to an AC.

**Failure-mode walkthrough for the Compose ≥ 2.24 requirement** (`env_file` long syntax): if the VPS Compose is too old, the model-parse error hits at `docker compose down` *before* it stops anything — old containers stay up, and even if the script continued, the empty-`DOMAIN` check aborts. Fails safe either way, site stays up.

**Self-review deferrals** (deploy-time snapshot vs boot-time read; transitional grep NOTE): both reasoned and I agree — the equality check bounds divergence at every deploy, and `DOMAIN` changing warrants a redeploy regardless.

## For the human merger

- Do the one-off `docker compose version` check on the VPS (≥ 2.24) before merging, as the PR asks.
- The implementer couldn't run `docker compose config` in the sandbox; the first post-merge deploy is the real verification. Watch that deploy run.
- Nit, non-blocking: the "failed after 120s" message can understate elapsed time — with `--max-time 10` per attempt the loop's worst case is ~6 min (bounded by `command_timeout: 15m`). Cosmetic only.
- Suggestion, not tied to a standard: the embedded shell in `deploy.yml` is getting long; if it grows again, extracting it to a repo script would make it shellcheck-able.

lennons301:
Pre-merge check the reviewer asked for: VPS `docker compose version` = **v5.1.0** (requirement: ≥ 2.24 for `--no-deps -T` behavior the deploy relies on). ✅ Safe on that axis.

Remaining before/at merge (for the human merger):
- This merge auto-deploys; the first post-merge deploy **is** the real end-to-end verification of the new HTTPS front-door check — watch the Actions run.
- Merge when the box is idle (deploy restarts the orchestrator).

PR #34, opened 2026-08-01
Phase 5: container resource limits and derived slot capacity

### LINKED ISSUE #14: Phase 5: container resource limits and derived slot capacity
## Parent

#11 — Phase 5: Autonomous Ticket-Loop + Fleet Observability (spec)

## What to build

Agents can no longer take the platform down, and the number of them that may run at once follows the hardware rather than a hard-coded guess.

Every agent container runs under hard memory and CPU limits, so an agent that runs away fails its own task instead of OOM-killing the orchestrator and Caddy alongside it. This was originally a Phase 6 hardening item; unattended parallelism makes it a prerequisite.

Slot capacity is derived from what the Docker daemon reports, so resizing the VPS is understood after a reboot with no config change. The check is expressed as a capacity provider — "is a slot available?" — rather than a direct local-Docker query, so Phase 7's on-demand remote compute can answer the same question differently without reworking the scheduler.

## Acceptance criteria

- [ ] Every agent container — interactive, implement, review and triage alike — is created with hard memory and CPU limits
- [ ] Slot count is derived at boot from the daemon's reported CPU and total memory: available memory less an orchestrator reserve, divided by the per-agent allocation, capped by CPU, minimum 1
- [ ] The derivation yields 2 on the current CX22 (4 GB), matching today's practical ceiling
- [ ] An explicit override exists for when the derivation is wrong for a workload
- [ ] An agent that exceeds its memory cap fails its own task; the orchestrator and Caddy keep serving
- [ ] Capacity is consumed through a provider seam, not a direct Docker query at the call site
- [ ] The derivation is unit-tested table-driven, including a resize changing the answer

## Seams under test

`workflow:tdd` — /tdd must not write a test at an unconfirmed seam, and an unattended pass has no one to confirm with. These are the confirmed seams:

- `deriveCapacity(daemonInfo, overrides) -> { slots, perAgentMemory, cpuQuota }` — pure; table-driven over daemon shapes, including a resize changing the answer
- The capacity provider interface (`isSlotAvailable()`) that callers consume instead of querying Docker directly — the seam Phase 7 later swaps

No Docker, GitHub, Discord or database mocking — if a test needs one, it is at the wrong seam.

## Blocked by

None — can start immediately.

lennons301:
Closes #14

## What this does

- **Hard limits on every agent container**: `createWorkspaceContainer` — the single creation site for all agent kinds — now sets `Memory`, `MemorySwap` (= `Memory`, so the cap bites without swapping the box to death) and `NanoCpus` on `HostConfig`. A runaway agent OOMs its own process and fails its task through the existing turn-failure path; the orchestrator and Caddy are outside the cgroup.
- **Slot capacity derived at boot** (`src/lib/orchestrator/capacity.ts`): `deriveCapacity(daemonInfo, overrides) -> { slots, perAgentMemory, cpuQuota }` — pure, `(memTotal − reserve) / perAgent`, capped by CPU, minimum 1, garbage-daemon-shape safe. First `getCapacity()` call queries `docker info`; a VPS resize is picked up on restart with no config change.
- **Provider seam**: queue pickup consumes capacity via `CapacityProvider.isSlotAvailable()` (async so Phase 7's remote compute can answer differently), never a direct Docker query at the call site. Occupancy counts live containers **plus** pickups still provisioning theirs, so back-to-back polls can't overfill the box.
- **Overrides**: `CAPACITY_SLOTS` (explicit slot count) and `AGENT_MEMORY_MB` (per-agent allocation — container cap and slot divisor together). Invalid values are ignored by the validated derivation rather than halting the fleet.
- **Tests**: table-driven at the two seams the ticket confirmed, including a CX32 resize changing the answer. 45 pass; no Docker/GitHub/Discord/DB mocking.

## Deviation from the spec's illustrative numbers

The Phase 5 spec sketches ~1.5 GB reserve / ~1.2 GB per agent. Against the daemon's **reported** total (a "4 GB" box reports ~3.8–3.9 GiB — verified locally: 4,106,444,800 bytes on a 4 GB WSL host), a 1536 MiB reserve derives **1** slot, violating the binding criterion "yields 2 on the current CX22". The reserve default is therefore 1 GiB (orchestrator ≈300 MB + Caddy ≈50 MB + system headroom), keeping 2×1200 MiB agents + reserve inside the reported total. Pinned by the "as the daemon actually reports it" test row.

## Self-review findings (vs origin/main, spec = #14)

Acted on:
- Per-agent memory override added (`AGENT_MEMORY_MB`) — "wrong for a workload" covers agent sizing, not just slot count.
- `CAPACITY_SLOTS` / `AGENT_MEMORY_MB` documented in `.env.example`.

Declined, with reasons:
- *Rename `cpuQuota` → `perAgentNanoCpus`*: the ticket's confirmed seam names the field `cpuQuota`; kept the spec's vocabulary, units documented in JSDoc.
- *Restart accounting hole (in-memory occupancy misses containers surviving a restart)*: doesn't exist under current semantics — `recoverOrphanedTasks()` force-removes every running task's container **before** `startQueue()`, so no container outlives the process to be miscounted. The Phase 5 restart-re-claim ticket changes those semantics and must own the accounting then.
- *Construct the provider in `startQueue` instead of lazily in the poll*: keeps `startQueue` synchronous; the lazy init is awaited in-tick, so no pickup is ever skipped.

Notes for verification on the VPS:
- Boot log now states the derived capacity (`[orchestrator] Capacity: N agent slot(s), …`) — expect **2** on the CX22.
- The memory-cap→task-fails path is structural (hard cap + existing exec-failure handling) and can't be unit-tested without Docker mocking, which the ticket forbids — worth one deliberate OOM on the VPS before ticking that box.

Pre-existing repo lint debt (19 problems in `custom-server.js`, `preview-pane.tsx`, `discord/client.ts`, `output-parser.test.ts`, `turn-manager.ts`) is untouched — none of it is in files this PR changes, and there is no CI lint gate.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,6 +11,11 @@ KEEP_CONTAINERS=false
 MAX_TURNS=50
 MAX_BUDGET_USD=5.00
 
+# Capacity — slots derive at boot from the Docker daemon's CPU/memory;
+# override only when the derivation is wrong for a workload
+# CAPACITY_SLOTS=2
+# AGENT_MEMORY_MB=1200  # per-agent hard memory cap; also the slot divisor
+
 # Deployment (production only)
 # DATABASE_URL=/data/interlude.db  # set in docker-compose.yml environment block; override here only if needed
 # DOMAIN — production's single source of truth is Doppler (interlude/prd): the
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -15,6 +15,8 @@ export interface AppConfig {
   maxTurns: number;
   /** Max budget in USD per task (default: 5.00) */
   maxBudgetUsd: number;
+  /** Explicit agent slot count, overriding the boot-time derivation. Null = derive from the Docker daemon */
+  capacitySlots: number | null;
   /** Domain for subdomain-based preview (e.g. "interludes.co.uk"). Null = path-based fallback */
   domain: string | null;
   /** GitHub App ID (from app settings page) */
@@ -72,6 +74,9 @@ export function getConfig(): AppConfig {
     keepContainers: process.env.KEEP_CONTAINERS === "true",
     maxTurns: parseInt(process.env.MAX_TURNS ?? "50", 10),
     maxBudgetUsd: parseFloat(process.env.MAX_BUDGET_USD ?? "5.00"),
+    capacitySlots: process.env.CAPACITY_SLOTS
+      ? parseInt(process.env.CAPACITY_SLOTS, 10)
+      : null,
     domain: process.env.DOMAIN ?? null,
     githubAppId: process.env.GITHUB_APP_ID ?? null,
     githubAppPrivateKey: process.env.GITHUB_APP_PRIVATE_KEY ?? null,
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -17,6 +17,8 @@ export interface AppConfig {
   maxBudgetUsd: number;
   /** Explicit agent slot count, overriding the boot-time derivation. Null = derive from the Docker daemon */
   capacitySlots: number | null;
+  /** Per-agent memory allocation in MiB (container cap + slot divisor). Null = default */
+  agentMemoryMb: number | null;
   /** Domain for subdomain-based preview (e.g. "interludes.co.uk"). Null = path-based fallback */
   domain: string | null;
   /** GitHub App ID (from app settings page) */
@@ -76,6 +78,9 @@ export function getConfig(): AppConfig {
     maxBudgetUsd: parseFloat(process.env.MAX_BUDGET_USD ?? "5.00"),
     capacitySlots: process.env.CAPACITY_SLOTS
       ? parseInt(process.env.CAPACITY_SLOTS, 10)
+      : null,
+    agentMemoryMb: process.env.AGENT_MEMORY_MB
+      ? parseInt(process.env.AGENT_MEMORY_MB, 10)
       : null,
     domain: process.env.DOMAIN ?? null,
     githubAppId: process.env.GITHUB_APP_ID ?? null,
--- a/src/lib/docker/container-manager.ts
+++ b/src/lib/docker/container-manager.ts
@@ -3,6 +3,7 @@ import { getDocker } from "./client";
 import { getImageName, ensureImage } from "./image-builder";
 import { getConfig, PLATFORM_REPO_URL } from "../config";
 import { getInstallationToken } from "../github/client";
+import { getCapacity } from "../orchestrator/capacity";
 
 /**
  * Bash run at container setup: install an env-based git credential helper
@@ -89,6 +90,8 @@ export async function createWorkspaceContainer(
   // DNS-safe subdomain derived from task ID (last 8 chars of ULID, lowercased)
   const previewSubdomain = `task-${options.taskId.slice(-8).toLowerCase()}`;
 
+  const capacity = await getCapacity();
+
   const container = await docker.createContainer({
     Image: getImageName(),
     name: containerName,
@@ -98,6 +101,11 @@ export async function createWorkspaceContainer(
     HostConfig: {
       NetworkMode: "interlude",
       Binds: binds.length > 0 ? binds : undefined,
+      // Hard caps: a runaway agent fails its own task, never the platform.
+      // MemorySwap = Memory disables swap, so the cap bites immediately.
+      Memory: capacity.perAgentMemory,
+      MemorySwap: capacity.perAgentMemory,
+      NanoCpus: capacity.cpuQuota,
     },
     NetworkingConfig: {
       EndpointsConfig: {
--- a/src/lib/orchestrator/__tests__/capacity.test.ts
+++ b/src/lib/orchestrator/__tests__/capacity.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { deriveCapacity } from "../capacity";
+import { createLocalCapacityProvider, deriveCapacity } from "../capacity";
 
 const MiB = 1024 * 1024;
 const GiB = 1024 * MiB;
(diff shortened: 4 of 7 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
Reviewed with fresh context against issue #14 and repo standards. Approving; this PR is gated (`human-signoff`), so a human merges after this.

## Objective signals

No CI checks report on this branch, so I ran everything locally at the PR head (1030a90):
- `pnpm test`: 45/45 pass
- `npx tsc --noEmit`: clean
- `eslint` on the changed files: clean. The repo's 19 pre-existing lint problems are all in files this PR does not touch, matching the PR body's claim.

## Acceptance criteria

- **Hard limits on every agent kind** — verified `createWorkspaceContainer` is the *only* `docker.createContainer` call site in `src/` (sole caller: `turn-manager.startTask`), so `Memory`/`MemorySwap`/`NanoCpus` on `HostConfig` covers interactive today and implement/review/triage structurally when those kinds land. `MemorySwap = Memory` correctly disables swap so the cap bites.
- **Derived at boot** — `getCapacity()` runs in `initOrchestrator` before `startQueue`, caches, logs the result; formula matches the spec (`floor((memTotal − reserve)/perAgent)`, CPU-capped, min 1) and is garbage-safe (NaN/0 daemon shapes → 1, tested).
- **Yields 2 on CX22** — tested for both nominal 4 GiB and the daemon-reported ~3.7 GiB shape. The 1 GiB reserve (vs the spec's illustrative ~1.5 GB) is a justified deviation: I checked the arithmetic — with a 1536 MiB reserve, 2 slots needs a reported total ≥ ~3.84 GiB, which a real 4 GB box does not reliably report; with 1 GiB it needs ≥ ~3.34 GiB, comfortably cleared. The binding criterion wins and the test row pins it.
- **Overrides** — `CAPACITY_SLOTS` and `AGENT_MEMORY_MB`, both validated so garbage is ignored rather than halting the fleet (tested), both documented in `.env.example`.
- **Provider seam** — queue pickup consumes `CapacityProvider.isSlotAvailable()` (async, ready for Phase 7); no Docker query at the call site. Occupancy = activeTasks ∪ processingTasks with no double-count (checked `occupiedSlots` against how turn-manager registers/removes entries, including the failure path at `startTask`'s catch).
- **Table-driven derivation tests incl. resize** — CX32 row changes the answer; CPU-cap and memory-only-resize rows too. Tests sit exactly at the two confirmed seams; no Docker/GitHub/Discord/DB mocking anywhere in them, per the ticket's `workflow:tdd` constraint.
- **OOM fails the task, platform survives** — structural (hard cap + existing exec-failure path marks the task failed and removes the container); correctly not unit-tested since that would need Docker mocking the ticket forbids.

## Notes for the human merging

1. **One deliberate OOM on the VPS** is worth doing before ticking the "agent exceeds cap fails its own task" box — the PR body says the same. Also confirm the boot log reports 2 slots on the CX22.
2. `MemorySwap` requires swap accounting enabled in the kernel cgroup config; if the VPS kernel lacks `swapaccount=1` Docker ignores it with a warning while `Memory` still applies. Worth a glance at the daemon log on first deploy.
3. Suggestion only (no standard to tie it to): `src/lib/docker/container-manager.ts` now imports from `src/lib/orchestrator/capacity.ts` while capacity imports `docker/client` — no import cycle, but the limits could be passed into `createWorkspaceContainer` as options if the layering ever needs cleaning up.
4. `cpuQuota` holding NanoCpus is a slightly misleading name, but the ticket's confirmed seam names the field, the self-review declined the rename with that reason, and the units are documented in JSDoc — the ticket overrides the smell.

The self-review section in the PR body is honest: acted-on and declined findings both check out against the code (in particular, `recoverOrphanedTasks` does force-remove running-task containers before `startQueue`, so the restart accounting hole genuinely belongs to the restart-re-claim ticket).

lennons301:
Pre-merge check from the review: VPS swap accounting is fine — `docker info` shows no swap-limit warning, and the box runs kernel 6.8 with cgroup v2 (swap accounting default-on; `swapaccount=1` is cgroup-v1-era). `MemorySwap` will be enforced. ✅

The reviewer's other two asks are post-merge by nature; I'll run them right after this lands:
- watch the boot log for **2 slots** derived on the CX22
- a deliberate OOM task to prove the cap-fails-task path (will coordinate so it doesn't collide with real runs)

Merge notes: auto-deploys + restarts orchestrator; merge when the box is idle.

lennons301:
Post-merge verifications from the review — both done:

**Boot-log slot derivation ✅** — VPS orchestrator logged `Capacity: 2 agent slot(s), 1200 MiB + 1 CPU per agent` on the post-merge restart. The CX22 answer, derived from the live daemon.

**Deliberate OOM ✅** — dispatched a diagnostic task through the real API→queue→`createWorkspaceContainer` path, then drove a page-touching memory hog inside the container's cgroup:
- `docker inspect`: `Memory=1258291200 MemorySwap=1258291200 NanoCpus=1000000000` on the live container — limits applied via the orchestrator path
- `memory.peak` stopped at **1258291200 — the cap to the exact byte**; `memory.events` shows `oom 3, oom_kill 2`; hog SIGKILLed; container `OOMKilled=true`
- Note the kernel killed the offending *process*, not PID 1 — a runaway subprocess dies without failing the task; whole-container death (the task-fails path) only occurs if the main process balloons
- Gotcha for future testers: `Buffer.alloc` without touching pages never becomes resident (782 GB "allocated" at 653 MB RSS) — fill the buffers

**Found while testing, needs owner attention:** the VPS agent's Claude OAuth session has **expired and cannot refresh** (`/home/deploy/.claude/.credentials.json`) — every agent task on the VPS currently fails at auth. Predates this PR; would have blocked the Phase 5 pilot on its first run.

PR #66, opened 2026-08-03
Fleet dashboard: terminal-status tasks can render as running sessions (stale container_status)

### LINKED ISSUE #46: Fleet dashboard: terminal-status tasks can render as running sessions (stale container_status)
Found by the owner on the dashboard's first production viewing (2026-08-02): a task **cancelled on 2026-03-18** appeared as a running interactive session — 136 days stale.

**Root cause, verified on the prod DB:** the task row has `status='cancelled'` but `container_status='idle'` — cancellation never cleared the column. `loadFleetRows` includes any task with non-null `container_status` (src/lib/fleet/rows.ts:50) and the view derives session cards from it without checking the task's own status.

**Fix, both sides (the read model is pure and table-testable — this is exactly its seam):**
1. `buildFleetView`: a task in a terminal status (completed / cancelled / failed) is never an active session, whatever `container_status` says — add the test row
2. Lifecycle: cancel/complete/fail paths clear `container_status` so the data can't lie in the first place

**Interim:** the one stale prod row is being cleared by hand (owner-run UPDATE).

Part of the dashboard practice-review batch (#21's follow-ups).

lennons301:
Closes #46

## What happened

On the dashboard's first production viewing a task **cancelled on 2026-03-18** appeared as a running interactive session — 136 days stale. The task row had `status='cancelled'` but `container_status='idle'`: `buildFleetView`'s occupant filter keyed only off `container_status` and never consulted the task's own status.

## Fix — both sides

**1. Read model (the durable defense, the pure table-testable seam).**
- `buildFleetView` now excludes any task in a terminal status (`completed` / `cancelled` / `failed`) from `occupants`, regardless of `container_status`. This is the single seam feeding slot accounting (`slots.used` / `segments` / `saturated`) and the interactive session cards, and it is shared by the dashboard **and** the daily digest, so both are covered.
- New test asserts a cancelled/completed/failed task carrying a stale `container_status='idle'` counts as no occupant and no running card — it fails without the guard across all three statuses.

**2. Lifecycle — clear `container_status` on terminal transitions.**
- The runtime cancel / complete / fail paths (`cancelTask`, `teardownTaskContainer`, the fail paths in `turn-manager.ts`) **already** null `container_status`, so the class of bug can no longer be produced at runtime; the stale prod row is a March-era legacy artifact.
- The one terminal-status write that didn't clear the column was the restart-recovery sweep cancelling a run's *queued* tasks (`init.ts`). Queued tasks carry no container, so this is a provable no-op today — I made it consistent anyway so every terminal-status write nulls the column, but the real protection here comes from part 1 and the pre-existing runtime paths, not this line.

## Self-review (`/code-review` against `origin/main`)

Ran the code-review skill with issue #46 as the spec. It raised two findings; I acted on the substantive one:
- **`currentTaskOf` (acted on):** the untouched sibling of the fixed filter — it picked a run's "current face" by non-null `container_status` alone, so a finished pass with a stale `container_status` could surface as the run's running-card title / recent-item / blocked-question link instead of the live pass. Same stale-data lie on a different surface. Applied the terminal guard there too, with a test that a live review pass wins over an earlier completed pass holding `container_status='idle'`.
- **`init.ts` no-op (acknowledged, not removed):** the review correctly noted the added `containerStatus: null` is a no-op for queued tasks. Kept as a stated uniform invariant (every terminal-status write nulls the column) and called out honestly above — the runtime paths are what actually clear the column.

## Validation

- `pnpm test` — 365 passed (2 new fleet-view tests)
- `pnpm lint` — clean on all changed files (main carries pre-existing lint errors unrelated to this change)

### DIFF
--- a/src/lib/fleet/__tests__/fleet-view.test.ts
+++ b/src/lib/fleet/__tests__/fleet-view.test.ts
@@ -710,6 +710,44 @@ describe("buildFleetView — running", () => {
     ]);
   });
 
+  it("picks a run's live pass as its face, ignoring a finished pass with a stale container_status", () => {
+    // Issue #46, sibling path: currentTaskOf must not surface a terminal task
+    // as a run's current face just because it still carries container_status.
+    const view = buildFleetView(
+      baseRows({
+        projects: [makeProject({ id: "p1", name: "lemons" })],
+        runs: [makeRun({ id: "r1", projectId: "p1", status: "reviewing" })],
+        tasks: [
+          // Ordered first, so a naive `.find` would return it.
+          makeTask({
+            id: "t-stale",
+            projectId: "p1",
+            runId: "r1",
+            kind: "implement",
+            title: "Old finished pass",
+            status: "completed",
+            containerStatus: "idle",
+            createdAt: new Date(2026, 7, 1, 8, 0, 0),
+          }),
+          makeTask({
+            id: "t-live",
+            projectId: "p1",
+            runId: "r1",
+            kind: "review",
+            title: "Live review pass",
+            status: "running",
+            containerStatus: "running",
+            createdAt: TODAY_9AM,
+          }),
+        ],
+      })
+    );
+
+    expect(view.running).toHaveLength(1);
+    expect(view.running[0].taskId).toBe("t-live");
+    expect(view.running[0].title).toBe("Live review pass");
+  });
+
   it("excludes finished runs and containerless interactive tasks from running", () => {
     const view = buildFleetView(
       baseRows({
--- a/src/lib/fleet/__tests__/fleet-view.test.ts
+++ b/src/lib/fleet/__tests__/fleet-view.test.ts
@@ -172,6 +172,29 @@ describe("buildFleetView — slots", () => {
     expect(view.slots.used).toBe(0);
   });
 
+  it("never counts a terminal-status task as a slot occupant, even with a stale container_status", () => {
+    // Issue #46: a task cancelled months ago still carried
+    // container_status='idle' and rendered as a running interactive session.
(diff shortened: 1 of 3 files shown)


lennons301-reviewer (approved):
Approving — verified against issue #46.

**Ticket coverage (both parts):**
1. Read-model guard — `buildFleetView` now excludes terminal-status tasks (completed/failed/cancelled) from `occupants` regardless of `container_status`. This is the single seam feeding slot accounting, segments, saturation, and the interactive session cards, shared by dashboard and digest. The new table test drives all three terminal statuses against a stale `container_status='idle'` and genuinely fails without the guard (kind=interactive would otherwise pass the existing idle filter).
2. Lifecycle — confirmed the runtime cancel/complete/fail paths in turn-manager.ts (565, 620, 715, 819, 941) already null `container_status`; the init.ts queued-task cancel was the one terminal write that didn't, now made consistent. A provable no-op for queued tasks (no container), correctly called out as such in the PR body rather than oversold.

**Sibling fix (in-scope):** extending the terminal guard to `currentTaskOf` addresses the same stale-data class on the run's-face path (running card title, recent item, blocked-question link) — same root cause, small, and covered by a test where a live review pass wins over an earlier completed pass holding a stale container status. No regression: for terminal runs where all tasks are now terminal, `currentTaskOf` falls back to the most-recent-created task, so recent-item titles still resolve.

**Signals:** full suite 365/365 green, lint clean on all changed files. No false exclusion risk — genuinely live work carries status running/blocked, never terminal, so no active session is hidden.

Minimal, defensive, fully table-tested, and consistent with the pure read-model conventions of src/lib/fleet. Safe to land on main.

Issue #17, opened 2026-07-30
Phase 5: review pass, structured verdict and orchestrator-posted review


lennons301:
## Parent

#11 — Phase 5: Autonomous Ticket-Loop + Fleet Observability (spec)

## What to build

**The ticket that makes work land unattended.** After the implement pass opens its PR, a review pass reads it with fresh context and a fresh clone, using the canonical `ticket-reviewer` definition, and returns a verdict. On an armed PR an approval lands it.

The reviewer's GitHub credential never enters a container. The pass returns a structured verdict and the **orchestrator** posts the review, so approval is an act of trusted code reading a parsed result rather than of a model holding the identity that can approve its own project's work. An unparseable verdict is an error, not a maybe: no merge, tell the owner.

The reviewer has three verdicts, not two. It may approve, request changes, or escalate to human sign-off — it can always add human oversight, and can never remove it.

## Acceptance criteria

- [ ] The review pass runs in a separate container with a fresh clone and fresh context, and receives no credential beyond the ordinary short-lived App token used for cloning
- [ ] The reviewer PAT never enters any container; the orchestrator posts the review
- [ ] Verdicts are approve, request-changes or escalate; an unparseable verdict blocks the merge and notifies
- [ ] Approve on an armed PR lands it; escalate disarms auto-merge, applies `human-signoff` and posts a comment-type review
- [ ] Request-changes is delivered to the still-live implement container as a follow-up turn within the same attempt, rather than starting a fresh attempt
- [ ] The review pass carries its own budget, separate from the implement allowance
- [ ] Reviewing is a separately queued unit of work drawing a slot like any other pass
- [ ] **Prerequisite:** `REVIEWER_GH_TOKEN` is mirrored into the `interlude/prd` Doppler config (a service token is scoped to one config, so the orchestrator cannot read `platform/prd`), and the rotation note — updating both places — is recorded where the next person will find it

## Seams under test

`workflow:tdd` — /tdd must not write a test at an unconfirmed seam, and an unattended pass has no one to confirm with. These are the confirmed seams:

- The verdict parser: reviewer pass output -> approve | request-changes | escalate | unparseable. NDJSON fixtures, following `output-parser.test.ts` and its `stream-fixture.ndjson`
- `decideNext`'s verdict-to-action mapping — unparseable must never yield `armAutoMerge`

No Docker, GitHub, Discord or database mocking — if a test needs one, it is at the wrong seam.

## Blocked by

[...]

PR #75, opened 2026-08-04
Autonomous retry collides with the previous attempt's remote branch (non-fast-forward) — adopt agent/issue-N if it exists

### LINKED ISSUE #72: Autonomous retry collides with the previous attempt's remote branch (non-fast-forward) — fresh containers must adopt agent/issue-N if it exists
Observed on platform#8 (2026-08-04, first pilot runs): attempt 1 pushed `agent/issue-8` and opened PR #17 before dying at the turn limit. Attempts 2 and 3 each started a fresh container, created a new local `agent/issue-8` from the default branch, redid the work, and had their push rejected **non-fast-forward** — the remote branch from attempt 1 already existed and diverged. Both attempts were doomed before they started; $4.14 spent to fail twice. Ticket exhausted to `ready-for-human`.

**Root cause:** container workspace setup always creates the issue branch fresh from the default branch. The laptop runner never hits this because its persistent worktree naturally continues the same branch across attempts.

**Fix:** when the workspace clones, check for an existing remote `agent/issue-<n>`; if present, check it out as the base (continuing the previous attempt's work — matching the laptop worktree semantics, and PR-preserving since the PR points at that branch). A retry should *continue* the branch, not race it.

**Sibling observations from the same strikeout:**
- `MAX_TURNS` default (50) is too low for investigation-heavy tickets — attempt 1 was working honestly when it hit the ceiling. Per-attempt turn/budget settings deserve deliberate values in Doppler `prd` (spec envisaged $20/attempt).
- The turn-limited attempt left a live PR (#17, partial work) — the exhaustion comment on the issue doesn't mention it; a human triaging the exhausted ticket should know a PR exists.

lennons301:
Closes #72

## Problem

On platform#8's first pilot runs, attempt 1 pushed `agent/issue-8` and opened PR #17 before dying at the turn limit. Attempts 2 and 3 each started a fresh container, created a new local `agent/issue-8` from the default branch, redid the work, and had their push **rejected non-fast-forward** — the remote branch from attempt 1 already existed and had diverged. Both retries were doomed before they started.

Root cause: container workspace setup always created the issue branch fresh from the default. The laptop runner never hits this because its persistent worktree naturally continues the same branch across attempts.

## Fix

**1. Adopt the branch (the ticket's core ask).** `buildSetupScript` gains a tri-state checkout mode:

- `create` — branch fresh off the default (triage/interactive, whose per-task branch never exists remotely). Unchanged behaviour.
- `existing` — check out a branch that must already exist (review/repair on the PR branch). Unchanged behaviour (was `checkoutExisting: true`).
- `adopt` — **new**, used by implement passes: check out `agent/issue-<n>` if a previous attempt already pushed it (continuing that branch, PR-preserving), else create it fresh. A full clone fetches the remote-tracking ref for every branch, so `git rev-parse --verify` gates and plain `git checkout` DWIMs a local tracking branch.

The implement prompt now tells the agent to build on any commits an earlier attempt left rather than restarting.

**2. Re-link the existing PR (necessary to actually preserve it).** Adopting the branch makes the retry's push *succeed*, but then `createDraftPr` hit GitHub's 422 "a pull request already exists for this head", swallowed it, and returned null — so the run stayed unlinked, never marked the PR ready, and never routed it to review. That's the same stuck end-state, reached via a successful push instead of a rejected one. `createDraftPr` now falls back to adopting the existing open PR for the head (`findOpenPrForHead`); the caller re-links it, `finishImplementPass` marks it ready and routes it to review, and the duplicate "Draft PR opened" issue comment is suppressed on adoption.

## Tests

- `buildSetupScript` renders the adopt sequence (rev-parse gate + both checkout forms); `create`/`existing` modes preserved.
- `createDraftPr` creates fresh normally, **adopts** the existing open PR when create is rejected, and returns null only when create fails *and* no open PR exists; `findOpenPrForHead` covered directly.
- Full suite: 382 passing. Lint clean on all changed files.

## Self-review (`/code-review` vs `origin/main`)

Ran the review skill; acted on its main finding (the orphaned-PR gap above — part 2 of this PR). Two findings I did **not** action, deliberately:

- *adopt keys on "branch exists" not "this run pushed it"*: a reopened+re-armed ticket whose old branch was never deleted could adopt a stale branch. This is exactly the semantics the ticket specifies ("check for an existing remote `agent/issue-<n>`; if present, check it out"). The estate deletes branches on merge and re-arming is a deliberate human action, so I followed the spec rather than adding run-ownership tracking of branches (a larger design change, out of scope).
- *no end-to-end turn-manager test*: the `isImplementPass → adopt` mapping is a one-line ternary and the repo doesn't mock the full `startTask` flow; I covered the two units that carry the logic (setup-script rendering and PR adoption) instead.

## Out of scope (sibling observations noted on the ticket)

The ticket flags two related-but-separate items from the same strikeout, left for their own tickets: the `MAX_TURNS` default (50) being too low for investigation-heavy tickets, and the exhaustion issue-comment not mentioning a live partial PR. Neither is part of the branch-collision fix.

### DIFF
--- a/src/lib/docker/__tests__/container-manager.test.ts
+++ b/src/lib/docker/__tests__/container-manager.test.ts
@@ -63,13 +63,32 @@ describe("buildSetupScript", () => {
     expect(script).toContain('if [ -n "$DOPPLER_TOKEN" ]');
   });
 
+  it("creates a fresh branch by default", () => {
+    expect(script).toContain('git checkout -b "$GIT_BRANCH"');
+    expect(script).not.toContain("rev-parse");
+  });
+
   it("checks out the existing remote branch for a review pass", () => {
     const reviewScript = buildSetupScript(
       "https://github.com/lennons301/platform.git",
-      true
+      "existing"
     );
     expect(reviewScript).toContain('git checkout "$GIT_BRANCH"');
-    expect(reviewScript).not.toContain('git checkout -b');
+    expect(reviewScript).not.toContain("git checkout -b");
+  });
+
+  it("adopts an existing agent/issue branch on retry, else creates it (issue #72)", () => {
+    const adoptScript = buildSetupScript(
+      "https://github.com/lennons301/platform.git",
+      "adopt"
+    );
+    // Continue a previous attempt's remote branch when it exists...
+    expect(adoptScript).toContain(
+      'git rev-parse --verify --quiet "refs/remotes/origin/$GIT_BRANCH"'
+    );
+    expect(adoptScript).toContain('git checkout "$GIT_BRANCH"');
+    // ...otherwise branch fresh (first attempt).
+    expect(adoptScript).toContain('git checkout -b "$GIT_BRANCH"');
   });
 });
 
--- a/src/lib/docker/container-manager.ts
+++ b/src/lib/docker/container-manager.ts
@@ -6,15 +6,44 @@ import { getInstallationToken } from "../github/client";
 import { getCapacity } from "../orchestrator/capacity";
 
 /**
+ * How setup gets onto `$GIT_BRANCH`:
+ *  - `create`  — make a fresh branch off the default (triage/interactive use a
+ *    unique per-task name that never exists on the remote).
+ *  - `existing` — check out a branch that must already exist on the remote (a
+ *    review or repair pass inspecting the PR branch).
+ *  - `adopt` — continue the branch if a previous attempt already pushed it,
+ *    else create it fresh. An autonomous retry runs in a brand-new container,
+ *    so without this it would branch off the default, redo the work, and have
+ *    its push rejected non-fast-forward against attempt 1's remote branch
+ *    (issue #72). Adopting continues the same branch — matching the laptop
+ *    runner's persistent worktree, and preserving the PR that points at it.
+ */
+export type BranchCheckoutMode = "create" | "existing" | "adopt";
+
+function checkoutCommand(mode: BranchCheckoutMode): string {
+  switch (mode) {
+    case "existing":
+      return 'git checkout "$GIT_BRANCH"';
+    case "adopt":
+      // `git rev-parse` resolves the remote-tracking ref that a full clone
+      // fetches for every remote branch; when it exists, plain `git checkout`
+      // DWIMs a local branch tracking it, otherwise branch off the default.
+      return 'if git rev-parse --verify --quiet "refs/remotes/origin/$GIT_BRANCH" >/dev/null; then git checkout "$GIT_BRANCH"; else git checkout -b "$GIT_BRANCH"; fi';
+    case "create":
+      return 'git checkout -b "$GIT_BRANCH"';
+  }
+}
+
+/**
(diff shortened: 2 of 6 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
Approving. GATED (human-signoff correctly applied — the diff touches src/lib/docker/** [agent-sandbox] and src/lib/orchestrator/autonomy/** [autonomy-control]); this approval only satisfies branch protection, a human merges.

## What I verified

**Objective signals:** No CI configured on the branch, so ran locally — full suite 382 passing (matches the PR claim), and eslint clean (exit 0) on all six changed files. Note the repo's known lint baseline (16 pre-existing errors on main) does not touch these files.

**Against the ticket (#72):** The core ask is done exactly — container setup now adopts a pre-existing remote agent/issue-<n> instead of always branching fresh off the default. The tri-state BranchCheckoutMode ('create'/'existing'/'adopt') is mapped correctly in turn-manager.ts:168 (review/repair → existing, implement → adopt, triage/interactive → create), so a retry continues the branch rather than racing it non-fast-forward.

The adopt shell command is correct: the main repo is a full clone (not --depth 1), so refs/remotes/origin/<branch> exists for every remote branch; 'git rev-parse --verify --quiet ... >/dev/null' gates and plain 'git checkout' DWIMs a tracking branch, else 'git checkout -b'. The compound if…fi sits cleanly in the && chain and fails fast if a checkout errors.

**Part 2 (PR re-link) is necessary, not scope creep:** adopting the branch makes the retry's push succeed, at which point createDraftPr hits GitHub's 422 'a PR already exists for this head'. The new findOpenPrForHead fallback (owner-qualified head, per_page 1 — GitHub allows at most one open PR per head) recovers the existing PR, the caller re-links it via updateTask, finishImplementPass marks it ready and routes to review, and the duplicate 'Draft PR opened' issue comment is suppressed on adoption. Without this the run would still dead-end (swallowed 422 → null → unlinked), so 'PR-preserving' from the ticket genuinely required it. The generalised catch (look up an open PR on any create failure, not only 422) is if anything more robust and preserves the null-on-no-PR original behaviour.

**Consistency / smells:** The checkoutExisting→checkout rename is complete — no stale references anywhere in src. Replacing the boolean with a named union is an improvement over primitive-obsession, not a regression. Error handling follows the file's existing graceful-degradation convention (return null/false on API failure). No duplicated logic; the checkout construction is centralised in checkoutCommand.

**Scope:** The two sibling observations in the ticket (MAX_TURNS default too low; exhaustion comment not mentioning a live partial PR) are explicitly framed in the ticket itself as separate items and are correctly deferred out of this PR.

Tests cover the units carrying the logic: setup-script rendering for all three modes, and createDraftPr fresh-create / adopt-on-422 / null-when-no-PR plus findOpenPrForHead directly. The PR's stated judgement to skip a full startTask E2E for the one-line kind→mode ternary is reasonable.

Sound work; deferring to a human for the merge per the gate.

PR #82, opened 2026-08-04
Preflight: App-inaccessible repo reports as 'no branch protection' instead of naming the real problem

### LINKED ISSUE #70: Preflight: App-inaccessible repo reports as 'no branch protection' instead of naming the real problem
Found during the first real pilot enablement (Platform project, 2026-08-04).

**Symptom:** preflight on lennons301/platform reported `no branch protection on the default branch` even after protection was correctly applied. Actual root cause: the GitHub App isn't installed on that repo, so every App-token API call 404s — and `preflight.ts:148` falls back silently to `defaultBranch = "main"` when `repos.get` fails (the repo's real default is `master`), then reports the downstream 404 as a missing-protection failure.

**Fix:** the repo-access check should be first-class and short-circuiting: if `repos.get` fails with 404/403 under the App token, `preflightReason` = "GitHub App cannot access <owner>/<repo> — add it to the App installation" and skip the dependent checks. Never silently assume a default branch.

**Also observed (same session):** collaborator/label check failures cleared correctly once fixed, so the check granularity is right — it's only the access-failure path that misattributes.

Preflight otherwise did its job: fail-closed, nothing claimable, reasons surfaced in the API response and dashboard.

lennons301:
Closes #70

## What was wrong

`computePreflight` collapsed three distinct owner actions into one misleading `no branch protection on the default branch` reason:

1. It fell back to `defaultBranch = "main"` when `repos.get` failed, so a repo whose real default is `master` would 404 the protection probe and read as unprotected.
2. `getBranchProtection`'s **403** `Resource not accessible by integration` (the App missing the `Administration: read` permission — which interlude's App never needed before preflight existed) was flattened into the same `false` as a genuine **404** `Branch not protected`.

Both come together in the pilot symptom: the App *could* see `lennons301/platform` (collaborator/label reads passed), but the branch-protection probe 403'd on a missing permission, and the "main" guess would also have mis-probed had access been the issue.

## The fix (body + the owner's "Diagnosis correction" comment)

- **Never guess the default branch.** `defaultBranch` starts `null` and is only set from a successful `repos.get`; a null branch short-circuits to a named access reason instead of probing a guessed branch.
- **Three-state `BranchProtectionStatus`** (`protected` / `unprotected` / `forbidden`). `probeBranchProtection` maps `403 → forbidden` and `404 → unprotected`, and `evaluatePreflight` names the missing **App permission** distinctly from a missing **protection rule** — three different owner actions, per the correction's (a)/(b)/(c).
- **The access reason names `owner/repo`** so the owner knows which installation to fix.
- `evaluatePreflight` stays pure and table-tested; new cases cover the `forbidden` and access reasons.

## Beyond the ticket (from self-review, disclosed)

Self-review (`/code-review` vs `origin/main`) surfaced that the `repos.get` catch applied the "add it to the App installation" reason to **every** failure, including a transient 500/network blip — telling the owner to fix an installation that isn't broken. That is the same misattribution class the ticket targets, so I fixed it: repo access is a three-state `RepoAccess` (`accessible` / `inaccessible` / `unreachable`) — only a 404/403 is `inaccessible` (the actionable installation fix); anything else is `unreachable` with a retryable reason, mirroring how `probeBranchProtection` already separates its 403/404 from transient faults. Both still fail closed.

## Review findings I considered and did not act on

- **Duplicated fail-closed literals** across the short-circuit returns — pre-existing house style (the old `appInstalled: false` blocks had the same shape); each explicit literal is clearer at its call site than a positional helper. Left as-is.
- **`repo: string` living inside `PreflightChecks`** — it's context for the reason string rather than a check, but keeping it in the struct lets `evaluatePreflight` stay pure and table-testable with a single object. Defensible; left as-is.

## Validation

- `pnpm test` — 399 passed (21 files), including 11 `evaluatePreflight` cases.
- `npx tsc --noEmit` — clean.
- `pnpm lint` on the two changed files — clean.

### DIFF
--- a/src/lib/orchestrator/autonomy/__tests__/preflight.test.ts
+++ b/src/lib/orchestrator/autonomy/__tests__/preflight.test.ts
@@ -4,13 +4,16 @@ import { HUMAN_SIGNOFF_LABEL } from "../gates";
 
 // The check-to-status/reason mapping is the whole safety property here: a repo
 // is only "passing" when every requirement holds, and a failure must name what
-// is missing. computePreflight's GitHub I/O is a thin shell over this; the pure
-// function is what gets exhaustive coverage.
+// is missing — and, per issue #70, name the *right* missing thing: an
+// unreachable repo, a forbidden endpoint, and an unprotected branch are three
+// different owner actions. computePreflight's GitHub I/O is a thin shell over
+// this; the pure function is what gets exhaustive coverage.
 
 const ALL_PASS: PreflightChecks = {
   repoConfigured: true,
-  appInstalled: true,
-  branchProtected: true,
+  repo: "owner/repo",
+  repoAccessible: true,
+  branchProtection: "protected",
   reviewerIsCollaborator: true,
   signoffLabelExists: true,
 };
@@ -25,8 +28,9 @@ describe("evaluatePreflight", () => {
     // is the prerequisite that actually blocks the owner.
     const result = evaluatePreflight({
       repoConfigured: false,
-      appInstalled: false,
-      branchProtected: false,
+      repo: "",
+      repoAccessible: false,
+      branchProtection: "unprotected",
       reviewerIsCollaborator: false,
       signoffLabelExists: false,
     });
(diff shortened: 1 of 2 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/preflight.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/preflight.ts` |

lennons301-reviewer (approved):
Approving (gated — human-signoff label present, auto-merge off; this approval satisfies branch protection and a human performs the merge).

## Verified

**Objective signals** (no CI on branch, ran locally):
- `pnpm test` — 399/399 pass, incl. 11 `evaluatePreflight` cases.
- ESLint on both changed files — clean.
- `tsc --noEmit` — one error, but in `src/lib/docker/__tests__/container-manager.test.ts`, which this PR does not touch and is identical on origin/main. Pre-existing, unrelated to this change.

**Gating** — correct: the diff touches `src/lib/orchestrator/autonomy/**` (autonomy-control gate), so human-signoff + no auto-merge is exactly right.

**Against the ticket + the owner's diagnosis-correction comment:**
- Never guesses the default branch — `defaultBranch` starts null, set only from a successful `repos.get`; null short-circuits to a named access reason instead of probing a guessed "main". Fixes the root cause at the old preflight.ts:148.
- Repo access is first-class and short-circuiting, and the reason names owner/repo.
- The correction's three distinct owner actions all land: `probeBranchProtection` maps 403 -> forbidden (names the missing "Administration: read" permission) and 404 -> unprotected (names the missing rule), kept separate from the repo-access reason. This correctly resolves the real pilot symptom (repos.get succeeds, getBranchProtection 403s).

**Scope** — the `unreachable` vs `inaccessible` split on repos.get is disclosed in the PR body, is the same misattribution class the ticket targets, is well contained and newly tested. In-spirit, not unrelated creep. Change is fully contained to preflight.ts + its test; no other consumer referenced the removed `appInstalled`/`branchProtected` fields.

**Smell baseline** — clean; booleans replaced by named domain types (RepoAccess / BranchProtectionStatus) is a Primitive-Obsession improvement. `httpStatus` is a legitimate shared helper.

Non-blocking suggestions: probeBranchProtection still fails closed to 'unprotected' on a transient non-403/404 error (now logged + self-heals on the 5-min refresh — reasonable); and the new I/O-mapping branches have no direct unit test, consistent with the file's thin-shell/pure-function testing philosophy.

PR #84, opened 2026-08-04
Rebuild agent image when Dockerfile.agent changes (#78)

### LINKED ISSUE #78: Agent image never rebuilds — ensureImage checks existence, not freshness (Dockerfile.agent changes silently ignored)
Found immediately after #71 merged (2026-08-04): the deploy rebuilt the *app* image, but the *agent* image is built lazily by `ensureImage` (src/lib/docker/image-builder.ts), which returns early if `interlude-agent:latest` exists — so a merged `Dockerfile.agent` change (jq/yq, #71) never reached running agents. Hotfixed with a manual `docker build` on the VPS.

**Fix options:**
1. Deploy workflow rebuilds the agent image when `Dockerfile.agent` changed (cheap, explicit)
2. `ensureImage` tags the image with a content hash of `Dockerfile.agent` and rebuilds on mismatch (self-contained, no CI coupling)

Option 2 is more in keeping with the orchestrator owning its own runtime. Gated path (`src/lib/docker/**`).

lennons301:
Closes #78

## Problem

`ensureImage` (`src/lib/docker/image-builder.ts`) only checked whether `interlude-agent:latest` **existed**, returning early if it did. So a merged `Dockerfile.agent` change — e.g. the jq/yq addition in #71 — never reached running agents until someone rebuilt the image by hand on the VPS.

## Fix (Option 2 from the issue)

The issue offered two fixes and preferred **Option 2** — *"`ensureImage` tags the image with a content hash of `Dockerfile.agent` and rebuilds on mismatch (self-contained, no CI coupling)"* — as "more in keeping with the orchestrator owning its own runtime". This implements exactly that:

- `buildImage` stamps the image with a label (`co.interlude.agent-dockerfile-sha256`) carrying the SHA-256 of the `Dockerfile.agent` it was built from.
- `ensureImage` compares the on-disk Dockerfile hash against the stamped one and rebuilds unless they match — so it also rebuilds when the image is **missing** or **unstamped** (an old image, or the manual `docker build` hotfix). That one-time re-stamp is cheap thanks to Docker's layer cache.
- No deploy-workflow / CI changes — the orchestrator owns its runtime freshness.

Hashing only `Dockerfile.agent` is sufficient because the build context is that file alone (`entries: ["Dockerfile.agent"]`), so the hash captures the entire build input. (Inherent to Option 2: a `node:22-slim` upstream digest change under the same tag won't trigger a rebuild — out of scope for a content-hash-of-the-Dockerfile approach.)

## Tests

New `src/lib/docker/__tests__/image-builder.test.ts` exercises the missing / stale-label / matching / unstamped branches of `ensureImage`, and asserts `buildImage` stamps the current hash. Full suite green (413 tests); lint clean on changed files.

## Notes for the reviewer

- **Gated path.** This touches `src/lib/docker/**`, gated `agent-sandbox` in `docs/agents/review-gates.yaml`, so it must get `human-signoff` and must not be auto-merged.
- **Self-review (rework-avoidance, not the gate).** I ran `/code-review` against `origin/main`. Spec axis: faithful implementation of the preferred option, nothing missing. Standards axis raised two judgement calls: (1) duplicated inspect shape between `imageExists` and `builtDockerfileHash` — **acted on**, extracted a shared `inspectImage()` helper; (2) `Dockerfile.agent` is read/hashed twice on the build path (`ensureImage` then `buildImage`) — **left as-is** on purpose, so exported `buildImage` stays self-contained; the extra read is negligible.
- **Pre-existing, out of scope:** `ensureImage` runs per container-create, so concurrent callers on a stale hash could briefly overlap rebuilds against the same tag. This race predates the change and only appears in the short window after a Dockerfile change before the first rebuild re-stamps.

### DIFF
--- a/src/lib/docker/__tests__/image-builder.test.ts
+++ b/src/lib/docker/__tests__/image-builder.test.ts
@@ -1,0 +1,98 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { createHash } from "node:crypto";
+import { readFileSync } from "node:fs";
+import path from "node:path";
+
+const HASH_LABEL = "co.interlude.agent-dockerfile-sha256";
+
+// The hash the real Dockerfile.agent on disk should produce — computed exactly
+// the way the module under test does, so the "fresh" cases carry a genuinely
+// matching label rather than a hard-coded string that could drift.
+const realHash = createHash("sha256")
+  .update(readFileSync(path.join(process.cwd(), "Dockerfile.agent")))
+  .digest("hex");
+
+// Mutable fake Docker, reconfigured per test via `state.inspect`.
+const { state, buildImageSpy } = vi.hoisted(() => ({
+  state: {
+    inspect: async (): Promise<unknown> => ({ Config: { Labels: {} } }),
+  },
+  buildImageSpy: vi.fn(async (): Promise<unknown> => "BUILD_STREAM"),
+}));
+
+vi.mock("@/lib/docker/client", () => ({
+  getDocker: () => ({
+    getImage: () => ({ inspect: () => state.inspect() }),
+    buildImage: buildImageSpy,
+    modem: {
+      followProgress: (_stream: unknown, onFinished: (err: unknown) => void) =>
+        onFinished(null),
+    },
+  }),
+}));
+
+// Keep the build hermetic: no real tar stream off the filesystem. The Dockerfile
+// is still read for real by node:fs/promises, which is what yields realHash.
+vi.mock("tar-fs", () => ({ pack: () => "TAR_STREAM" }));
+
+import { ensureImage, buildImage, imageExists } from "../image-builder";
+
+const missing = async (): Promise<never> => {
+  throw new Error("no such image: interlude-agent:latest");
+};
+
+beforeEach(() => {
+  buildImageSpy.mockClear();
+  state.inspect = async () => ({ Config: { Labels: {} } });
+});
+
+describe("ensureImage", () => {
+  it("builds when the image is missing", async () => {
+    state.inspect = missing;
+    await ensureImage();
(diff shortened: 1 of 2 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
Approving — sound work, faithful to issue #78. This PR is GATED (human-signoff label, auto-merge off); approval only satisfies branch protection and a human performs the merge.

**Verified**
- Spec: implements the ticket's preferred Option 2 (content-hash stamp + rebuild-on-mismatch) exactly, no scope creep.
- Tests: new file 6/6; full suite 413/413 across 23 files (matches the PR claim). ESLint clean on both changed files. The two tsc errors are in container-manager.test.ts (untouched, byte-identical to main) — pre-existing per the documented baseline.
- Correctness the mocked tests can't reach, checked against installed deps: dockerode 4.0.9 → docker-modem 5.0.6 buildQuerystring JSON-stringifies object options, so `labels` serialises to the JSON-map form the Engine API /build expects and lands in Config.Labels, exactly where inspectImage() reads it back — stamp→read round-trip is sound. Dockerfile.agent has no COPY/ADD and the tar context is that file alone, so hashing it captures the full build input. All four branches (missing / stale / matching / unstamped) route correctly; imageExists() behaviour preserved.

**Smell baseline:** clean — the duplicated inspect shape was extracted to inspectImage(); the deliberate double Dockerfile read is negligible and keeps buildImage self-contained.

**Non-blocking observation (no action required):** deploy.yml still carries `docker rmi interlude-agent:latest` with a 'force rebuild agent image' comment (from fb81f65). It's now redundant belt-and-suspenders — harmless, and correctly left untouched given Option 2's no-CI-coupling stance; a future cleanup could drop it so the comment doesn't imply the deploy is the mechanism.

Issue #74, opened 2026-08-04
Pin and configure the agent model per pass kind (implement / review / triage)


lennons301:
`execClaudeTurn` passes no `--model`, so agent containers run whatever the CLI resolves as the account default — an implicit dependency that can change under us, unrecorded in the run ledger.

**Fix:**
- Config: `AGENT_MODEL` (default for implement) + optional `AGENT_MODEL_REVIEW` / `AGENT_MODEL_TRIAGE` overrides — review and triage passes are strong candidates for a cheaper tier
- Pass `--model` in `execClaudeTurn` from the task's kind
- Record the model used on the run row so spend numbers are interpretable

Raised by the owner during the first pilot runs (2026-08-04).

lennons301:
Armed `ready-for-agent` under Sean's standing authorization (2026-08-02), confirmed 2026-08-04: run the pilot-fix chain through the laptop loop. Human decision, agent applied the label.

lennons301:
🤖 Attempt 1/3 starting.

PR #85, opened 2026-08-04
Ticket directive: per-ticket model override, clamped to an allowlist

### LINKED ISSUE #80: Ticket directive: per-ticket model override, clamped to an allowlist
Owner decision on #74's follow-up (2026-08-04): scoping should be able to choose the model for a ticket — e.g. a mechanical rename doesn't need Opus; a gnarly refactor might warrant it explicitly.

**Design:** add `model:` to `TicketDirectives` (parsed from the Workflow section like `budget:`/`max-turns:`), resolved through the same `resolveAgentModel` seam, **clamped to an allowlist** of permitted values (e.g. `opus`, `sonnet`, `haiku` aliases) — a semi-trusted ticket body must not be able to name arbitrary models, mirroring the $75 budget clamp's reasoning. Unknown value ⇒ ignore directive, use defaults, note it in the run (never fail the run on a bad directive).

Recorded on `runs.model` as with #74.

lennons301:
Closes #80

## What

Adds a per-ticket `model:` directive so scoping can choose the tier a ticket's work runs on — a mechanical rename needn't burn Opus; a gnarly refactor can ask for it explicitly.

- **Parse + clamp (trusted code).** `model:` joins `TicketDirectives`, parsed from the Workflow section exactly like `budget:`/`max-turns:` and clamped to `ALLOWED_TICKET_MODELS` (`opus`/`sonnet`/`haiku`) in `budgets.ts`. A semi-trusted body may only *select a tier*, never name an arbitrary model — mirroring the $75 budget clamp. Only the clamped value ever reaches `--model`.
- **Resolved through the `resolveAgentModel` seam.** It gains an already-clamped `ticketModel` override that wins for the run-carrying kinds (implement, repair, interactive). **Review and triage keep their own cheaper tier regardless** — the ticket chooses the model its *work* runs on, not the reviewer's.
- **Recorded on `runs.model`** (as with #74): pinned at claim time from the directive, then resolved consistently when the implement pass starts.
- **Never fails on a bad directive.** An unrecognised value is ignored (the run keeps its default) and *noted*, not swallowed — a `console.warn` plus a line on the claim comment. The honoured tier is likewise noted on the claim comment.

## Tests

- `ticket.test.ts` — parses/lowercases an allowlisted alias, ignores non-allowlisted/arbitrary model ids and a bare `model:`, first-wins on repeat, code-fence exclusion; plus `rawModelDirective`.
- `config.test.ts` — override applies to implement/repair/interactive, never to review/triage, falls back to base when null.
- `decide.test.ts` — the claim action carries an allowlisted model and drops an unknown one.

Full suite green (421 passed); lint clean on changed files. (The two pre-existing `tsc` errors in `container-manager.test.ts` are on `main`, untouched here.)

## Notes for the reviewer

- **"note it in the run" — interpretation.** There's no free-text notes column on `runs`, and adding one for this edge case felt like more than the ticket asked. I read "the run" as its lifecycle record: an ignored directive is surfaced on the run's issue thread (the claim comment) and in the orchestrator log, while `runs.model` truthfully records the model actually used (the default). Happy to persist it on a column instead if you'd prefer.
- **Minor scope beyond the strict ask** (both benign): the *honoured* case also gets a claim-comment note (operator visibility), and `interactive` shares the override branch — harmless, since interactive tasks have no run, and it preserves #74's kind grouping.
- Self-reviewed via `/code-review` against `origin/main`; the documentation-drift finding is addressed (runbook now documents the directive). The `rawModelDirective`/parser duplication is a deliberate, documented raw-vs-clamped split kept small by sharing the `DIRECTIVE_LINE` regex and `workflowSectionLines`.

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -210,6 +210,17 @@ pass normally but then forces `human-signoff` regardless of gate matches, carryi
 `<text>` as the note for what needs your decision. Use it for
 agent-doable-but-risky work you want to eyeball before merge.
 
+### Model tier (`model:`)
+
+A `model: <alias>` directive in a ticket's Workflow section picks the tier the
+implement (and repair) pass runs on — `opus`, `sonnet` or `haiku`. Use it to
+match spend to the work: a mechanical rename doesn't need Opus; a gnarly
+refactor can ask for it explicitly. The set is a fixed allowlist (issue text is
+semi-trusted, so it may only select a tier, never name an arbitrary model); an
+unrecognised value is ignored — the run keeps its default model and the claim
+comment notes that it was dropped. Review and triage passes keep their own
+(cheaper) tier regardless. The chosen tier is recorded on the run.
+
 ### Capacity / slots
 
 Slots derive at boot from the Docker daemon's CPU and memory
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -49,8 +49,10 @@ export const runs = sqliteTable("runs", {
   // Per-exec turn limit from a ticket's max-turns directive; null = default
   maxTurns: int("max_turns"),
   // Model the implement pass ran on (issue #74), so this run's spend is
-  // interpretable against its tier. Set when the implement pass starts; null
-  // means AGENT_MODEL was unset and the CLI resolved the account default.
+  // interpretable against its tier. A ticket's `model:` directive (issue #80)
+  // pins it from claim time; otherwise it is set when the implement pass
+  // starts. Null means AGENT_MODEL was unset (no directive) and the CLI
+  // resolved the account default.
   model: text("model"),
   totalCostUsd: real("total_cost_usd").notNull().default(0),
   pullRequestNumber: int("pull_request_number"),
--- a/src/lib/__tests__/config.test.ts
+++ b/src/lib/__tests__/config.test.ts
@@ -60,3 +60,33 @@ describe("resolveAgentModel (issue #74)", () => {
     expect(resolveAgentModel("implement", c)).toBeNull();
   });
 });
+
+describe("resolveAgentModel ticket model override (issue #80)", () => {
+  it("lets a ticket model override the base for the work-carrying kinds", () => {
+    const c = cfg({ agentModel: "base-model" });
+    expect(resolveAgentModel("implement", c, "opus")).toBe("opus");
+    expect(resolveAgentModel("repair", c, "opus")).toBe("opus");
+    expect(resolveAgentModel("interactive", c, "opus")).toBe("opus");
+  });
+
+  it("overrides even when no base model is configured", () => {
+    const c = cfg({ agentModel: null });
+    expect(resolveAgentModel("implement", c, "sonnet")).toBe("sonnet");
+  });
+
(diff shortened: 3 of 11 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/ticket.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/ticket.ts` |

lennons301-reviewer (approved):
Approving. Reviewed with fresh context against issue #80. This is a gated PR (human-signoff label, auto-merge off) — this approval satisfies branch protection only; a human still performs the merge.

## Objective signals
- Full suite green locally: 421 passed (22 files).
- Lint clean on all changed source files.
- The only `tsc` errors are the two pre-existing ones in `src/lib/docker/__tests__/container-manager.test.ts`, which this PR does not touch — matches the PR body's claim. No new type errors.

## Against the ticket (#80)
Every clause is implemented, and only what was asked:
- `model:` joins `TicketDirectives`, parsed from the Workflow section exactly like `budget:`/`max-turns:` (same `DIRECTIVE_LINE` shape, code-fence exclusion, first-wins, case-insensitive).
- Clamped to an allowlist (`ALLOWED_TICKET_MODELS` = opus/sonnet/haiku) in trusted code — the parser. Tests confirm `gpt-4`, a full model id, and a bare `model:` are all dropped, so a semi-trusted body can only select a tier. This mirrors the $75 budget clamp's reasoning as the ticket asked.
- Resolved through the `resolveAgentModel` seam via a new already-clamped `ticketModel` override; review and triage correctly keep their own cheaper tier regardless (verified in config.test.ts and by reading the switch).
- Unknown value ⇒ ignored, never fatal, and surfaced (console.warn + claim-comment note) rather than swallowed.
- Recorded on `runs.model`: pinned at claim from the directive, then resolved consistently when the implement pass starts. Verified the write-back path in turn-manager (lines 125/135-147) and that both call sites (125, 512) thread `run?.model`. The column already exists from #74, so no migration is needed — correctly none added.

## Interpretation the human should weigh (not a defect)
"note it in the run": there is no free-text notes column on `runs`, so the implementer surfaces an ignored/honoured directive on the run's issue thread (claim comment) + orchestrator log, with `runs.model` truthfully recording the model actually used. Defensible and clearly disclosed; the implementer offered a column instead if preferred. Flagging so the human merger can decide, no change required.

## Minor, non-blocking
- `rawModelDirective` re-walks the Workflow lines that `parseTicketDirectives` already scans (mild Duplicated Code). It's a deliberate, documented raw-vs-clamped split that shares `workflowSectionLines` and `DIRECTIVE_LINE`, and the residual loop is ~3 lines on a cold path. Acceptable as-is; could later be folded by having the parser also return the raw value.
- The `interactive` kind falls in the override branch, but interactive tasks carry no run so `run?.model` is always null in practice — benign, and it preserves #74's kind grouping.

Clean, tightly-scoped, well-tested, faithful to the ticket and to the established directive pattern.

Issue #95, opened 2026-08-04
Duplicate concurrent review passes queued for one run


lennons301:
## What happened

During the 2026-08-04 OOM incident, run 01KZ6TVVVVJCRSTTBGJ7D7YAR1 (LPS #141, PR #144) had **two review tasks running simultaneously** — 01KZ6W6G580W83EKDKMTWXJ3TA and 01KZ6W7DJFEZPK35BYQKPACPDY, both `kind: review`, both `status: running`, containers started 4 seconds apart the moment slots freed.

## Why (hypothesis)

The sweep's in-flight-review check didn't see the earlier review as live — likely because the first pass's container had been OOM-killed / hung while its task row still said `running`, or because the check keys on container liveness rather than task status (or vice versa). The sweep queued a replacement while the original was still queued/running, and both launched when capacity appeared.

## Impact

Duplicate review spend (~$1–5 each), racing verdicts for the same PR, and extra memory pressure at exactly the moment the box could least afford it (contributing to #93's overcommit).

## Proposed fix

Before queueing a review pass, treat any existing task for the same run with `kind = review` and `status IN (queued, running)` as in-flight — and make the ungraceful-death path (container gone, task still `running`) mark the task terminal first, so the retry is a deliberate replacement rather than a racing duplicate. A test for "review pass dies without status transition → exactly one replacement queued" would pin it.

lennons301:
Armed under Sean's direction (2026-08-05, supervising session): incident defect batch runs via the laptop ticket-loop while VPS autonomy stays off. Sequencing: wave 1 = #93/#94/#96 in parallel worktrees; wave 2 = #95 then #97 (both touch the ungraceful-death seam, run after wave 1 lands).

lennons301:
🤖 Attempt 1/3 starting.

Issue #90, opened 2026-08-04
Fleet dashboard: review/triage passes invisible as activity; unparseable-verdict runs missing from needs-you


lennons301:
Owner observation during the pilot re-test (2026-08-04): while review passes ran for PRs #17/#20, the dashboard gave no indication — 'it looked like the PRs had no review'. Two gaps:

1. **Review (and triage) tasks aren't legible as fleet activity.** The run card's phase pipeline exists (`implement ▸ review ▸ merge`) but an in-flight review pass should read unambiguously as such, with its spend.
2. **A run parked by an unparseable verdict** (`review_result.kind === 'unparseable'`, terminal by design) **isn't a needs-you bucket cause** — it's exactly a 'needs you' state (the design says 'the owner is told'), but #21's bucket list predates #17's verdict machinery, so these runs sit in 'gated' indistinguishable from happily-waiting ones.

Both are `buildFleetView` changes with table-driven tests at the established seam.

lennons301:
Scope addition from the owner (2026-08-04): the needs-you **sign-off bucket is firing prematurely** — it currently flags gated runs whose PRs are *not yet human-mergeable* (PR #17 is CHANGES_REQUESTED mid-cycle; PR #20 has no posted review). Correct semantics: a gated run belongs in the sign-off bucket only once its review verdict is **approve** (i.e., the PR is genuinely one human merge away). Before that it's fleet activity, not a needs-you item — unless it parks via the unparseable path (point 2), which is its own distinct cause.

lennons301:
Armed `ready-for-agent` under Sean's standing authorization — prioritized to run first at his direction (2026-08-04).

lennons301:
🤖 Attempt 1/3 starting.

PR #55, opened 2026-08-02
Phase 5: triage pass for new issues

lennons301:
Closes #23

## What this does

A handwritten issue gets met on arrival. `issues.opened` on a registered project marks it `needs-triage` (the label is the queue) and kicks the sweep; the reconciliation sweep picks up strays carrying the label the same way. A short, read-only pass (`kind: triage`, no run, $2 allowance, hard 15-turn cap, ordinary slot from the shared pool between review and implement priority) reads the issue against the repo's context and returns exactly one of three exits as its final message:

- `TRIAGE: recommend` → assessment comment + Discord recommendation; **no label** — arming stays with a human
- `TRIAGE: needs-info` → `needs-info` label + the specific questions
- `TRIAGE: ready-for-human` → `ready-for-human` label + a suggested grilling agenda

The orchestrator applies the consequences; the pass itself labels, comments, edits and closes nothing. `needs-triage` removal is the acted-on latch, ordered last so every earlier failure is retryable by the next sweep. The parsed exit is stored on the new `tasks.triageResult` column (migration `0012`), so it survives a restart without re-running the pass.

**The ceiling, enforced orchestrator-side and proven by test:** `parseTriageExit`'s vocabulary is exactly three exits — a pass claiming `TRIAGE: ready-for-agent` parses as unparseable — and `decideNext` derives labels from the exit *kind* alone, drawn from `ADVISORY_TRIAGE_LABELS` (which excludes `ready-for-agent`); the executor refuses any `applyTriage` carrying a non-advisory label as defence in depth. The existing arming-boundary stress matrix gains the triage surface plus a composed raw-stream→parse→decide proof, and the action vocabulary cannot express a body edit or an issue close at all. An unparseable exit fails closed: nothing applied, owner told once, one bounded retry (`MAX_TRIAGE_PASSES_PER_ISSUE = 2`), then the issue parks visibly on `needs-triage`.

**Arming route (exit one):** a label click on GitHub works as it always did; a reply of exactly **"yes"** to the Discord recommendation makes the orchestrator comment the route on the issue *first* ("Armed via Discord: <user> confirmed…") and then apply `ready-for-agent` (`armIssueFromDiscord` — deliberately unreachable from `decideNext` or any pass output). Anything short of an explicit yes, including silence, does nothing.

Triage spend joins the daily-cap sum by kind in `spend.ts` (triage owns no run; interactive stays exempt by construction), and `decideNext` skips triage pickup once the cap is reached.

## Judgement calls the ticket didn't settle

[...]

PR #98, opened 2026-08-05
Review sessions in the UI: link the run card to its review task, not the finished implement pass

### LINKED ISSUE #96: Review sessions in the UI: click-through lands on the finished implement task; review task sometimes never appears
## What happened (owner report, 2026-08-04 LPS batch)

Watching the fleet dashboard during the first multi-ticket autonomous run:

1. Review sessions visibly spun up, but clicking on them sometimes navigated to the run's **finished implement task** instead of the review task.
2. Sometimes the review task didn't visibly appear at all, despite a review container running.

## Context that may matter for repro

- Observed while the host was under heavy memory pressure (see #93), when several tasks had stale `running` rows after ungraceful container deaths — possibly related to #66's stale `container_status` class of bug, and to #95's duplicate review passes (a duplicate review may render over/instead of the visible one).
- A run's card links somewhere — if it links to the run's first/primary task rather than the currently-active task, that alone explains symptom 1.

## What good looks like

- A running review session on the dashboard links to the review task's own detail view.
- Every live container corresponds to exactly one visible session in the UI, whatever its kind.

Needs repro + investigation once the batch is done; filing now so the observation isn't lost.

lennons301:
Closes #96

## The bug

Watching the fleet dashboard during the LPS batch, a running **review** session's card sometimes navigated to the run's **finished implement task** instead of the review task (symptom 1), and sometimes the review task didn't visibly appear at all despite a review container running (symptom 2).

Root cause: the Running card linked to `activePassOf(run.id)?.id`, and **neither** `activePassOf` nor `currentTaskOf` selected by `tasks.kind` — even though the orchestrator's own `workingTaskOf` does. During review the implement pass is parked `status:"running"`/`containerStatus:"idle"` **by design** (issue #17) while the review pass runs. Once the review pass reaches a terminal `task.status` (its container torn down) but the run is still `reviewing`/`gated` until the next sweep posts the verdict, `activePassOf` finds no busy pass and falls back to the only live task — the parked implement. So the review card links to the finished implement pass, and the review session has no visible face of its own. Under the slow sweeps of the memory-pressure incident, that window was wide open.

## The fix

Resolve a run's dashboard face **by phase**:

- `phaseKinds(run.status)` names the pass kind for the phase — `review` for `reviewing`/`gated`, `implement`/`repair` otherwise — mirroring the orchestrator's `workingTaskOf`.
- `currentPassOf` picks **within that phase only**: prefer the container actually executing, else the newest (deliberately including a *finished* review pass, so a reviewing card stays on its review pass even after the container is gone). It does **not** fall back across phases — before the review pass is queued the card links nowhere (a non-clickable review card) rather than deep-linking to the parked implement, which would be the exact symptom again.
- Duplicate review passes (#95) resolve to the one really running.

It also unifies the two previously-divergent selectors (`activePassOf`, `currentTaskOf`) into one `currentPassOf`, used by the running card, the blocked deep-link, and the recent ledger. A shared `isBusyPass` predicate is extracted; a boolean `hasBusyPass` gates the gated-in-flight "is this fleet activity" check so #90's between-sweeps semantics are preserved.

## Deliberate scope / consequences (called out in self-review)

- **Recent-ledger title change** (side effect of unifying selectors): a terminal run is now titled by its **implement** pass (the actual work, e.g. "Add auth") rather than whatever pass was newest (often "Review PR #55: …"). This is an improvement and is what the ledger should show; flagging it since it wasn't strictly requested.
- **Invariant is bounded to *executing* containers.** A parked-idle implement is a live container that holds no slot (#17) and intentionally has no separate session card; duplicate review containers (#95, a separate cause listed in the issue) still collapse to one card. Fully honouring "every live container → exactly one card" for those would need the one-card-per-run model to change — out of scope here.

## Tests

Added four regression cases to `fleet-view.test.ts`: a reviewing run links to its review pass after the review container is gone; phase (not stale busy implement) decides the face; duplicate review passes resolve to the running one; and a reviewing run with no review pass yet renders a non-clickable card instead of linking to the parked implement. Full suite: 460 passing; lint and typecheck clean on the changed files (two pre-existing, unrelated `container-manager.test.ts` type errors remain on `main`).

## Self-review

Ran `/code-review` (Standards + Spec) against `origin/main` with #96 as the spec. Acted on all findings I agreed with: extracted the duplicated busy predicate, made `hasBusyPass` a boolean, sharpened the misleading `pickPass` comment, and — the substantive one — removed the cross-phase fallback that reintroduced symptom 1 in the pre-review-pass window.

### DIFF
--- a/src/lib/fleet/__tests__/fleet-view.test.ts
+++ b/src/lib/fleet/__tests__/fleet-view.test.ts
@@ -881,7 +881,7 @@ describe("buildFleetView — running", () => {
   });
 
   it("picks a run's live pass as its face, ignoring a finished pass with a stale container_status", () => {
-    // Issue #46, sibling path: currentTaskOf must not surface a terminal task
+    // Issue #46, sibling path: currentPassOf must not surface a terminal task
     // as a run's current face just because it still carries container_status.
     const view = buildFleetView(
       baseRows({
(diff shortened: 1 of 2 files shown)


lennons301-reviewer (approved):
Approving. Reviewed with fresh context against issue #96 and repo standards.

**What I verified**
- **Objective signals (no PR CI configured, so ran locally):** full suite 460 passing; ESLint clean on both changed files; `tsc --noEmit` reports only the two pre-existing, unrelated errors in `container-manager.test.ts` (a file this PR does not touch) — nothing new in the fleet files. Branch is current with `origin/main`.
- **Ticket fit:** the root cause (the card resolved its face via `activePassOf`/`currentTaskOf`, neither of which selected by `tasks.kind`, so during review it fell back to the parked implement pass) is fixed by resolving the face **by phase** — `phaseKinds(run.status)` mirrors the orchestrator's `workingTaskOf`, and `currentPassOf` picks within that phase only. Symptom 1 (click lands on the implement task) and the duplicate-review case of symptom 2 (#95) are both addressed, with four new regression tests pinning each: reviewing card stays on the review pass after its container is gone; phase beats a stale-busy implement; duplicate reviews resolve to the running one; and the pre-review-pass window renders a non-clickable review card rather than deep-linking to the parked implement (the exact symptom).
- **Scope:** the change stays inside the fleet read model — no rendering/markup/CSS, no autonomy-control reducer, no schema. The two disclosed side effects are sound: the recent-ledger now titles a terminal run by its implement pass ("Add auth") rather than the newest pass ("Review PR #55: …"), which is the right thing to show; and the gated in-flight activity gate now uses `hasBusyPass` (a live, non-idle pass on a slot), which matches the existing comment's stated intent better than the old `activePassOf !== null` and keeps the #90 tests green.
- **Regression safety:** no dangling references to the removed `activePassOf`/`currentTaskOf`; both #90 gated-activity tests still pass because their review pass is `containerStatus: running` (busy). The invariant's remainder ("every live container → exactly one card" for parked-idle implements) is correctly scoped out — a parked implement holds no slot by design (#17) and #95 tracks duplicate review containers separately.
- **Smell baseline:** clean — the PR reduces divergence by unifying two selectors into `currentPassOf` and extracting the shared `isBusyPass` predicate; names are clear; no speculative generality.

Gate check: changed paths (`src/lib/fleet/**`) match no estate or repo human-signoff glob, so the armed state is correct. A defect here would be a mislinked dashboard card on an observability surface — no safety, security, data, or spend exposure — so I'm comfortable with this landing on main unsupervised.

PR #99, opened 2026-08-05
Verdict parser rejects a valid review verdict that isn't the final message's first line

### LINKED ISSUE #94: Verdict parser rejects a valid review verdict that isn't the final message's first line
## What happened

LPS PR #144's review pass (task 01KZ71HS93TJN9CMFCSYD6K88P, 2026-08-04) completed successfully: it ran the repo's tests (661 passing), typecheck, and lint, wrote a full acceptance-criteria table, and approved. Its final message read:

```
Objective signals verified locally:

- **Tests:** `vitest run` — 84 files, 661 tests, all pass ...
- **Typecheck:** `tsc --noEmit` clean ...
- **Lint:** `biome check .` — 0 errors ...

VERDICT: approve

Reviewed `d789646`/`0c39d07` against issue #141 ...
```

The orchestrator discarded it:

```
[orchestrator] Review task 01KZ71HS93TJN9CMFCSYD6K88P verdict: unparseable
[autonomy] Unparseable review verdict for lennons301/last-person-standing#141 PR #144: final message does not start with a VERDICT: line
```

A correct, evidenced approval was thrown away over layout, the run parked in needs-you, and a human had to recover the verdict from the messages table and post it by hand.

## Why

The parser requires the final message to **start with** the `VERDICT:` line. Models routinely preface the verdict with a verification summary — arguably good behavior we shouldn't punish. This will keep biting: it costs a ~$1 review pass plus the bounded format-retry (#89/#92) every time, and two format failures park the run.

## Proposed fix

- Scan the final message for the first line matching `^VERDICT:\s*(approve|request-changes|escalate)` instead of requiring position zero. Anchor to line start to avoid matching quoted text mid-line; take the first match.
- Belt and braces: strengthen the review-pass prompt to demand the verdict line first — but the parser must tolerate the preamble regardless, because the prompt alone demonstrably doesn't hold.

lennons301:
Closes #94

## What was wrong

LPS PR #144's review pass produced a correct, evidenced `VERDICT: approve` — but the model prefaced it with a tests/typecheck/lint verification summary. The parser required the final message to *start with* the `VERDICT:` line, so it returned `unparseable`, the run parked in needs-you, and a human recovered the verdict by hand. A good review was thrown away over layout.

## The fix

Per the ticket's proposed fix:

- **Parser (`verdict.ts`)** — scan the final message for the first line matching `^VERDICT:\s*(approve|request-changes|escalate)` instead of demanding position zero. Line-start anchoring still rejects the marker quoted mid-sentence (`…my conclusion is VERDICT: approve`); the first matching line wins; the body is everything after the verdict line.
- **Prompt (`buildReviewPrompt`)** — belt-and-braces: strengthened to demand the `VERDICT:` line first and steer verification notes into the body. The parser tolerates a preamble regardless, since the ticket notes the prompt alone demonstrably doesn't hold.

## Tests

- New fixtures `verdict-approve-with-preamble.ndjson` (reproduces the incident) and `verdict-request-changes-with-preamble.ndjson`, with tests asserting both parse correctly.
- The mid-line-quoted malformed case stays `unparseable` (test renamed to reflect *why*: it's quoted mid-line, not merely non-first-line).
- Full suite green (458 tests); lint clean on changed files. Two pre-existing `tsc` errors in `container-manager.test.ts` are untouched baseline noise.

## Self-review notes (from `/code-review` against `origin/main`)

Findings I considered and **deliberately did not act on**, with reasoning — flagged here rather than silently ignored:

1. **Divergence with the triage parser (Duplicated Code).** `parseReviewVerdict` now scans; `parseTriageExit` still demands position zero. Sharing a `parseMarkedExit` helper (or making triage scan too) is out of scope for #94 — the preamble incident was reported for the *review* verdict, and changing triage's parse tolerance is its own decision with its own consequences. A shared helper parameterised by scan-vs-position-zero would just abstract over a real behavioural difference. Left as a possible follow-up.
2. **Dropped end-anchor also accepts trailing decoration on the line** (`VERDICT: approve — ship it`). This is faithful to the ticket's literal proposed regex (`^VERDICT:\s*(…)`, no `$`), not scope creep. Kept.
3. **The preamble is dropped, not posted.** Intentional per the ticket ("everything after that line is the body"); the prompt now steers notes into the body, and the empty-body guard + bounded format-retry keep a body-less request-changes/escalate fail-safe (it re-queues, never merges).
4. **First-match could be fooled by a reviewer reciting the format in a quoted example line.** Inherent to the ticket's chosen "take the first match"; the prompt discourages it. Accepted tradeoff.


lennons301:

| gate | glob | path |
|---|---|---|
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/fixtures/verdict-approve-with-preamble.ndjson` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/fixtures/verdict-request-changes-with-preamble.ndjson` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/verdict.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/verdict.ts` |

lennons301-reviewer (approved):
_Recovered review: the ticket-loop review pass completed this verdict but failed to submit it to GitHub (runner reported success without verifying — to be filed against the platform runner). Posted by the supervising session on the pass's behalf._


Reviewed PR #99 against issue #94 with fresh eyes. The change does what the ticket asked — and only that.

**Objective signals (CI reported none, so I ran them locally):**
- Full suite green: 458 tests pass (`vitest run`).
- Lint clean on all changed files (`eslint`, exit 0).
- The two pre-existing `tsc` errors in `container-manager.test.ts` are untouched baseline noise (documented in memory as the known lint/typecheck baseline), unrelated to this diff.

**Ticket fidelity — verified the fix matches the proposed remedy:**
- `parseReviewVerdict` (`src/lib/orchestrator/autonomy/verdict.ts`) now scans for the first line matching `^VERDICT:\s*(approve|request-changes|escalate)` instead of demanding position zero, and takes the body as everything after that line. This is exactly the ticket's proposed regex and "first match wins" rule.
- Line-start anchoring is preserved, so the incident's inverse — a marker quoted mid-sentence (`…my conclusion is VERDICT: approve`) — stays `unparseable`. I confirmed this against the `verdict-malformed.ndjson` fixture and its test.
- The prompt (`buildReviewPrompt`) is strengthened belt-and-braces to demand the VERDICT line first and steer verification notes into the body, while the parser tolerates a preamble regardless — matching the ticket's "the prompt alone demonstrably doesn't hold."

**Safety property holds (this is the autonomy-control surface — the thing that decides what auto-merges):**
- Nothing turns a non-approval or an unparseable stream into an approval: `kind` is taken only from a captured `VERDICT:` line; no such line ⇒ still `unparseable`; a request-changes/escalate with an empty body ⇒ still `unparseable`. All three properties remain tested.
- The new fixtures reproduce the real LPS #144 incident (approve-with-preamble) and a request-changes-with-preamble, and the tests assert both the kind and the exact body (preamble dropped, post-verdict text kept) — they genuinely exercise the new behavior rather than restating it.
- The reason-string rename ("does not start with…" → "has no VERDICT: line") is applied consistently across the parser, the fleet-view "needs you" sample, and the retry-prompt feedback test. I grepped the tree: no stale references to the old string remain, and no consumer switches on the exact reason text (it is only displayed and fed back as free-form prompt text), so the rename is safe.

**Smell baseline / self-review notes — I agree with the author's disposition:**
- The Duplicated-Code divergence they flagged (`parseTriageExit` still requires position zero while `parseReviewVerdict` now scans) is correctly scoped out. The incident was reported against the review verdict; a shared helper would abstract over a genuine behavioral difference, and loosening triage's parse tolerance is a separate decision with its own consequences. Sound as a follow-up, not a blocker here.
- Dropping the end-anchor (accepts `VERDICT: approve — ship it`) and first-match-wins are faithful to the ticket's literal regex and its stated tradeoffs; the trailing decoration lands on the verdict line and is harmlessly excluded from the body.

**Non-blocking suggestion (not tied to a standard — a suggestion only):** two explanatory comments outside this diff now describe the pre-#94 behavior as the cause of the retry path — `src/lib/orchestrator/autonomy/decide.ts:714` and `src/db/schema.ts:85` both say a pure format slip is one "whose final message just didn't lead with a VERDICT: line." After this fix, a message that doesn't *lead* with the marker but carries one later parses fine; the accurate framing is now "had no VERDICT: line at all." The mechanism they describe is unchanged and still correct, so this is minor doc-drift, not a defect — worth a one-line touch-up if convenient, but not a reason to hold the PR.

Gate: correctly carries `human-signoff` (touches `src/lib/orchestrator/autonomy/**`, the autonomy-control gate) with auto-merge off. My approval satisfies branch protection; a human performs the merge.

---

Relevant files:
- `src/lib/orchestrator/autonomy/verdict.ts` (the parser change)
- `src/lib/orchestrator/autonomy/workflow.ts` (prompt strengthening)
- `src/lib/orchestrator/autonomy/__tests__/verdict.test.ts` and the two new fixtures under `.../__tests__/fixtures/`
- Non-blocking suggestion targets: `src/lib/orchestrator/autonomy/decide.ts` (line 714) and `src/db/schema.ts` (line 85)

---

**Summary:** The reviewer returned **VERDICT: approve** — all 458 tests pass, lint is clean, and the fix faithfully matches issue #94's proposed remedy without over-reaching. The PR remains gated with `human-signoff`, so a human still performs the merge. There's one optional, non-blocking doc-drift touch-up noted (two comments in `decide.ts:714` and `schema.ts:85`) if you want to tidy it before merging.

PR #100, opened 2026-08-05
Stop parked autonomous containers to free memory (host OOM fix) (#93)

### LINKED ISSUE #93: Parked autonomous containers hold memory but no slot — host OOM took down the VPS (2026-08-04)
## What happened

First multi-ticket autonomous batch on lennons301/last-person-standing (2026-08-04): the CX22 host OOM-thrashed (load average 60–71, SSH unable to complete banner exchange for ~1.5h) with **4 agent containers live on a 2-slot box**. The app container never crashed — the host starved around it, taking the dashboard and all runs down with it.

dmesg evidence (host-level, global OOM — not cgroup):

```
[17:03:27] Out of memory: Killed process 276339 (node) total-vm:34963796kB, anon-rss:851460kB ... constraint=CONSTRAINT_NONE
[18:42:30] Out of memory: Killed process 283438 (node) total-vm:34963832kB, anon-rss:811896kB ... constraint=CONSTRAINT_NONE
```

plus per-container cgroup kills at 17:13 and 18:42.

## Why

`occupiedSlots()` (`src/lib/orchestrator/queue.ts`) deliberately excludes parked containers, per the comment at `turn-manager.ts:47-59`: a parked implement pass keeps its container but runs no agent process, so it holds no slot — otherwise two parked implements plus their two queued reviews would deadlock a two-slot box.

The design assumes parked ≈ free. That's true for CPU, false for memory: a parked container keeps its RSS (dev server left running for preview, node processes). So 2 implements park → both slots free → 2 review containers start → **4 containers, each with a 1200 MiB allowance, against ~2.7 GiB usable**. Cgroup limits don't reserve memory; the host OOMs on overcommit before any container hits its own cap.

## Proposed fix

`docker stop` the container when an **autonomous** pass parks (awaiting review, or blocked on a question); `docker start` on resume (verdict fix-up turn / owner's answer). Stopped containers keep their filesystem and branch state, cost ~1s to restart, and free all memory. The no-deadlock property is preserved — a stopped parked container holds neither slot nor RAM. Interactive sessions keep today's behavior (their dev server and next message are live concerns, and they already hold their slot while idle).

Defense in depth worth considering in the same change: when starting any container, check total live containers × per-agent allowance against the daemon's reported memory (the same numbers `deriveCapacity` already uses) and refuse/log rather than overcommit.

## Interim mitigation (already applied in prod)

`CAPACITY_SLOTS=1` in Doppler `interlude/prd` — serial claiming keeps worst-case live containers at ~2 (one parked awaiting review + one review pass). Remove the override once this issue ships.

lennons301:
Closes #93

## Problem

The first multi-ticket autonomous batch OOM-thrashed the CX22 host on 2026-08-04: **4 agent containers live on a 2-slot box**. `occupiedSlots()` deliberately excludes parked containers (a parked implement pass runs no agent process, so it holds no *slot*), but a parked container kept its **memory** — its dev server and node heap. Two implements parked → both slots read free → two review containers started → 4 containers × 1200 MiB against ~3 GiB usable. Cgroup limits *cap* but never *reserve* memory, so the host OOMed on overcommit before any container hit its own cap.

## Fix

**Primary — stop parked autonomous containers, restart on resume.** When an autonomous pass parks (implement awaiting its review verdict, or a run blocked on a question), `docker stop` its container; `docker start` it on resume (a review fix-up turn, an owner's answer, or an owner completing the task). A stopped container keeps its filesystem and branch state, costs ~1s to restart, and holds neither a slot nor RAM — so the no-deadlock property that motivated excluding parked containers from the slot count is preserved. Interactive sessions are never parked (their dev server and next message are live concerns), so they keep today's behavior.

Park seams covered: `finishImplementPass` (awaiting review), `parkBlockedRun` (blocked), and the re-park after a review fix-up turn (`processQueuedMessages`). Resume seams covered: `processQueuedMessages` (fix-up / answer) and `completeTask` (owner completion). The sweep's approve/escalate/merge release path force-removes the stopped container — no exec needed.

**Defense in depth — memory admission.** Before starting any container (fresh pickup *and* resume), `checkMemoryAdmission()` asks the daemon how many agent containers are actually running and refuses to start another when the live set plus the newcomer would exceed usable memory (the same `MemTotal − reserve` and per-agent figures `deriveCapacity` uses, now shared via `usableMemoryBytes()`). Independent of the in-memory slot count, so it also catches drift/leaks. The first container is always admitted (matching `deriveCapacity`'s slot floor of 1); it fails open on a Docker probe error so a daemon hiccup can't wedge the queue.

## Interim mitigation to remove after this ships

`CAPACITY_SLOTS=1` in Doppler `interlude/prd` — remove the override once this is deployed (noted in #93).

## Tests

`wouldOvercommitMemory` is a pure predicate with unit tests covering the incident (a 2-slot box admits 2 live containers, refuses the 3rd), the per-agent-override step, and the small-box floor. Full suite: 462 passing. Lint clean on changed files.

## Self-review

Ran the `/code-review` skill (fixed point `origin/main`, spec = issue #93). The **spec axis** reported the change fully implements #93 with no missing park/resume points. The **standards axis** found one real correctness bug — a mid-drain turn could re-park (stop) the container and then the loop would exec the next queued message into it — now fixed by breaking the drain whenever an implement/repair pass parks mid-loop (the next turn arrives via a fresh resume that restarts the container first). Also addressed: stale "container stays alive" comments and the triplicated usable-memory arithmetic.

Accepted residuals (noted, not fixed here):
- **Memory-gate TOCTOU / fairness:** pickup and resume each probe memory independently within one 2s poll, and a resume is not prioritized over new pickups. Both are latency/fairness concerns on a memory-tight box, never a deadlock (memory always frees as running work terminates), and the primary stop-on-park fix is the real guard — this probe is belt-and-suspenders.
- **`completeTask` starts without a memory check:** an owner-initiated terminal action that does one final push then removes the container; a momentary overcommit there is acceptable and asymmetric by design.

### DIFF
--- a/src/lib/docker/container-manager.ts
+++ b/src/lib/docker/container-manager.ts
@@ -413,6 +413,25 @@ export async function stopContainer(
   }
 }
 
+/**
+ * Start (or resume) a container's Docker process. Idempotent: the daemon
+ * answers an already-running container with a 304, which is not an error we
+ * care about — a genuine start failure (bad image, missing container) still
+ * propagates so the caller's turn fails loudly rather than execing into a dead
+ * container. Used to resume a container `stopContainer` parked to free memory
+ * (issue #93).
+ */
+export async function startContainer(
+  running: RunningContainer
+): Promise<void> {
+  try {
+    await running.container.start();
+  } catch (err) {
+    if ((err as { statusCode?: number })?.statusCode === 304) return; // Already running
+    throw err;
+  }
+}
+
 export async function removeContainer(
   running: RunningContainer
 ): Promise<void> {
--- a/src/lib/orchestrator/__tests__/capacity.test.ts
+++ b/src/lib/orchestrator/__tests__/capacity.test.ts
@@ -1,5 +1,9 @@
 import { describe, it, expect } from "vitest";
-import { createLocalCapacityProvider, deriveCapacity } from "../capacity";
+import {
+  createLocalCapacityProvider,
+  deriveCapacity,
+  wouldOvercommitMemory,
+} from "../capacity";
 
 const MiB = 1024 * 1024;
 const GiB = 1024 * MiB;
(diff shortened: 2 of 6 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
Approving. This is a gated PR (touches `src/lib/docker/**`, the agent-sandbox gate) — the `human-signoff` label is correctly in place and auto-merge is off, so this approval only satisfies branch protection; a human performs the merge. I have left the gate untouched.

## What I verified

**Objective signals** — Full suite green locally (462 passing, 23 files). ESLint clean on all six changed files. No CI checks are configured on the branch, so I ran them locally. PR is MERGEABLE against current main (main only moved `src/lib/fleet/*`, which this PR does not touch).

**Ticket (#93) — does what was asked, all of it, only it:**
- `docker stop` on every autonomous park seam: `finishImplementPass` (awaiting review) and `parkBlockedRun` (blocked on a question), plus the mid-drain re-park in `processQueuedMessages`.
- `docker start` on every resume seam: `resumeParkedContainer` (fix-up / answer) and `completeTask` (owner completion). The sweep's release path force-removes the stopped container via `removeContainer({force:true})` — no start needed.
- No-deadlock property preserved: `isParked` still excludes parked containers from `occupiedSlots`; a stopped parked container holds neither slot nor RAM.
- Interactive sessions untouched (`isParked` excludes `kind === interactive`).
- Defense-in-depth `checkMemoryAdmission` wired into both pickup (queue.ts) and resume (turn-manager.ts).

**Correctness — traced the incident and the tricky seams:**
- Replayed the 2026-08-04 scenario: two implements now stop-on-park, so max concurrent *running* containers on the 2-slot box is the two reviews (2×1200 MiB < ~2776 MiB usable) rather than four — the OOM cannot recur. The memory gate independently refuses a drifted 3rd start.
- `wouldOvercommitMemory` is consistent with `deriveCapacity`: it admits exactly `floor(usable/perAgent)` containers, always ≥ the (un-overridden) slot count and equal at the floor-of-1 case — so it never starves below configured slots, and it *is* stricter only when slots are overridden above what memory fits (the intended backstop). The shared `usableMemoryBytes` removes the earlier triplication (a real duplication smell, correctly addressed in the self-review commit).
- Deferred resume (`resumeParkedContainer` → false) returns before marking the message delivered and before the trailing re-park, so the queued turn is retried on the next poll with no lost message and no double-stop. Memory always frees as running work terminates, so this is latency, never deadlock.
- The mid-drain `break` after a re-park (blocked, or handed-over-PR park) correctly prevents execing the next queued message into a just-stopped container — the case flagged and fixed in self-review.
- Reaper still preserves parked containers (`all:true` listing, blocked/active-run guard); `checkMemoryAdmission`'s default listing counts only running containers, so parked ones correctly weigh nothing.

**Standards / smell baseline:** names are honest, no speculative generality, no scope creep. Tests cover the pure predicate (`wouldOvercommitMemory`) with incident/floor/override cases; the Docker lifecycle glue (`startContainer`) follows the established untested pattern of `stopContainer`/`removeContainer`, consistent with this repo's convention of testing command-builders and the `decideNext` reducer rather than imperative Docker calls.

## Non-blocking notes (suggestions, not tied to a standard)
- The two residuals the PR body already documents (pickup/resume TOCTOU + resume-vs-pickup fairness; `completeTask` starting without a memory check) are sound engineering calls — never a deadlock, asymmetric by design.
- Minor: `resumeParkedContainer` logs 'Deferring resume' every 2s poll while memory is saturated, without a `saturationLogged`-style guard like the pickup path has. Transient and harmless, but a candidate for the same one-shot log guard if it proves noisy.

Reminder for the human merging: the interim `CAPACITY_SLOTS=1` override in Doppler `interlude/prd` should be removed once this deploys (called out in #93 and the PR body).

Issue #97, opened 2026-08-04
Infra-killed passes burn attempts and format-retries: OOM'd runs failed at $0 spend


lennons301:
## What happened

In the 2026-08-04 OOM incident (#93), LPS #140 and #142 both ended `status: failed, attempt: 1, total_cost_usd: 0` — their implement passes were killed by the host OOM before doing any recorded work. Each burned one of the ticket's 3 attempts. Similarly, #137's bounded verdict format-retry (#89/#92) was consumed by a review pass that died with "review pass output has no result event" — an infra death, not a formatting failure.

## Why this is wrong

Restart recovery (issue #24) already embodies the right principle: an interruption that isn't the work's fault re-claims **without consuming an attempt**, bounded by MAX_INTERRUPTIONS_PER_TICKET. But that path only covers orchestrator restarts. A pass that dies to container OOM / docker error / host pressure while the orchestrator stays up is charged as a failed attempt (or a spent format-retry) exactly as if the agent had produced bad work.

## Proposed fix

Classify pass deaths: when a container exits without a terminal agent result (no result event, exit 137, docker error), count it against the interruption bound rather than the attempt/retry budget. The existing manual recipe (delete the runs rows and re-arm — platform#8/#16 precedent, 2026-08-04) stays as the escape hatch, but the ledger should get this right unattended.

lennons301:
Armed under Sean's direction (2026-08-05, supervising session): incident defect batch runs via the laptop ticket-loop while VPS autonomy stays off. Sequencing: wave 1 = #93/#94/#96 in parallel worktrees; wave 2 = #95 then #97 (both touch the ungraceful-death seam, run after wave 1 lands).

lennons301:
🤖 Attempt 1/3 starting.

PR #45, opened 2026-08-02
Phase 5: review pass, structured verdict and orchestrator-posted review

lennons301:
Closes #17

## What this builds

The loop's landing half: after an implement pass parks its PR behind the gate decision (#16), a **review pass** runs in its own container — fresh clone of the PR branch, fresh context, its own $5 budget, queued as a `kind: "review"` task drawing a slot like any other pass. Its output is parsed into a structured verdict by a pure parser, stored on the run (`runs.review_result`, additive migration 0009, restart-safe), and the **orchestrator** posts the review through `postReviewAsReviewer` using `REVIEWER_GH_TOKEN`. The PAT never enters any container; the review container gets nothing beyond the App token its setup uses for cloning — not even the project's Doppler secrets.

Verdict consequences, all mapped in `decideNext`:

- **approve** → approval posted; on an armed PR auto-merge lands it, the parked implement container is released
- **request-changes** → auto-merge disarmed, review posted, findings delivered to the **still-live implement container** as a follow-up turn in the same attempt; the fix-up re-enters gate evaluation, so new commits are re-gated before review cycle 2
- **escalate** → auto-merge disarmed, `human-signoff` applied, comment-type review posted
- **unparseable** → fails closed: no review posted, disarm + `human-signoff` + issue comment + Discord (once); *never* `armAutoMerge` — pinned by test
- a reviewed PR that closes settles the run ledger and releases any parked container

## Attempt-3 rework (review round 2, the one finding)

**`docs/agents/review-pass.md` is now gated under `autonomy-control`** in `docs/agents/review-gates.yaml` (exact-path glob — the evaluator's bash `[[ == ]]` semantics match literals directly, verified against `gates.ts`). The category comment now names the principle: the reviewer definition is injected verbatim into every review pass and is the thing that decides what auto-merges, so changing it always takes human sign-off. As you noted, no code or test changes were needed — the pure evaluator is already pinned by inline fixtures, and the live YAML is runtime data read from the default branch.

Self-review was re-run on the integrated head this attempt. Two judgement-call findings it surfaced that earlier rounds didn't name explicitly, disclosed rather than acted on — both would churn a twice-verified surface for a cosmetic gain on the final attempt:

[...]

PR #101, opened 2026-08-05
Duplicate concurrent review passes queued for one run (#95)

### LINKED ISSUE #95: Duplicate concurrent review passes queued for one run

lennons301:
Closes #95

## Problem

During the 2026-08-04 OOM incident, one run had **two review tasks running simultaneously** for the same PR — duplicate spend, racing verdicts, and extra memory pressure exactly when the box could least afford it.

The sweep's in-flight-review check keys on task status (`kind = review AND status IN (queued, running)`). A review pass whose container is OOM-killed or lost while the daemon hangs can leave its task stuck `running` with **no status transition** (corroborated by #96's "stale `running` rows after ungraceful container deaths"). Keyed on task status alone, that dead-but-stuck task reads as live forever: the run either **stalls** (never replaced), or — once anything re-queues — **races a duplicate** against its own PR.

## Fix

Pair the existing idempotency rule with an ungraceful-death path, so a run has *exactly one* review in flight — never zero when one is owed, never two racing.

- **`inFlightReviewTaskId`** — extract the in-flight check (issue #45) into a shared, tested helper so "an existing review task queued/running counts as in flight" is explicit.
- **`reapDeadReviewTasks`** — before each gather, confirm each running review task's container is actually gone (Docker probe fails safe: only a *definitively* dead container is reaped; a transient daemon error is left for the next sweep), mark it `failed`, drop its in-memory session entry and remove the dead container. The next sweep's in-flight check then queues exactly one deliberate replacement.
- **Verdict-store guards** in `finishReviewPass` and the review-pass catch, so a reaped-then-unhung pass can't clobber its replacement's run with a stale verdict (prevents the spec's "racing verdicts for the same PR").

Marking a review task `failed` does not fail the run or consume an attempt — a review death is infra, not bad work. Attempt/budget accounting for infra deaths is deliberately left to sibling **#97**.

New module `src/lib/orchestrator/autonomy/review-tasks.ts` is DB-only (no Docker/GitHub imports); the container-liveness probe is injected so the sweep passes the real daemon check and tests pass a fake.

## Test

`review-tasks.test.ts` pins the spec's requested scenario — *review pass dies without a status transition → exactly one replacement queued* (not zero, not two) — plus: a live review is left alone and no second is queued; queued/provisioning passes aren't reaped; the queued replacement isn't itself reaped or double-queued.

## Validation

- `vitest run` — 24 files, 473 tests pass.
- ESLint clean on all changed files.
- `tsc --noEmit` introduces no new errors (the two pre-existing errors are in an untouched test file, present on `main`).

## Self-review

Ran `/code-review` against `origin/main` with issue #95 as the spec. **Spec axis**: faithful and correct — no missing/wrong requirements; the memory-cleanup and verdict-guard additions were confirmed to serve the ticket's stated Impact, and false-positive safety (never reaping a healthy finishing pass) was verified. **Standards axis**: no hard violations; acted on all three duplication/consistency judgement calls in the follow-up commit.

### DIFF
--- a/src/lib/docker/container-manager.ts
+++ b/src/lib/docker/container-manager.ts
@@ -432,14 +432,20 @@ export async function startContainer(
   }
 }
 
-export async function removeContainer(
-  running: RunningContainer
-): Promise<void> {
+/** Force-remove a container, idempotently — an already-gone container is not
+ * an error. Shared by the handle- and name-based removers below. */
+async function forceRemove(container: Docker.Container): Promise<void> {
   try {
-    await running.container.remove({ force: true });
+    await container.remove({ force: true });
   } catch {
     // Already removed
   }
+}
+
+export async function removeContainer(
+  running: RunningContainer
+): Promise<void> {
+  await forceRemove(running.container);
 }
 
 /**
@@ -461,12 +467,7 @@ export async function isContainerRunning(name: string): Promise<boolean> {
 }
 
 /** Force-remove a container by name — for reaping a dead pass's container
- * whose `RunningContainer` handle the orchestrator no longer holds (issue #95).
- * Idempotent: an already-gone container is not an error. */
+ * whose `RunningContainer` handle the orchestrator no longer holds (issue #95). */
 export async function removeContainerByName(name: string): Promise<void> {
-  try {
-    await getDocker().getContainer(name).remove({ force: true });
-  } catch {
-    // Already removed
-  }
+  await forceRemove(getDocker().getContainer(name));
 }
--- a/src/lib/docker/container-manager.ts
+++ b/src/lib/docker/container-manager.ts
@@ -441,3 +441,32 @@ export async function removeContainer(
     // Already removed
   }
 }
+
+/**
+ * Whether a container (by name) is still running, for the ungraceful-death
+ * reaper (issue #95). Fail-safe on the side of not touching a live pass: a
+ * container the daemon has definitively lost (404) or that has exited returns
+ * false so its stale `running` task can be reaped; any *other* daemon error
+ * (a busy or briefly unreachable socket under memory pressure) returns true —
+ * assume live and let the next sweep retry, so a transient hiccup can never
+ * terminalize a pass that is genuinely still working.
+ */
+export async function isContainerRunning(name: string): Promise<boolean> {
+  try {
+    const info = await getDocker().getContainer(name).inspect();
+    return info.State?.Running === true;
+  } catch (err) {
+    return (err as { statusCode?: number })?.statusCode !== 404;
+  }
+}
+
+/** Force-remove a container by name — for reaping a dead pass's container
+ * whose `RunningContainer` handle the orchestrator no longer holds (issue #95).
+ * Idempotent: an already-gone container is not an error. */
+export async function removeContainerByName(name: string): Promise<void> {
+  try {
+    await getDocker().getContainer(name).remove({ force: true });
+  } catch {
+    // Already removed
+  }
+}
--- a/src/lib/orchestrator/autonomy/__tests__/review-tasks.test.ts
+++ b/src/lib/orchestrator/autonomy/__tests__/review-tasks.test.ts
@@ -1,0 +1,172 @@
+import { describe, it, expect, beforeEach } from "vitest";
+import { eq } from "drizzle-orm";
+import * as schema from "@/db/schema";
+import { createTestDb } from "@/test/create-test-db";
+import { inFlightReviewTaskId, reapDeadReviewTasks } from "../review-tasks";
+import { decideNext, passOutcomeSnapshot, type AutonomySnapshot } from "../decide";
+
+type Db = ReturnType<typeof createTestDb>["db"];
+
+const ISSUE_REF = "owner/repo#141";
+
+/** A run parked in `reviewing` (auto-merge armed) awaiting its review pass. */
+function seedReviewingRun(db: Db): void {
+  db.insert(schema.projects).values({ id: "p1", name: "Test", createdAt: new Date() }).run();
+  db.insert(schema.runs)
+    .values({
+      id: "run-1",
+      projectId: "p1",
+      githubIssue: ISSUE_REF,
+      attempt: 1,
+      mode: "autonomous",
+      status: "reviewing",
+      budgetUsd: 20,
+      pullRequestNumber: 144,
+      claimedAt: new Date(),
+    })
+    .run();
+  // The parked implement pass keeps its (stopped) container while its PR is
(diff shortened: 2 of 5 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
_Recovered review (platform#21, second occurrence): the ticket-loop review pass completed this verdict but never submitted it to GitHub. Posted by the supervising session on the pass's behalf._

## Verdict: APPROVE

PR #101 — "Duplicate concurrent review passes queued for one run (#95)". This PR is **GATED** (`human-signoff` label present, auto-merge off) — correctly, because it touches two gated globs: `src/lib/docker/**` (agent-sandbox) and `src/lib/orchestrator/autonomy/**` (autonomy-control). My approval only satisfies branch protection; a human still performs the merge.

### What I verified

**Faithful to the ticket.** Issue #95 asked for exactly three things, and the PR delivers each without overreach:
1. Treat any `kind=review` task in `(queued, running)` for the run as in-flight → `inFlightReviewTaskId` (`src/lib/orchestrator/autonomy/review-tasks.ts:47`), extracted from the previously-inline query in `gatherReviewState` (removes a duplication, not adds one).
2. Make the ungraceful-death path mark the stuck task terminal first → `reapDeadReviewTasks` (`review-tasks.ts:84`) + `reapOrphanedReviewPasses` (`sweep.ts:203`), run before every gather.
3. A test pinning "review pass dies without a status transition → exactly one replacement queued (not zero, not two)" → `review-tasks.test.ts:64`.

Attempt/budget accounting for infra deaths is explicitly deferred to sibling #97, and the PR stays within that boundary (a reaped review is marked `failed` but does not fail the run or consume an attempt).

**The "exactly one" invariant holds end-to-end**, which I traced through the reducer, not just the unit test:
- Pure hung-container case (no throw, the primary target): reaper flips the stuck task to `failed`, so `inFlightReviewTaskId` returns null and `decideNext` emits exactly one `startReview` (`decide.ts:817-828`). Confirmed by the first test via `decideNext`, not just a DB assertion.
- Reaped-then-unhangs case: the two verdict guards (`finishReviewPass` at `turn-manager.ts:924` and the catch's `alreadyReaped` at `turn-manager.ts:362`/`381`) both key on the reaped task's own status being non-`running`/`failed`, so a late-returning dead pass cannot clobber its replacement's `run.reviewResult`.
- The messy interleaving where the container dies *and* `runTurn` throws: even if the catch stores an unparseable verdict before the reaper's `await` resumes, `run.reviewResult != null` short-circuits at `sweep.ts:515` (into `pendingVerdicts`) *before* `awaitingReview.push`, so the run gets the fail-closed human-signoff path and **no** duplicate review is queued. The outcome is always safe (no bad auto-merge, no lost run).

**False-positive safety (never reap a healthy pass) is sound**, which is the load-bearing risk here:
- `isContainerRunning` (`container-manager.ts:454`) fails safe: only a definitively dead container (404, or `State.Running !== true` for an exited one) returns false; any other daemon error returns true (assume live). Under memory pressure a busy/unreachable socket therefore cannot terminalize a working pass — it waits for the next sweep.
- A provisioning review (status `running`, `containerName` still null because `startTask` sets status at `turn-manager.ts:187` before the container exists at :250) is skipped by the `if (!task.containerName) continue` guard (`review-tasks.ts:104`). Pinned by the "still provisioning" test.
- Teardown sets status `completed` *before* removing the container (`turn-manager.ts:983-987`), so a normally-finishing pass is out of the `running` set before its container disappears — the reaper can't catch it in the gap.
- The parked implement container (kept alive, task stays `running`, `kind=implement`) is untouched — the reaper filters on `kind=review`. Pinned by the seed in the test.

**Smell baseline** — clean. The three helpers introduced by the self-review commit are genuine de-duplication, not new smells: `forceRemove` (`container-manager.ts:437`) unifies the two container removers; `taskStatus` (`turn-manager.ts:1308`) is reused by both verdict guards; the shared `Db` type names the injected seam once. Names reveal intent; the DB-only module with an injected Docker probe is justified by testability, not speculative generality. `reapDeadReviewTasks` reaping globally (all runs) rather than per-run is correct cleanup and idempotent.

**Standards.** Container-limit / no-bind-mount conventions untouched; the reviewer-identity model is respected (no PAT handling changed). Documentation on the new module and the reaper is thorough and accurate to the behaviour I traced.

### Findings

None that block. One optional observation, not tied to a standard (a suggestion): `review-tasks.ts` imports both `* as schema` (for the `Db` type parameter) and the named `{ messages, tasks }` from the same module — mildly redundant but consistent with the codebase's existing pattern, and not worth a change.

### Recommendation to the human merger

The work is complete and correct; I'd be comfortable with it on `main`. It is gated for the right reason (Docker sandbox + autonomy control surface), so a human should give it the customary look before merging — the highest-leverage thing to eyeball is the fail-safe direction of `isContainerRunning` (`src/lib/docker/container-manager.ts:454`), since that single boolean is what stands between "reap a genuinely dead pass" and "never kill a live one." I've verified it errs toward "assume live," which is the safe default.

---

**Summary:** APPROVE. No blocking findings. The PR faithfully implements issue #95's "exactly one review pass" invariant, all 473 tests pass, ESLint is clean on the five changed files, and the two `tsc` errors are pre-existing on `main` in a file this PR doesn't touch. It's correctly gated for `human-signoff` — a human still performs the merge, with `isContainerRunning`'s fail-safe direction flagged as the highest-leverage thing to eyeball.

Issue #89, opened 2026-08-04
Unparseable review verdict: one bounded format-retry before falling to human oversight


lennons301:
First live occurrence (platform#16 / PR #20, 2026-08-04): the review pass wrote a substantively fine review whose final message didn't start with the required `VERDICT:` line. The orchestrator correctly failed closed (nothing posted, run gated, notify-once) — but terminally: no retry exists, so a pure *format* slip costs a human intervention.

**Fix:** on `kind: unparseable`, re-queue the review pass **once** with the parse failure appended to its prompt ('your previous verdict was unparseable because X — restate it starting with VERDICT:'). Second unparseable → current behavior (human oversight). Bounded, cheap, removes the most common cause.

**Sibling fact from the same incident:** the notify-once went to logs because `DISCORD_FLEET_CHANNEL_ID` is unset — the owner found the stall by watching the dashboard. Configuring the fleet channel is operational, not code.

lennons301:
Armed `ready-for-agent` under Sean's standing authorization (2026-08-04) — queued behind #90.

lennons301:
🤖 Attempt 1/3 starting.

lennons301:
🤖 Attempt 2/3 starting.

Issue #54, opened 2026-08-02
Native executor: treat a CONFLICTING parked PR as needs-integration, not done (mirror of platform#14)


lennons301:
platform#14 gives the laptop runner a mergeability step: PR-state dispatch after ticket selection, a repair pass (merge main, never rebase), repair accounting that never burns an attempt, and `mergeable: UNKNOWN` polled rather than trusted. Its out-of-scope note explicitly delegates the same policy to interlude's native executor — this ticket.

**Gap, verified in code:** `src/lib/orchestrator/autonomy/` polls PR state (`getPrState`) for merged/closed but never reads `mergeable`. Every human merge of a gated PR can flip other parked runs' PRs to `CONFLICTING`, and nothing notices — unattended, that's an invisible stall (observed three times in one day on the laptop loop during Phase 5 delivery: PRs #38, #41, #52).

**Mirror policy, adapted to the reducer:**
- Snapshot gains each pending run's PR `mergeable` state (`UNKNOWN` → re-poll, never a verdict)
- `decideNext`: a run whose PR is `CONFLICTING` emits a **needs-integration** action → executor queues a repair task (fresh container, merge default branch in — merge only, no rebase/force-push), then the normal gate re-evaluation + review pass re-run on push
- Repair never consumes an attempt; a repair that leaves the PR `CONFLICTING` escalates to `ready-for-human` + a needs-you entry
- Fleet dashboard: parked-but-conflicting is a distinct visible state (needs-you bucket), not silence

**Upstream / when to arm:** after platform#14's implementation lands and settles (same policy, one source of truth for its semantics) and the Phase 5 milestone tail (#24, #26) is done.

Found during Phase 5 delivery, 2026-08-02.

lennons301:
Armed `ready-for-agent` under Sean's standing authorization (2026-08-02) — upstreams now ready: platform#14 merged (runner has the repair machinery to mirror) and the Phase 5 milestone is closed. Human decision, agent applied the label.

lennons301:
🤖 Attempt 1/3 starting.

Issue #24, opened 2026-07-30
Phase 5: restart recovery and interruption accounting


lennons301:
## Parent

#11 — Phase 5: Autonomous Ticket-Loop + Fleet Observability (spec)

## What to build

Interlude deploys itself: merging one of its own PRs restarts the orchestrator, potentially mid-run. That makes restart-during-active-work a routine event here, not an edge case — so recovery has to be deterministic, and the ticket must not be charged for the platform's downtime.

## Acceptance criteria

- [ ] On boot, runs whose container is gone or whose exec stream was lost are marked interrupted and their container stopped
- [ ] An interrupted run is re-claimed **without** consuming one of its three attempts — work already pushed to the branch survives; only the in-flight turn is lost
- [ ] Interruptions are counted separately and bounded, so a ticket that crashes the orchestrator cannot loop forever
- [ ] The container reaper skips containers belonging to live runs
- [ ] Explicitly tested: a deploy-triggered restart while autonomous runs are active leaves the fleet consistent, re-claims without consuming attempts, and reaps nothing it shouldn't

## Seams under test

`workflow:tdd` — /tdd must not write a test at an unconfirmed seam, and an unattended pass has no one to confirm with. These are the confirmed seams:

- `decideNext`'s interruption accounting: an interrupted run re-claims without consuming an attempt, interruptions are counted separately, and the re-claim is bounded

No Docker, GitHub, Discord or database mocking — if a test needs one, it is at the wrong seam.

## Blocked by

- Autonomous pickup — label to claim to implement pass to PR

lennons301:
🤖 Attempt 1/3 starting.

lennons301:
🤖 Attempt 2/3 starting.

PR #102, opened 2026-08-05
Infra-killed passes burn attempts and format-retries: classify container deaths as interruptions

### LINKED ISSUE #97: Infra-killed passes burn attempts and format-retries: OOM'd runs failed at $0 spend

lennons301:
Closes #97

## What was wrong

A pass whose container died while the orchestrator stayed up was charged exactly as if the agent had produced bad work:

- In the 2026-08-04 OOM incident, LPS **#140/#142** implement passes were host-OOM-killed before any recorded work (`status: failed, attempt: 1, $0`) — each burned one of the ticket's 3 attempts.
- **#137**'s bounded verdict format-retry (#89/#92) was consumed by a review pass that died with "review pass output has no result event" — an infra death, not a formatting failure.

Restart recovery (#24) already embodied the right principle — an interruption that isn't the work's fault re-claims **without consuming an attempt**, bounded by `MAX_INTERRUPTIONS_PER_TICKET` — but it only covered orchestrator restarts.

## The fix

Classify in-flight pass deaths by whether a **terminal agent `result` event** arrived (new pure `passProducedResult`, the boundary the issue names: "no result event, exit 137, docker error"):

- **Implement pass, no terminal result** (thrown mid-turn, or a non-throwing turn that emitted no result event): mark the run `interrupted`, not `failed`. It counts toward the interruption bound, so the sweep re-claims the ticket without consuming an attempt — still bounded, so a ticket that reliably kills its container routes to `ready-for-human` like an exhausted one. A terminal result that *then* failed (budget/turn exhaustion, a post-turn push error) is still a real failed attempt.
- **Review pass, no terminal result**: leave the verdict unstored and mark the pass failed so the sweep queues a fresh replacement review — the same recovery the **#95** reaper already gives a hung review container (its comment explicitly named #97 as the generalization), consuming no format-retry. A genuine format slip (result present, no `VERDICT:` line) still spends the retry as before.

Infra deaths auto-recover, so they no longer announce a "task failed" on the issue or in Discord. The exhaust/Discord/runbook wording for the interruption bound now names container deaths alongside restarts.

The manual recipe (delete the runs rows and re-arm) stays as the escape hatch; the ledger now gets this right unattended.

## Boundary, made explicit

`passProducedResult` returns **true** whenever a `result` event exists — even `is_error` / `error_max_turns` — so an errored-but-terminal turn stays a genuine failure/format-slip (pinned by the `verdict-errored-turn` vs `verdict-truncated` fixtures, the two shapes that both parse as unparseable verdicts). Only a stream with no `result` event at all is an interruption.

## Scope notes (from self-review with `/code-review`)

- **Setup failures** (container provisioning / clone before the first turn) now also classify as interruptions rather than failed attempts — a deliberate, in-spirit generalization of "not the work's fault," still bounded.
- **Repair passes** (#54) are unchanged: a repair infra death still fails via `finishRun` — repairs are never attempts, and their run is a live `reviewing`/`gated` one, out of scope here.
- Reworded exhaustion messages (Discord + issue comment + runbook) are a justified consequence of interruptions now having a second source.
- I **kept `passProducedResult(rawStream)` self-contained in `finishReviewPass`** rather than threading the caller's flag in — the function already parses the verdict from the same stream, so checking it for a result event is cohesive. I **declined** merging `interruptImplementPass` with `failImplementAttempt` into one parameterized skeleton — their semantics (interrupt-no-attempt vs fail-and-consume) and messaging differ enough that a shared skeleton would over-couple; the shared teardown was extracted instead (`removeTaskContainer`).

## Tests / limits

- New unit tests pin `passProducedResult` (the load-bearing new logic) against the existing stream fixtures; full suite green (480 passing), lint + tsc clean on changed files.
- The DB-transition glue (`interruptImplementPass` marking the run interrupted; the review re-queue) is not directly unit-tested — `turn-manager.ts` has no test harness in this repo (heavy DB + Docker I/O), so this matches the norm. The reducer side (interrupted runs → re-claim without an attempt) is already covered in `decide.test.ts`, and `passProducedResult` is covered here.

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -199,6 +199,14 @@ answer it, or leave it; it doesn't need cancelling to free a slot.)
 - Review passes carry their own **~$5**; triage a small allowance with a low turn cap.
 - **3 attempts**, then the loop swaps `ready-for-agent` for **`ready-for-human`**,
   comments a summary and pings Discord.
+- **Interruptions don't count as attempts.** A run lost to an orchestrator restart
+  *or* to a container that died before finishing (OOM / exit 137 / docker error, i.e.
+  no terminal agent result) is marked `interrupted`, not `failed`: the sweep re-claims
+  the ticket without consuming an attempt. Re-claims are still bounded
+  (`MAX_INTERRUPTIONS_PER_TICKET`, 5), so a ticket that reliably kills its container
+  eventually routes to `ready-for-human` like an exhausted one. A review pass that
+  dies the same way re-queues a fresh replacement instead of burning its one
+  format-retry.
 - **$500/day** estate-wide autonomous cap pauses pickup (announced once, shown on
   the dashboard, resets at local midnight). Interactive work is exempt by
   construction (it has no run).
--- a/src/lib/discord/notifications.ts
+++ b/src/lib/discord/notifications.ts
@@ -486,7 +486,8 @@ export async function notifyAttemptsExhausted(
         ? new EmbedBuilder()
             .setTitle(`Interruption bound hit — ${payload.issueRef} needs you`)
             .setDescription(
-              `${payload.interruptions} runs lost to orchestrator restarts ` +
+              `${payload.interruptions} runs lost to interruptions — restarts or ` +
+                `containers that died before finishing ` +
                 `($${payload.totalSpendUsd.toFixed(2)} autonomous spend). Re-claims are ` +
                 `bounded, so the ticket is now \`ready-for-human\`; the story is on the issue.`
             )
--- a/src/lib/orchestrator/autonomy/__tests__/pass-output.test.ts
+++ b/src/lib/orchestrator/autonomy/__tests__/pass-output.test.ts
@@ -1,0 +1,64 @@
+import { describe, it, expect } from "vitest";
+import fs from "fs";
+import path from "path";
+import { finalPassMessage, passProducedResult } from "../pass-output";
+
+// Fixtures follow the shape of __tests__/stream-fixture.ndjson: the raw
+// stream-json a pass emits, ending (when it finishes cleanly) in a `result`
+// event whose `result` field is the turn's final message.
+function fixture(name: string): string {
+  return fs.readFileSync(path.join(__dirname, "fixtures", name), "utf8");
+}
+
+describe("finalPassMessage", () => {
+  it("extracts the final message from a clean result event", () => {
+    expect(finalPassMessage(fixture("verdict-approve.ndjson"))).toMatchObject({
+      ok: true,
+    });
+  });
+
+  it("fails on a stream that never reached a result event", () => {
+    expect(finalPassMessage(fixture("verdict-truncated.ndjson"))).toMatchObject({
+      ok: false,
+    });
+  });
+});
+
+// The classifier the interruption-vs-failure split hangs on (issue #97): a
+// stream with a terminal `result` event means the agent process ran to a
+// terminal state, so any downstream parse failure is the work's; a stream
+// with none means the container died before finishing (OOM / docker error /
+// lost stream), which must not be charged to the attempt or format-retry
+// budget. The boundary is exactly the errored-turn vs. truncated fixtures:
+// both parse as unparseable verdicts, but only one is an infra death.
+describe("passProducedResult", () => {
+  it("is true for a clean run that ended in a result event", () => {
+    expect(passProducedResult(fixture("verdict-approve.ndjson"))).toBe(true);
+  });
+
+  it("is true for an errored turn that still emitted a result event", () => {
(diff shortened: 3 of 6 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/pass-output.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/pass-output.ts` |

lennons301-reviewer (approved):
## Approve (gated — human merges)

Fresh-eyes review against issue #97. Objective signals green: full suite 480/480 passing locally, lint clean on all changed files, changed files type-check (the one `tsc` error is in `container-manager.test.ts`, untouched by this PR and pre-existing on `origin/main`). No CI checks are configured on the branch, hence the local run.

### Verified against the ticket
- **Implement pass, no terminal result** (thrown mid-turn OR non-throwing turn with no `result` event) → run marked `interrupted` via new `interruptRun`, which mirrors the restart-recovery marking in `init.ts` exactly (`status: interrupted`, `interruptionCount + 1`, `finishedAt`). It feeds the identical accounting the sweep/`decideNext` already use for #24 restarts, so it counts against `MAX_INTERRUPTIONS_PER_TICKET` (bounded → `ready-for-human`), never a strike. This is exactly the #140/#142 OOM-at-$0 scenario the ticket names.
- **Review pass, no terminal result** → `finishReviewPass` leaves `reviewResult` null and marks the task `failed`, which frees `inFlightReviewTaskId` to return null so the sweep queues one replacement review. Null result keeps it out of `pendingVerdicts`, so no format-retry is spent — the #137 scenario. This reuses the established #95 reaper recovery (whose doc-comment already cited #97).
- Setup failures classify as interruptions too (in-spirit generalization, still bounded); repair passes deliberately unchanged. Messaging (exhaust comment, Discord embed, runbook) accurately renamed to name container deaths alongside restarts.

### Correctness
- The classifier boundary is sound: `producedResult = turnResult.subtype !== null` for implement passes (raw isn't captured there) and `passProducedResult(rawStream)` for reviews both key on the presence of a terminal `result` event — no false positives (an errored-but-terminal turn stays a genuine failure, pinned by the `verdict-errored-turn` vs `verdict-truncated` fixtures) and no false negatives (a genuine completion always emits a result event).
- Re-entry is safe: the interrupt comment is fire-and-forget and `removeContainer`/`forceRemove` swallow all errors, so after `interruptRun` bumps the count the remaining teardown cannot throw back into the catch and double-count.
- `removeTaskContainer` extraction cleanly dedupes five terminal-teardown copies (a Duplicated-Code win); declining to merge `interruptImplementPass` with `failImplementAttempt` is the right call — their ledger semantics differ.

### Test coverage
Pure classifier (`passProducedResult`) is well unit-tested including the noise and empty-output edges; the reducer side (interrupted → re-claim without attempt, bound → human) is covered in `decide.test.ts`. The DB/Docker glue in `turn-manager.ts` is untested, matching the repo norm (no harness for that module). Acceptable.

### Notes for the human merging (non-blocking)
- Review infra-death re-queue is **unbounded** (unlike implement interruptions): a PR whose review reliably OOMs would re-queue each sweep forever. This is a pre-existing property of the #95 path this PR deliberately reuses, and each dead review costs ~$0 so it doesn't trip the daily cap — worth awareness, not a defect here.
- The review-pass handling correctly *diverges* from the ticket's literal 'count against the interruption bound' — interrupting the whole run would discard intact implement work; re-queuing the review is the right reading.

Complete, correctly scoped (6 files, no creep), and consistent with #24/#95 conventions.

Issue #63, opened 2026-08-03
Seed composition + follow-on slash routing


lennons301:
## Parent

Decomposed from #50 (Mobile generation sessions spec).

## What to build

Starting a generation session produces a deterministic first turn, and typed slash commands keep working mid-session:

- One pure function composes the seed: (session skill, optional issue anchor, optional agenda) → first-turn prompt. The agent fetches any anchored issue itself via `gh`; the orchestrator passes a reference, not a body.
- The spike ticket's finding selects the internals: slash text passed through if stream-json expands it, otherwise the function inlines the skill content itself.
- Follow-on turns are first-class: the common pipeline is grill → `/to-spec` → `/to-tickets` → arm as later turns of one conversation. If expansion doesn't work for injected turns, outgoing user messages starting with a known skill slash are routed through the same composition mechanism — seed and follow-on turns share one path, and a typed slash never silently degrades into the agent improvising the skill from memory.
- The seed for grill/to-tickets sessions carries the arming convention: publish unlabelled, apply `ready-for-agent` only on owner confirmation in chat, and post an audit comment on each armed issue recording the session and confirmation route.

**Seams under test** (pre-agreed): the seed-composition pure function, fixture-driven — no Docker, GitHub, or database in any test.

## Acceptance criteria

- [ ] Fixture tests cover each session skill (including `wayfinder`) × anchored/freeform × agenda/none, asserting the exact composed prompt
- [ ] A created session's first turn is the composed seed
- [ ] A slash-prefixed follow-up message invokes its skill by the same mechanism as the seed turn (per the spike's finding)
- [ ] Seed content for publishing skills includes the arm-on-confirmation + audit-comment convention

## Blocked by

- #59
- #61

lennons301:
Armed: `ready-for-agent` applied to bring in the Mobile generation sessions functionality (batch #50). Route: owner-confirmed in an interactive Claude Code session on 2026-08-10 ("arm everything we need to get this functionality in"), under the standing arming authorization for interlude tickets (2026-08-02, interactive sessions only).

lennons301:
🤖 Attempt 1/3 starting.

Issue #60, opened 2026-08-03
Agent containers install mattpocock-skills at start


lennons301:
## Parent

Decomposed from #50 (Mobile generation sessions spec).

## What to build

Every agent container gets the generation toolchain at boot, deliberately rather than by inheriting host state (the host Claude-directory mount is already retired, so containers currently have no plugins at all):

- The `gh` CLI is present in the agent container image.
- Container start installs the **latest** mattpocock-skills plugin — no pinning anywhere, by decision: the VPS session is an extension of the laptop, which tracks upstream at user scope.
- The resolved plugin version is logged visibly into the task feed (and run ledger where a run exists) as the forensic trail for "what version ran?".
- Install failure fails the task fast and visibly before the agent runs — a generation session silently degraded to freeform chat is worse than a retryable boot error.
- One mechanism for every container kind (interactive, implement, review, triage), which also closes the standing gap where `workflow:<skill>` labels name skills the container doesn't have.

## Acceptance criteria

- [ ] A task's feed shows the resolved mattpocock-skills version at session start
- [ ] The agent inside the container can invoke an installed skill
- [ ] A simulated install failure fails the task before any agent turn runs, with a visible error
- [ ] Generated container config/script tests assert the install step and version logging (container-manager script-content tests are the prior art)

## Blocked by

None — can start immediately.

lennons301:
Armed: `ready-for-agent` applied from the frontier of the Mobile generation sessions batch (decomposed from #50). Route: owner-confirmed in an interactive Claude Code session on 2026-08-03, under the standing arming authorization for interlude tickets (2026-08-02, interactive sessions only).

lennons301:
🤖 Attempt 1/3 starting.

lennons301:
🔧 Repair pass starting — merging main into PR #105 to resolve conflicts.

Issue #50, opened 2026-08-02
Mobile generation sessions — run the ticket-loop generation half from Interlude


lennons301:
## Problem Statement

The estate's ticket-loop workflow has two halves. The execution half is becoming location-independent: Phase 5 turns Interlude into an autonomous executor that implements, reviews and merges tickets unattended. The generation half — grilling an idea until its decisions are resolved, triaging a handwritten issue, turning agreement into a spec, decomposing a spec into tickets, and arming them — still runs only on the laptop, because that is where the mattpocock-skills plugin, an authenticated `gh`, and the repo checkouts live.

So when I'm away from the laptop, the pipeline stalls at exactly the point that most needs me: a fresh idea evaporates before it's captured properly, a new issue sits untriaged, and Phase 5's triage pass can tell me "this needs a grilling session" without there being any way to hold that session. The result is that autonomous capacity idles while generation waits for me to be at a desk.

## Solution

Interlude's interactive chat tasks (mobile-friendly PWA, persistent container, multi-turn) become **generation sessions**: from the phone, I start a session for a given GitHub issue — or freeform, for a fresh idea — and the agent in the container runs the same generation skills my laptop runs (`/grill-me`, `/grill-with-docs`, `/triage`, `/to-spec`, `/to-tickets`), at the same (latest) versions.

The governing principle is **extension of my laptop**: the session can complete the entire generation pipeline in one sitting — grill → decisions into CONTEXT.md/ADRs on a branch with its auto-created PR → spec → tickets published with dependency edges → tickets armed on my in-chat confirmation with an audit trail — with no "finish this on the laptop" step. The container is made capable of this by two additions: the skills plugin installed at container start (always latest, version logged), and a short-lived GitHub App installation token exposed per exec so `gh` works, exactly mirroring the existing git-credential pattern.

## User Stories

1. As the estate owner, I want to start a grilling session for a given GitHub issue from my phone, so that thinking work happens when I have time for it, not when I have a laptop.
2. As the estate owner, I want to start a freeform grilling session with no issue anchor, so that a fresh idea gets captured and stress-tested before it evaporates.
3. As the estate owner, I want to start a triage session for a given issue, so that I can move it through the label lifecycle from anywhere.
4. As the estate owner, I want the new-task flow to offer a session type and an optional issue picker, so that starting a session is a few taps, not typing a ritual incantation.
[...]

PR #103, opened 2026-08-06
Spike: does stream-json input expand slash commands? (#59)

### LINKED ISSUE #59: Spike: does stream-json input expand slash commands?
## Parent

## What to build

An empirical, documented answer — not production code. Interlude drives agent sessions by injecting user turns via stream-json input. It is documented that `claude -p "/skill-name"` expands user-invoked skills (including `disable-model-invocation: true` ones, which all generation skills are); it is **undocumented** whether a stream-json user turn does the same.

Test both turn positions against a real headless session:

1. **Seed turn** — first injected user message starts with a skill slash command.
2. **Mid-session follow-up turn** — a later injected message starts with one (this is where `/to-spec` and `/to-tickets` typically arrive in a generation session).

The outcome selects the seed-composition internals (blocked ticket): slash text passed through as-is, or the orchestrator inlining the skill content itself.

## Acceptance criteria

- [ ] Both turn positions tested against a real headless stream-json session with a user-invocable `disable-model-invocation: true` skill
- [ ] Result posted on this ticket with evidence (transcript excerpts showing the skill did or did not actually load)
- [ ] A stated recommendation for the seed-composition ticket: slash passthrough or inline composition, for each turn position

## Blocked by

None — can start immediately.

lennons301:
Closes #59

## What this is

The first spike for the **Mobile generation sessions** initiative (#50). It answers, empirically, whether an injected user turn whose text starts with a skill slash command (`/to-spec`, `/grill-me`, …) expands that skill in a headless Claude Code session — for both the **seed** turn and a **mid-session follow-up** turn. The outcome selects the internals of the blocked seed-composition ticket (#63).

Per the ticket, this is **an empirical, documented answer, not production code**. The deliverable is a findings doc: `docs/research/2026-08-06-spike-stream-json-slash-expansion.md` (new `docs/research/` home — the repo had none for spike/research notes).

## Method

A probe skill (`disable-model-invocation: true`, like every generation skill) whose SKILL.md body instructs the agent to emit an unguessable sentinel `Z7QX-PROBE-4K9M-LOADED`. The sentinel exists only inside SKILL.md, so its appearance in output *proves* the body loaded. Runs used the same flags Interlude uses (`--output-format stream-json --verbose --dangerously-skip-permissions`), with `--replay-user-messages` to capture what the CLI did with each injected turn.

## Result: yes — both positions, both transports

| Transport | Seed | Mid-session follow-up |
|-----------|------|-----------------------|
| `--input-format stream-json` (the ticket's literal ask) | ✅ expands | ✅ expands |
| text `-p` + `--resume` (Interlude's **actual** transport) | ✅ expands | ✅ expands |

Negative controls confirm the leading `/` is what drives it: the skill name without a slash → model doesn't recognise it (sentinel absent); a slash naming an uninstalled skill → `Unknown command`, turn ends (no silent degradation). The CLI rewrites `/spike-probe` into `<command-name>/spike-probe</command-name>` and the assistant's thinking block literally reasons *"the user has invoked the `/spike-probe` skill"*.

## Recommendation for #63: slash passthrough, both positions

The seed-composition pure function should emit the **bare slash string** for both the seed and follow-on turns. **No SKILL.md inlining is needed** — the "inline the skill content" fallback contemplated in #50/#63 can be dropped. Follow-on routing collapses to "pass the slash through".

Two things #63 must carry, both documented in the doc:
- **Transport discrepancy:** Interlude does **not** use `--input-format stream-json` today. `buildClaudeTurnCommand` (`src/lib/docker/container-manager.ts`) uses text `-p "$CLAUDE_PROMPT"` + `--resume`; "stream-json" there is the *output* format. Both transports expand slashes, so passthrough is robust either way.
- **Hard prerequisite #60:** `/name` only expands if the skill is installed at container start; an uninstalled `/name` fails loudly with `Unknown command`. Passthrough is only correct once #60 lands and the session verifies the install.

## Validation

Docs-only change (one new Markdown file). No justfile; `pnpm lint` is ESLint over TS/TSX (doesn't cover Markdown) and already fails on `main` with pre-existing baseline errors, and there are no changed code files for the test suite to exercise — so neither adds signal here.

## Self-review (`/code-review` vs `origin/main`, spec #59)

Both axes came back clean (no hard violations; the transport claim was verified against `container-manager.ts`). Acted on two findings: aligned the Reproduction SKILL.md `description` with the Method section, and added the `init`-event excerpt backing the "skill absent from the model tool list" claim. Two judgement-call findings I deliberately kept: the header's full-URL issue references (a standalone findings doc benefits from a clickable reference block, vs bare `#n` in prose), and some intentional restatement of the transport-discrepancy point across Method / its own subsection / Sources (each serves a different reader need).

The full result is also posted as a comment on #59 per its acceptance criteria. A separate `ticket-reviewer` with fresh eyes remains the approving gate.

### DIFF
--- a/docs/research/2026-08-06-spike-stream-json-slash-expansion.md
+++ b/docs/research/2026-08-06-spike-stream-json-slash-expansion.md
@@ -1,0 +1,206 @@
+# Spike: does stream-json input expand slash commands?
+
+- **Issue:** [#59](https://github.com/lennons301/interlude/issues/59) (decomposed from [#50](https://github.com/lennons301/interlude/issues/50), Mobile generation sessions)
+- **Selects internals of:** [#63](https://github.com/lennons301/interlude/issues/63) (Seed composition + follow-on slash routing)
+- **Date:** 2026-08-06
+- **Claude Code CLI:** 2.1.223
+- **Location note:** the repo had no `docs/research/` or `docs/spikes/` home for
+  empirical findings, so this file starts one. Design specs stay in `docs/specs/`.
+
+## Question
+
+Generation sessions (#50) drive the pipeline by injecting user turns whose text
+starts with a skill slash command — `/grill-me` at the seed, then `/to-spec`,
+`/to-tickets` as later turns. All generation skills are user-invocable and
+`disable-model-invocation: true`.
+
+It is **documented** that `claude -p "/skill-name"` expands such a skill. It was
+**undocumented** whether the same happens when the user turn arrives via
+`--input-format stream-json`, and whether it holds at a **mid-session follow-up**
+position, not just the first turn. The answer selects the seed-composition
+internals in #63: pass the slash text through as-is, or have the orchestrator
+inline the SKILL.md content itself.
+
+## Method
+
+An unguessable sentinel proves loading. A dedicated probe skill was installed as
+a personal skill (`~/.claude/skills/spike-probe/SKILL.md`), mirroring how the
+generation skills are user-invocable and `disable-model-invocation: true`:
+
+```markdown
+---
+name: spike-probe
+description: Internal spike probe for the stream-json slash-expansion experiment (issue 59).
+disable-model-invocation: true
+---
+... body instructs: your entire response must be exactly
+SPIKE_PROBE_OK sentinel=Z7QX-PROBE-4K9M-LOADED
+```
+
+The sentinel `Z7QX-PROBE-4K9M-LOADED` exists **only** inside SKILL.md. If it
+appears in the assistant output, the SKILL.md body was loaded into context — it
+cannot be guessed and is not in the model's training data.
+
+All runs: `--output-format stream-json --verbose --dangerously-skip-permissions`
+(matching Interlude's real invocation), `--model claude-haiku-4-5-20251001`
+(skill loading is a CLI concern, model-independent; haiku keeps it cheap), with
+`--replay-user-messages` on the stream-json runs to capture exactly what the CLI
+did with the injected turn. Raw transcripts were kept for every run.
+
+Two transports were tested, because the ticket's framing ("stream-json input")
+and Interlude's *actual* code diverge — see [Transport discrepancy](#transport-discrepancy-important-for-63):
+
+- **stream-json input** — the literal ticket ask: `claude -p --input-format
+  stream-json`, user turns piped as JSON objects.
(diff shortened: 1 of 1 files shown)


lennons301-reviewer (approved):
Approving. This is a `workflow:research` spike; the deliverable is a documented empirical answer, and it delivers all three acceptance criteria from #59.

**Acceptance criteria — all met**
- Both turn positions tested against a real headless session with a user-invocable `disable-model-invocation: true` probe skill: seed (row A) and mid-session follow-up (row B) on the `--input-format stream-json` transport, plus the same two positions on Interlude's actual text-`-p`+`--resume` transport (C/D).
- Result posted on #59 with transcript-excerpt evidence (the CLI rewriting `/spike-probe` into `<command-name>` form, the sentinel appearing in output, the init-event tool-list line).
- A per-position recommendation for #63: slash passthrough for both seed and follow-on, inlining dropped.

**What I verified independently**
- The one load-bearing source claim — that Interlude does not use `--input-format stream-json` and `buildClaudeTurnCommand` emits text `-p "$CLAUDE_PROMPT"` + `--output-format stream-json` (output only) + `--resume` on follow-ups — is accurate against `src/lib/docker/container-manager.ts` (lines 274-307).
- Methodology is sound: the unguessable sentinel lives only in SKILL.md, and the two negative controls (name without a leading `/` → not recognised; uninstalled `/name` → `Unknown command`) make the causal attribution to the leading slash airtight rather than assumed.
- Docs-only change (one new Markdown file), no gated paths under `docs/agents/review-gates.yaml` or the estate defaults, no code or config touched — low blast radius, correctly armed.

**Minor, non-blocking observations (suggestions, not required changes)**
- The probe was installed as a personal skill; production generation skills install via the mattpocock-skills plugin (#60). The CLI treats an installed skill uniformly regardless of source, and the doc correctly flags #60 install as a hard prerequisite, so the finding generalises — just noting the source difference wasn't exercised.
- The 'missing skill' control (`Unknown command`) was run only on the text transport, not on stream-json; given both transports expand identically in the installed case, the loud-failure conclusion is reasonable to carry over.
- The new `docs/research/` directory is a sensible, well-explained home for spike notes.

The recommendation is internally consistent with the evidence and honestly caveated on the #60 prerequisite. Comfortable landing this on main.

PR #105, opened 2026-08-10
Agent containers install mattpocock-skills at start

### LINKED ISSUE #60: Agent containers install mattpocock-skills at start

lennons301:
Closes #60

## What this does

Every agent container now gets the generation toolchain at boot, deliberately rather than by inheriting host state:

- **`gh` CLI in the image** (`Dockerfile.agent`) — installed from GitHub's own apt repository (canonical, arch-aware for amd64/arm64). Generation-session skills and the ticket-loop tooling drive the tracker through `gh`; the container's short-lived App token is already exposed per exec like git's.
- **mattpocock-skills installed at container start** — `buildSkillsInstallScript()` adds the `mattpocock` marketplace and installs `mattpocock-skills@mattpocock` at user scope, deliberately unpinned (always latest). It runs in the setup script's single `&&` chain for **every** container kind (interactive, implement, review, triage) — one mechanism, which also closes the gap where a `workflow:<skill>` label named a skill the container didn't have.
- **Fail-fast install** — a marketplace/clone/install error, or a "succeeded but resolved no version" install, aborts setup with a non-zero exit, which `execSetup` turns into a thrown error **before any agent turn runs**. A session never silently degrades to freeform chat.
- **Version logged as the forensic trail** — the resolved version is echoed behind a shared marker; `execSetup` parses it (`parseSkillsVersion`) and returns it. `startTask` writes it into the task feed and, where a run exists, onto the new `runs.skillsVersion` ledger column (first-write-wins, so it reflects the implement pass — mirroring how `model`/`effort` pin to the implement pass).

## Acceptance criteria

- [x] A task's feed shows the resolved mattpocock-skills version at session start (`mattpocock-skills plugin installed (vX.Y.Z)`)
- [x] The agent inside the container can invoke an installed skill — verified end-to-end: the plugin lands `enabled: true`
- [x] A simulated install failure fails the task before any agent turn runs, with a visible error — the install is a fail-fast gate on the setup `&&` chain; verified in a bash harness across three failure modes
- [x] Generated container config/script tests assert the install step and version logging (`buildSetupScript`, `buildSkillsInstallScript`, `parseSkillsVersion`, and a `Dockerfile.agent`-content test)

## Verification

Ran the exact commands end-to-end in an isolated config with an **empty** environment (no Claude auth):

- `claude plugin marketplace add`, `install`, and `list --json` all run **unauthenticated** — plugin management needs no OAuth token, so running it in the setup exec (which carries only `GIT_AUTH_TOKEN`) is correct.
- The plugin installed and resolved to the latest version, landing `enabled: true`.
- The fail-fast bash chain was exercised for success / empty-version / failed-install; a failed install exits with its own code and does **not** fire the misleading version-guard message (the guard is a self-contained `if`, not a `|| { exit 1; }` that would bind to the whole preceding chain).

`pnpm test` → 491 passing. `pnpm lint` is clean on all changed files (the repo's 16 pre-existing lint errors on `main` are untouched).

## Self-review notes (`/code-review` against `origin/main`)

Acted on the findings I agreed with: the ledger clobbering bug (now first-write-wins), a warn when a successful setup yields no marker, dropping `2>/dev/null` on the `list` call (stderr never reaches jq, so keeping it visible aids debugging), and tightening the version-capture regex against stray exec-frame bytes.

One finding I **did not** act on, by design: *"bake the plugin into the image to avoid a per-start network dependency."* This directly contradicts the ticket — #60 mandates install **at container start** (so it tracks upstream, "the VPS session is an extension of the laptop"), **always latest** (baking freezes the version), and **fail-fast** ("a generation session silently degraded to freeform chat is worse than a retryable boot error"). The uniform hard dependency across all container kinds is the ticket's "one mechanism" requirement.

## Migration

Adds `runs.skills_version` (`drizzle/0017_run_skills_version.sql`). Run `npx drizzle-kit push` (or the app auto-migrates on boot).


lennons301:

| gate | glob | path |
|---|---|---|
| data-migrations | `**/*.sql` | `drizzle/0017_run_skills_version.sql` |
| persistence | `drizzle/**` | `drizzle/0017_run_skills_version.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0017_snapshot.json` |

lennons301-reviewer (dismissed):
Approving (gated — the agent-sandbox + persistence globs matched, so a human still merges; this only satisfies branch protection).

## Verified
- **Objective signals green.** `pnpm test` → 491 passing; ESLint clean on all changed files; migration chain intact (0016 id == 0017 prevId); journal `when` bumped +1 day per the repo's drizzle convention.
- **Ticket #60 fully satisfied.** gh CLI added to the image (canonical arch-aware apt install); latest mattpocock-skills installed unpinned at user scope for **every** container kind via the shared setup `&&` chain; resolved version logged to the feed and (first-write-wins) onto `runs.skills_version`; install failure is a fail-fast gate before any agent turn.
- **Fail-fast logic is correct.** The `SKILLS_VERSION=$(... | jq ... // empty)` capture plus the self-contained `if [ -z ]; then exit 1; fi` guard covers both non-zero install and 'succeeded-but-no-version'; using a scoped `if` (not `|| { exit 1; }`) correctly avoids misattributing an earlier clone/checkout failure. `parseSkillsVersion` restricts the capture to `[\w.+-]+` so a stray exec-frame byte can't bleed into the version.
- Good call declining the self-review 'bake into the image' suggestion — it contradicts #60's install-at-start / always-latest / fail-fast mandate.

## Non-blocking observations (for the human merging)
1. The Dockerfile comment says the App token is 'exposed per exec exactly like git's.' The token reaches execs only as `GIT_AUTH_TOKEN`; `gh` authenticates from `GH_TOKEN`/`GITHUB_TOKEN`, neither of which is set in any exec env. #60 requires only that `gh` be *present* (satisfied), so this isn't a ticket gap — but if a follow-up expects skills to drive the tracker via authenticated `gh`, wiring `GH_TOKEN=$GIT_AUTH_TOKEN` is a loose end.
2. The feed message + ledger write in `turn-manager.ts` have no unit test; AC4 is met by the script-content tests (the ticket's cited prior art), but a small parse→feed→ledger test would harden it.

Neither warrants changes.

lennons301:

| gate | glob | path |
|---|---|---|
| data-migrations | `**/*.sql` | `drizzle/0018_run_skills_version.sql` |
| persistence | `drizzle/**` | `drizzle/0018_run_skills_version.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0018_snapshot.json` |

lennons301-reviewer (approved):
Approving. Reviewed with fresh eyes against issue #60 and repo standards. This PR is GATED (human-signoff label, auto-merge off) — approval satisfies branch protection only; a human still performs the merge.

## Ticket fit — all four acceptance criteria met

1. **Feed shows resolved version at session start** — `turn-manager.startTask` calls `insertSystemMessage(taskId, "mattpocock-skills plugin installed (vX.Y.Z)")` after `execSetup` returns the parsed version.
2. **gh CLI present in the image** — `Dockerfile.agent` installs `gh` from GitHub's own apt repo, arch-aware for amd64/arm64, keyring pinned the canonical way.
3. **Install failure fails the task before any turn** — `buildSkillsInstallScript()` is an `&&` operand of the setup chain, so a marketplace/clone/install non-zero exit aborts setup and `execSetup` throws before any agent turn. The self-contained `if [ -z "$SKILLS_VERSION" ]` guard also catches the 'succeeded but resolved no version' case, and is correctly scoped (not `|| { exit 1; }`) so it can't mislabel an earlier failure. Fail-fast, one mechanism for every container kind — the ticket's requirement, which also closes the `workflow:<skill>` gap.
4. **Tests assert install step + version logging** — coverage on `buildSetupScript` (install present, chained with `&&`), `buildSkillsInstallScript` (marketplace add, user-scope install, unpinned, marker echo, fail-fast guard), `parseSkillsVersion` (extraction, last-marker, null, trailing exec-frame byte), and `Dockerfile.agent` content.

## Objective signals
- `pnpm test` → 497 passing locally (CI reports no checks on the branch).
- ESLint clean on all changed TS files (the Dockerfile 'warning' is just ESLint not linting Dockerfiles).

## Verified specifics
- `jq` and `curl` are already in the image (`Dockerfile.agent` line 8 / 5), so the `list --json | jq` resolve step has its dependencies.
- Install is `--scope user`, which lands in the image-owned `/home/node/.claude` that survives unshadowed post-#28 — consistent with 'the VPS session is an extension of the laptop at user scope.'
- Ledger write is first-write-wins (`isNull(runs.skillsVersion)` guard), mirroring how model/effort pin to the implement pass — correct given the plugin is unpinned and may drift between passes.
- `parseSkillsVersion` restricts the capture to `[\w.+-]+` rather than `\S+`, defending against a Docker exec demux frame-header byte abutting the marker (the stream isn't demuxed) — a real, thoughtful edge case.
- Migration `0018_run_skills_version.sql` + journal entry are consistent; `when` is bumped +1 day past 0017 per the repo's drizzle convention. Gated correctly under agent-sandbox + persistence per docs/agents/review-gates.yaml.

## Non-blocking notes (not conditions of approval)
- The PR body's Migration section names `drizzle/0017_run_skills_version.sql`, but the committed file is `0018` (stale prose after the repair-pass merge of main). Cosmetic only; the journal and file agree.
- `SKILLS_MARKETPLACE_SOURCE` is exported but used only internally — harmless, and symmetric with the other exported constants used by tests.
- Runtime `gh` authentication is out of this ticket's scope (image presence only) and depends on the turn env, not this change.

The end-to-end runtime claims (plugin lands `enabled: true`, `plugin list --json` `.id` matches `mattpocock-skills@mattpocock`, plugin mgmt needs no OAuth) I cannot re-run here without Claude auth; they are documented in the PR verification and the surrounding mechanism is sound. Good, well-commented work that does what the ticket asked and only that.

Issue #106, opened 2026-08-10
Fleet dashboard: a run stuck non-terminal (no PR, no BLOCKED) renders as a permanent ghost 'running' card


lennons301:
## What happens

A run that never reaches a terminal status keeps rendering as a live **running** card on the fleet dashboard forever — even when the slot is genuinely free (`slots.used = 0`, no container). It shows as a phantom active session and is the "needs a human to clear it" class of stuck status (cf. the recent #138 clear-by-hand).

**Concrete instance (2026-08-10):** onboarding run for moontide `#32` (run `01KZN80CM28VGER9Z2GTT9FJFZ`). Its issue is closed, its slot is free, its only task is `status=completed` — yet it still appears under `running` in `/api/fleet/stream`.

## Root cause

The fleet's `running` cards are derived from **run** status — a run in `claimed` / `implementing` / `reviewing` (`src/lib/fleet/fleet-view.ts:~375`). The trigger is a run whose implement pass **finishes without producing a PR and without a detected `BLOCKED:`**: nothing in `decideNext` / the pass-completion path drives that run to a terminal status, so it dangles in `implementing`/`reviewing` indefinitely and the card never clears.

In the observed case the empty pass came from the `BLOCKED:` detector requiring the marker on the *very first line* (`src/lib/orchestrator/autonomy/blocked.ts:16`); the agent wrote a sentence of reasoning first, so it was **not** parked, the implement pass completed empty (no diff → no PR), and the run was left dangling. (That first-line brittleness is worth its own fix — filing separately.)

## Why the existing mechanisms don't cover it

- **PR #66 (closed #46)** fixed a *different layer* — a terminal **task** with a stale `container_status` being counted as a slot occupant. It filters `occupants` by task status. It does not touch **run** status, so it does not clear this.
- **Restart-recovery (#24, `src/lib/orchestrator/init.ts:48`)** only marks a run `interrupted` if it owns a task still in `running` status. A dangling run whose tasks are all `completed` is skipped — so a redeploy does **not** clear it either.

There is currently **no automatic reconciliation** for a run stuck non-terminal with no running task.

## Proposed fix

1. **Finalize the run at pass completion.** When an implement pass completes with no PR and no detected `BLOCKED:`, drive its run to a terminal status (e.g. `failed`/`abandoned`, or route `ready-for-human`) instead of leaving it in `implementing`/`reviewing`. This is the durable fix and belongs in the deterministic `decideNext` seam.
[...]

PR #108, opened 2026-08-10
Autonomy: detect BLOCKED: marker on any line, not just the first (#107)

### LINKED ISSUE #107: Autonomy: BLOCKED: marker only detected on the first line — a preamble silently defeats the HITL park
## What happens

The autonomous HITL escape hatch — an agent emitting `BLOCKED: <question>` to park its run and ask a human — is only detected when the marker is the **very first line** of the turn's final message. If the agent writes *any* preamble before it, detection returns `null`, the run is **not** parked, and the implement pass completes anyway (usually empty, no PR). The owner never gets the Discord question, and the run is left dangling (→ the permanent ghost card in #106).

**Concrete instance (2026-08-10):** moontide `#32`. The agent's final message was:

```
The ticket's premise doesn't hold, and the resolution changes the deliverable materially, so I'm stopping rather than guessing.

BLOCKED: `README.md` is not empty — it's a deliberate 89-line README ... Should I cut it down to the minimal three-section README the ticket specifies (deleting the existing content)?
```

The first line is the preamble, so `detectBlockedQuestion` returned `null`. The run was never parked; the pass completed with no diff and no PR; a follow-up answer posted to the task had nothing to un-park (the queue only delivers replies to an **idle in-memory container**, `src/lib/orchestrator/queue.ts:96`, and the container was already gone).

## Root cause

`src/lib/orchestrator/autonomy/blocked.ts`:

```js
const firstLine = finalMessage.split("\n", 1)[0].replace(/\r$/, "");
const match = firstLine.match(/^BLOCKED:\s*(.*)$/);
```

Only line 0 is examined. The header comment justifies the false-negative bias with "a missed marker merely idles the run where the dashboard shows it" — but per #106 a missed marker does **not** merely idle: it leaves the run stuck non-terminal forever. So the cost of a false negative is much higher than assumed, and "narrate, then BLOCKED:" is extremely common LLM behaviour.

## Proposed fix

- Detect `BLOCKED:` at the **start of any line** in the final message (not just line 0), taking the rest of that line (optionally plus following non-blank lines) as the question. This tolerates a leading preamble while keeping the false-positive guard: `^BLOCKED:` at a line start won't match prose like "we were BLOCKED: by X".
- Keep: empty question → `null`.
- Reinforce (belt-and-braces, not the fix) the agent/system prompt to put the marker on its own line — but detection must not depend on the model complying.

## Acceptance criteria

- [ ] Pure table-test (`blocked.ts`'s existing seam): a `BLOCKED:` line **after** a preamble paragraph is detected and returns the question — the moontide #32 message parks.
- [ ] A `BLOCKED:` token mid-line inside prose (not at a line start) does **not** trigger (no false positive).
- [ ] Empty/whitespace question still returns `null`.
- [ ] Once fixed, such a run parks (`status=blocked`) and posts the Discord question, and a reply un-parks it — end to end.

## Related

- #106 — the permanent ghost 'running' card this produces (a missed BLOCKED leaves the run dangling non-terminal).
- #138 — same "stuck status cleared by hand" class.

lennons301:
Closes #107

## What changed

`detectBlockedQuestion` (`src/lib/orchestrator/autonomy/blocked.ts`) now matches the `BLOCKED:` marker at the **start of any line** in the turn's final message, not only line 0. "Narrate, then `BLOCKED:`" is common LLM behaviour, so the old first-line-only bias let a leading preamble silently defeat the HITL park — the run was never blocked, the pass completed empty, the owner never got the Discord question, and the run dangled non-terminal forever (the #106 ghost card). The concrete instance was moontide #32.

The false-positive guards are kept: the marker must **begin a line** (mid-line prose like "we were BLOCKED: by X" never matches), it stays case-sensitive/unindented (`^BLOCKED:`), and lines inside a fenced code block are skipped. Only the rest of the marker's line becomes the question; further context stays in the task chat.

- `blocked.ts` — line-scanning detector with fence tracking; header comment rewritten to explain the new false-negative cost (#106) and the remaining guards.
- `workflow.ts` — belt-and-braces prompt reinforcement: ask the agent to keep the marker on its own line (a short lead-in is fine), instead of the now-false claim that a preamble hides it. Detection no longer depends on the model complying.
- `turn-manager.ts` — corrected the park-path comment ("leads with" → "on any line").
- Tests — `blocked.test.ts` updated to the new contract (incl. the real moontide #32 message, blank-first-line, and after-a-fence cases) and `decide.test.ts` re-pointed from "mid-message marker doesn't park" to "preamble-then-marker parks" + a genuine mid-line-prose negative.

## Acceptance criteria

- [x] Table-test: a `BLOCKED:` line **after** a preamble paragraph returns the question — the moontide #32 message parks (`blocked.test.ts`, `decide.test.ts`).
- [x] A `BLOCKED:` token mid-line inside prose does **not** trigger.
- [x] Empty/whitespace question still returns `null` (including after a preamble).
- [x] E2E park path unchanged and wired: `detectBlockedQuestion` → `decideNext` → `evaluatePassOutcome` → `parkBlockedRun` (status `blocked` + Discord question; a reply restarts the container and un-parks). The deterministic pieces are covered by tests; the live container leg can only be exercised on the VPS.

## Reviewer note (scope decision)

The issue named only the mid-line-prose false positive; I **additionally kept** the existing code-fence guard (skip lines inside ``` / ~~~ fences) rather than drop it. It preserves an existing test and prevents a real false positive — an agent quoting the `BLOCKED: <question>` protocol as a fenced example would otherwise park itself with a nonsense question. The trade-off is a narrow residual false-negative if an agent emits a genuine marker right after an *unterminated* fence; I judged the quoted-example false positive the more likely of the two. Happy to drop the fence guard if you'd rather track the issue's wording literally.

## Validation

- `pnpm test` — 500 passed (25 files).
- `pnpm exec eslint` on all changed files — clean.
- `tsc --noEmit` — no new errors in touched files (two pre-existing errors remain in the unrelated `container-manager.test.ts`).

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -151,7 +151,8 @@ chat/preview is never affected.
 ### 4. Answer a blocked agent from Discord
 
 When an agent hits a decision the ticket doesn't settle, it emits
-`BLOCKED: <question>` as the first line of its turn. The orchestrator parks the
+`BLOCKED: <question>` on its own line (a short lead-in above it is fine — the
+detector scans every line, not just the first). The orchestrator parks the
 run (status `blocked`, container kept alive but **holding no slot**) and posts the
 question to the project's linked Discord channel — or, if the project has none, to
 the fleet channel (`DISCORD_FLEET_CHANNEL_ID`).
--- a/docs/specs/2026-07-30-phase5-autonomous-ticket-loop-design.md
+++ b/docs/specs/2026-07-30-phase5-autonomous-ticket-loop-design.md
@@ -529,6 +529,11 @@ fleet channel. My reply becomes a queued user message on the task and the
 existing queue delivers it as the next turn — the Phase 4 idle-and-reply plumbing
 carries this whole feature.
 
+> **Superseded by #107/#106 (2026-08-10):** the detector now matches `BLOCKED:`
+> at the start of *any* line, not first-line-only. The false-negative rationale
+> above was overturned — a missed marker did **not** merely idle the task; it
+> left the run a permanent non-terminal ghost (#106).
+
 Supervised mode (from a `checkpoint:` directive) runs the implement pass
 normally, then forces `human-signoff` regardless of gate matches and pings me
 with the checkpoint text. It reuses the gate machinery rather than adding a
--- a/src/lib/orchestrator/autonomy/__tests__/blocked.test.ts
+++ b/src/lib/orchestrator/autonomy/__tests__/blocked.test.ts
@@ -7,13 +7,52 @@ describe("detectBlockedQuestion — positive cases", () => {
       detectBlockedQuestion("BLOCKED: Should retries be configurable or fixed at 3?")
     ).toBe("Should retries be configurable or fixed at 3?");
   });
-});
 
-describe("detectBlockedQuestion — negative cases (biased toward false negatives)", () => {
-  it("ignores the marker mid-message", () => {
+  it("detects the marker after a preamble paragraph — narrate-then-BLOCKED (issue #107)", () => {
+    // The real moontide #32 message: a preamble line, a blank line, then the
+    // marker. First-line-only detection returned null and the run dangled.
+    const moontide =
+      "The ticket's premise doesn't hold, and the resolution changes the " +
+      "deliverable materially, so I'm stopping rather than guessing.\n\n" +
+      "BLOCKED: `README.md` is not empty — it's a deliberate 89-line README. " +
+      "Should I cut it down to the minimal three-section README the ticket " +
+      "specifies (deleting the existing content)?";
+
+    expect(detectBlockedQuestion(moontide)).toBe(
+      "`README.md` is not empty — it's a deliberate 89-line README. Should I " +
+        "cut it down to the minimal three-section README the ticket specifies " +
+        "(deleting the existing content)?"
+    );
+  });
+
+  it("detects the marker after a single preamble line", () => {
     expect(
       detectBlockedQuestion(
         "I reviewed the ticket.\nBLOCKED: Should retries be configurable?"
+      )
+    ).toBe("Should retries be configurable?");
+  });
+
+  it("detects the marker preceded by a blank first line", () => {
+    expect(detectBlockedQuestion("\nBLOCKED: Should retries be configurable?")).toBe(
+      "Should retries be configurable?"
+    );
+  });
+
+  it("detects the marker after a preceding fenced code block", () => {
+    expect(
+      detectBlockedQuestion(
+        "Here is the diff:\n```\nsome code\n```\nBLOCKED: Should retries be configurable?"
+      )
+    ).toBe("Should retries be configurable?");
+  });
+});
+
+describe("detectBlockedQuestion — negative cases (false-positive guards)", () => {
+  it("ignores the marker mid-line inside prose", () => {
+    expect(
+      detectBlockedQuestion(
+        'The spec says to emit "BLOCKED: <question>" when stuck, which I did not need.'
       )
     ).toBeNull();
   });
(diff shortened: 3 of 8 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/blocked.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/blocked.ts` |

lennons301-reviewer (dismissed):
The code change is correct and I verified it end to end on the deterministic path: `detectBlockedQuestion` now scans every line, the fence guard is preserved, and the wiring (`decide.ts:551` → `evaluatePassOutcome` → park) is intact. Locally: the 3 touched test files pass (159 tests, incl. the real moontide #32 message, blank-first-line, and after-a-fence cases), and eslint is clean on all changed files. The fence-guard scope decision is well-reasoned and honestly flagged, and I'm content to keep it. Acceptance criteria 1-3 are met by tests; criterion 4's live leg is fairly noted as VPS-only.

One blocker, and it's the only reason I'm not approving: the PR changed the BLOCKED-marker contract in the code comment and the implement prompt, but left two other docs describing the *old* contract, so the repo is now internally inconsistent. Neither is mentioned in the PR body, so this reads as an oversight rather than a deliberate scope call.

- `docs/runbook.md:154` — 'emits `BLOCKED: <question>` as the first line of its turn.' This is the operator runbook CLAUDE.md explicitly tracks; it now contradicts the new prompt ('a short lead-in above it is fine'). Please update to 'on its own line (not necessarily the first line).'
- `docs/specs/2026-07-30-phase5-autonomous-ticket-loop-design.md:522-526` — still asserts the reversed rationale ('first line only, to bias toward false negatives — a missed detection merely idles the task'), which #106 and this PR's own header comment overturn. As a dated design record I'd accept a one-line annotation that #107 supersedes this, rather than a rewrite.

Cite: estate documentation-currency standard (keep docs current) — updating docs that describe a contract you changed is part of the change, not scope creep. Once the runbook is corrected (and the spec noted) this is an approve.

lennons301:
Addressed the review's one blocker (doc drift) in 3585026:
- `docs/runbook.md` — corrected to the new contract: `BLOCKED:` on its own line, a short lead-in above it is fine (detector scans every line).
- `docs/specs/2026-07-30-…` — kept the dated paragraph and added a superseded-by-#107/#106 annotation rather than rewriting it, per the reviewer's preference.

Contract now consistent across code, prompt, runbook and spec. Ready for re-review / human sign-off (PR is `human-signoff` gated, so a human merges).

lennons301:

| gate | glob | path |
|---|---|---|
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/blocked.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/blocked.ts` |

lennons301-reviewer (approved):
Approve — sound and scoped to #107.

Verified:
- Ran the three affected test files locally: 159 passed. Lint clean on all changed files. (No CI is configured on the branch.)
- `detectBlockedQuestion` now scans every line for `^BLOCKED:` (blocked.ts), tolerating a preamble; line-start guard keeps mid-line/quoted/indented markers from matching; empty question still returns null (incl. after a preamble).
- Acceptance criteria 1-3 are covered by real table-tests, including the actual moontide #32 message in both blocked.test.ts and decide.test.ts. Criterion 4 (live container leg) is inherently VPS-only; the deterministic wiring through decideNext is unchanged and tested.
- The kept-and-generalised code-fence guard is a reasonable, well-disclosed scope decision: it preserves an existing test and blocks a genuine false positive (agent quoting the protocol in a fence), against a narrow acknowledged residual false-negative for an unterminated fence.
- Docs (runbook, Phase 5 spec, blocked.ts header, turn-manager comment, implement prompt) all updated to the any-line contract and read coherently.

Non-blocking suggestions:
- decide.ts:542 comment still says 'A finished pass that leads with the blocked marker' — stale first-line-only wording; every other comment/doc for this contract was updated, this one was missed.
- Any-line detection slightly widens the false-positive surface (e.g. a summary line 'BLOCKED: none'); this is the design the ticket deliberately chose and the prompt mitigates it — noted only.

Gated PR: this approval satisfies branch protection; a human performs the merge.

Issue #61, opened 2026-08-03
Generation sessions exist as a class


lennons301:
## Parent

Decomposed from #50 (Mobile generation sessions spec).

## What to build

A generation session becomes a first-class concept the platform can see, without disturbing interactive-task semantics:

- Tasks gain a nullable session-skill field (`grill-me` | `grill-with-docs` | `triage` | `to-spec` | `to-tickets` | `wayfinder`). Task kind stays `interactive` — no run row, so spend-cap exemption and pickup-priority semantics hold by construction, with zero changes to spend accounting or the autonomy reducer.
- The task-creation API accepts the session skill (and an optional issue anchor reference passed through to the session).
- The fleet dashboard labels generation sessions distinctly from plain chat tasks (e.g. "Session · grill-me"), using the established fleet-dashboard design language.
- The autonomous triage pass's `kind=triage` is untouched and remains distinct from a human-driven triage session.

## Acceptance criteria

- [ ] A session created via the API records its session skill and appears on the dashboard labelled as a session, not a plain chat task
- [ ] Ordinary interactive tasks are unaffected (null session skill, unchanged display)
- [ ] No changes to spend exemption or reducer behaviour — sessions have no run row
- [ ] Migration generated and applied for the schema addition

## Blocked by

None — can start immediately.

lennons301:
Armed: `ready-for-agent` applied from the frontier of the Mobile generation sessions batch (decomposed from #50). Route: owner-confirmed in an interactive Claude Code session on 2026-08-03, under the standing arming authorization for interlude tickets (2026-08-02, interactive sessions only).

lennons301:
🤖 Attempt 1/3 starting.

PR #109, opened 2026-08-10
Per-exec GH_TOKEN for generation sessions

### LINKED ISSUE #62: Per-exec GH_TOKEN for generation sessions
## Parent

## What to build

Generation sessions can publish to GitHub — read and comment on issues, create issues with dependency edges, move labels, create wayfinder maps with sub-issues — because `gh` inside the container is authenticated with a short-lived GitHub App installation token:

- The token is minted per exec through the same App machinery as the existing per-exec git credential — one minting path to audit, not two.
- It is exposed to the exec environment only, never persisted in the container, its git config, or anything durable.
- **Only generation-session execs receive it.** Autonomous implement/review/triage execs never do — their GitHub side-effects stay in the orchestrator, per the Phase 5 architecture. This isolation boundary is the security decision: a token that can create issues can also apply labels (including the launch button), so no unattended kind may ever hold it.
- Preflight verifies the App installation's permissions cover everything the generation skills use: issue creation, comments, labels, native dependency edges, and sub-issue creation (wayfinder map/child operations).

## Acceptance criteria

- [ ] Inside a generation-session container, `gh issue view` and `gh issue create` succeed against the project repo
- [ ] Generated exec config tests assert token injection for generation-session execs and its **absence** for implement/review/triage execs (negative assertions, in the style of the existing no-token-embedded tests)
- [ ] No token value persists anywhere durable after the exec ends
- [ ] Preflight check covers issues, labels, dependency edges, and sub-issues, with a stated reason on failure

## Blocked by

- #61

lennons301:
Closes #62

Decomposed from #50 (Mobile generation sessions). Blocked-by #61 (merged), which
made a generation session a first-class class: an interactive task carrying a
non-null `sessionSkill`.

## What this does

**Per-exec `GH_TOKEN`, generation sessions only.** A generation session's Claude
turn now receives the same short-lived GitHub App installation token as `GH_TOKEN`
that `GIT_AUTH_TOKEN` already carries — so `gh issue view` / `gh issue create` and
the rest of the generation skills work inside the container. It is minted through
the one existing App path (`getInstallationToken`), exposed to the exec env only,
and never written to git config or anything durable — exactly mirroring the git
credential.

The isolation boundary is the security decision: injection is gated on
`isGenerationSession(task)` (the single definition of the concept, added to the
schema). Autonomous `implement` / `review` / `triage` / `repair` execs — and
ordinary chat tasks — receive no `GH_TOKEN` at all. A token that can create issues
can also apply the launch-button label, so no unattended kind may hold it.

**Preflight covers the permission generation needs.** The preflight now verifies
the App installation grants **Issues: write**. Every generation operation the
ticket lists — issue creation, comments, labels, native dependency edges, and
sub-issue creation — lives under that one permission (confirmed against GitHub's
fine-grained-permission docs), so a single check covers all four and its failure
names them in the stated reason. It is read from the installation's declared grant
(captured from the token mint), not probed by mutating an issue — deterministic and
side-effect-free — and fails closed like every other check.

## Acceptance criteria

- [x] Inside a generation-session container, `gh issue view` / `gh issue create`
  succeed against the project repo — the turn exec now carries `GH_TOKEN`.
- [x] Exec-config tests assert injection for every session skill and its **absence**
  for each autonomous kind (implement/review/triage/repair) and for a plain chat
  task, in the style of the existing no-token assertions.
- [x] No token value persists durably — `GH_TOKEN` is exec-scoped env, never
  written to disk/git config (same lifetime as `GIT_AUTH_TOKEN`).
- [x] Preflight covers issues, labels, dependency edges, and sub-issues, with a
  stated reason on failure.

## Self-review (`/code-review` vs `origin/main`, spec #62)

Acted on all four findings:
- Reused the cached token mint for the permission read (no second
  `apps.getInstallation` per project; removes a duplicated App-config guard).
- Made the token-absence test exhaustive over autonomous kinds (added `repair`).
- Reworded the preflight reason to be capability-framed rather than
  consumer-framed.

**One finding I deliberately scoped out:** the reviewer noted the Issues:write
check lives in the *autonomy* preflight, which only runs for autonomy-enabled
projects and does not itself gate generation-session *launch*. Extending the
existing preflight is what the ticket asks ("Preflight check covers…"); gating an
attended session's launch on autonomy infrastructure is a separate design decision
and out of scope here. A missing permission surfaces to the driver live as a `gh`
error, and Issues:write is a genuine prerequisite for the autonomous loop's own
issue/label operations too — so the check belongs in the shared preflight.

### DIFF
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -15,6 +15,22 @@ export const SESSION_SKILLS = [
 
 export type SessionSkill = (typeof SESSION_SKILLS)[number];
 
+/**
+ * A generation session is an interactive task running one of the estate's
+ * generation skills (issue #61): kind stays `interactive`, and the non-null
+ * sessionSkill is what distinguishes it from an ordinary chat task and from the
+ * autonomous `kind=triage` pass. This predicate is the single definition of the
+ * concept — issue #62 keys the per-exec `gh` token off it, so that a token able
+ * to create issues (and therefore apply the launch-button label) only ever
+ * reaches an attended session, never an unattended implement/review/triage exec.
+ */
+export function isGenerationSession(task: {
+  kind: string;
+  sessionSkill: SessionSkill | null;
+}): boolean {
+  return task.kind === "interactive" && task.sessionSkill !== null;
+}
+
 export const projects = sqliteTable("projects", {
   id: text("id").primaryKey(),
   name: text("name").notNull(),
--- a/src/lib/docker/__tests__/container-manager.test.ts
+++ b/src/lib/docker/__tests__/container-manager.test.ts
@@ -10,6 +10,11 @@ import {
   SKILLS_PLUGIN_ID,
   SKILLS_VERSION_MARKER,
 } from "../container-manager";
+import {
+  isGenerationSession,
+  SESSION_SKILLS,
+  type SessionSkill,
+} from "@/db/schema";
 
 // Capture the options passed to docker.createContainer so we can assert on the
 // HostConfig the orchestrator asks Docker for.
@@ -177,6 +182,7 @@ describe("buildTurnEnv", () => {
       prompt: "do the thing",
       gitAuthToken: "ghs_abc",
       claudeCodeOauthToken: null,
+      ghToken: null,
     });
     expect(env).toContain("CLAUDE_PROMPT=do the thing");
     expect(env).toContain("GIT_AUTH_TOKEN=ghs_abc");
@@ -187,6 +193,7 @@ describe("buildTurnEnv", () => {
       prompt: "p",
       gitAuthToken: "t",
       claudeCodeOauthToken: "sk-ant-oat01-xyz",
+      ghToken: null,
     });
     expect(env).toContain("CLAUDE_CODE_OAUTH_TOKEN=sk-ant-oat01-xyz");
   });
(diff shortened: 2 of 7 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| credentials | `src/lib/github/client.ts` | `src/lib/github/client.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/preflight.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/preflight.ts` |

lennons301-reviewer (approved):
Approving. Reviewed against issue #62 with fresh eyes; the change does what the ticket asks and only that.

## What I verified

**Isolation boundary (the security decision).** `isGenerationSession(task)` = interactive kind AND non-null sessionSkill, defined once in the schema and used at both exec sites — `startTask` (initial turn) and `processQueuedMessages` (follow-up turns). I confirmed `task` is the full `tasks` row at both call sites, so `kind`/`sessionSkill` are present. Every autonomous kind (implement/review/triage/repair, all kind != interactive) and a plain chat task (interactive, null skill) evaluate to false and therefore receive no GH_TOKEN. `execClaudeTurn`/`buildTurnEnv` have exactly one env-injection path, and it only pushes GH_TOKEN when a non-null token is passed. The autonomous `kind=triage` pass is correctly excluded (only the attended `sessionSkill=triage` interactive session gets a token).

**Token lifetime.** GH_TOKEN is the same App installation token as GIT_AUTH_TOKEN, minted through the one existing `getInstallationToken` path, injected into the exec `Env` only — never written to git config or anything durable. Same lifetime and same daemon-visibility profile as the existing git credential; no new persistence risk.

**Preflight.** `getInstallationPermissions` reuses the cached mint (the installation token already carries the full grant), so no second `apps.getInstallation` round-trip — by the time `hasIssuesWrite` runs, `getOctokit` has already populated the cache. It reads the declared `issues === "write"` grant (deterministic, side-effect-free) and fails closed on any read error. The failure reason names all four operations (issues, comments, labels, dependency edges, sub-issues), which all live under Issues:write.

**Objective signals.** `gh` is installed in Dockerfile.agent (prerequisite for the AC). Lint clean on all changed files; the two affected test suites pass (48 tests), including exhaustive positive assertions over SESSION_SKILLS and negative assertions over every autonomous kind + chat.

## Note for the human merging (non-blocking)

The Issues:write check lives in the autonomy preflight, which only runs for autonomy-enabled projects (`refreshAllPreflights` filters `autonomyEnabled === true`). Generation sessions are attended and can launch on non-autonomy projects, where this check never gates launch. The PR body calls this out deliberately: it satisfies the literal AC ("Preflight check covers..."), a missing permission surfaces live to the attended driver as a `gh` error, and the autonomous loop needs Issues:write regardless. I agree this is a reasonable scoping and out of scope for #62; flagging it only so the gap is a known one, not a surprise.

Minor doc nit (not worth a change): the `buildTurnEnv`/`TurnOptions` comments list "implement/review/triage" as the excluded kinds and omit "repair"; the predicate handles repair correctly regardless, and the test covers it.

PR #127, opened 2026-08-11
Reliability: a finalized run's owed review can never wedge pickup

### LINKED ISSUE #124: Reliability: a finalized run's owed review can never wedge pickup
## Parent

#115 — a starved/orphaned review pass silently halted the whole frontier (root cause chain and agreed design in the parent).

## What to build

A task must never outlive its run, and a dead run's task must never block new pickup.

- When an autonomous run is finalized — its PR merged or closed, or the run marked `failed` or `exhausted` — any task it owns that is still `queued` or `running` is terminalized (`cancelled`), through one shared helper called from every finalization point. (Runs still `reviewing`/`gated` legitimately have an owed review and are untouched.)
- The slot-reservation accounting that gates new claims (`claimableSlots`) counts only tasks whose run is still live. A queued task belonging to a finalized run never suppresses pickup.

This closes the exact incident: a human merged a gated PR, the run finalized to `merged`, its still-`queued` review task was orphaned, and that one dead task forced `claimableSlots = 0` (`no-slots`) on every sweep — halting all pickup for ~1.5h while the dashboard showed a free slot and ready tickets.

## Acceptance criteria

- [ ] A run finalized (merged/closed PR, `failed`, or `exhausted`) that still owns a `queued` or `running` review or implement task ends with that task terminalized; no task remains non-terminal under a terminal run.
- [ ] New-ticket pickup is not suppressed by a queued task whose run is terminal — the claim accounting ignores dead-run tasks.
- [ ] Regression test reproducing the incident shape: a gated run + open PR + owed review + the slot held by an interactive session, then the PR merges — the frontier continues to claim ready tickets rather than parking on `no-slots`.
- [ ] The existing self-heal (re-queue an owed review for a live `reviewing`/`gated` run with no in-flight review task) is preserved and still fires.

## Blocked by

lennons301:
Closes #124

## What this fixes

The LPS #135 incident: a human merged a gated PR whose review pass was still `queued` (the single slot was busy). The run finalized to `merged`, but its queued review task was **orphaned** — left non-terminal forever. That one dead task was counted by the slot-reservation accounting (`claimableSlots`), forcing `no-slots` on every sweep and **halting all new pickup for ~1.5h** while the dashboard showed a free slot and ready tickets.

## The fix — two defences

Both live in the DB-only, unit-testable `review-tasks` module.

1. **Write side — `cancelOrphanedRunTasks(db, runId)`** (`review-tasks.ts`). When a run reaches a terminal status, any pass it still owns that is non-terminal is set `cancelled` (an abandoned pass consumes no attempt — like a user-cancel or a restart-interrupted queued pass). It returns the cancelled tasks so the caller removes any container they held. Called from **every finalization point**:
   - sweep: `executeFinalizeRun` (merged/closed), `executeFailAttempt` (failed), `executeExhaust` (exhausted), via the `terminalizeFinalizedRunTasks` wrapper that also tears down containers;
   - turn-manager: `finishRun` (container error / user-cancel), so the invariant holds at the non-sweep terminal transitions too.

   A parked implement/repair task is released to `completed` by its own path first, so this only sweeps up genuine orphans. Runs still `reviewing`/`gated` (legitimately owed a review) are never touched.

2. **Read side — `queuedTasksReservingSlots(queued, liveRunIds)`** (`review-tasks.ts`). `gatherSnapshot` now filters the reservation counts through it: only a queued task whose run is still live (or a run-less interactive/triage pass) reserves a slot. So a dead-run task can never suppress a claim even if a finalization point were ever missed.

The existing gated→review self-heal is preserved — live runs are untouched, and `inFlightReviewTaskId` is unchanged.

## Acceptance criteria

- [x] A finalized run (merged/closed/failed/exhausted) leaves no owned task non-terminal — covered across the three sweep finalizers **and** turn-manager's `finishRun`, for every non-terminal status (`queued`/`running`/`blocked`).
- [x] Pickup is not suppressed by a queued task whose run is terminal — `queuedTasksReservingSlots` drops dead-run tasks from `claimableSlots`.
- [x] Regression test reproducing the incident shape — see below.
- [x] Existing self-heal preserved and still fires (asserted in tests).

## Tests

- `review-tasks.test.ts`: cancels queued/running/blocked owned tasks; leaves terminal, other-run, and live-run tasks alone (self-heal preserved); returns containers for cleanup; reserving-count drops dead-run tasks; an integrated DB-seam regression chaining finalize → orphan cancelled → reservation counts only the live run's review.
- `decide.test.ts`: with the orphan excluded (count 0), a free slot + ready ticket claims rather than parking on `no-slots`; a live-run queued review still reserves its slot (the count is not blanket-zeroed).

`pnpm test` green (564), lint clean on changed files, no new `tsc` errors (2 pre-existing `container-manager.test.ts` errors exist on `main`, unrelated).

## Self-review notes (code-review skill, fixed point `origin/main`, spec #124)

Acted on the findings I agreed with:
- **Turn-manager finalizers were uncovered** (AC-1 gap) → wired the helper into `finishRun`.
- **`blocked` was excluded** from the cancel filter → now swept too, so the invariant is total.
- **Return type name** `ReapedReviewTask` mis-described cancelled tasks of any kind → renamed to `TerminalizedTask` (deprecated alias retained for the reaper).

Where I made a judgement call rather than change code:
- **Duplication** with `reapDeadReviewTasks` is deliberate: the two share a shape but differ in the parts that matter (WHERE filter, terminal status `cancelled` vs `failed`, message) and encode distinct concepts (finalize-cleanup vs infra-death reaping); extracting a micro-helper would couple them and obscure intent.
- **AC-3 as one end-to-end test**: `gatherSnapshot` does live Octokit/Docker I/O, so a single merge→finalize→claim test isn't practical without mocking the world. Coverage is split across the DB-seam integrated test (write+read) and the `decideNext` claim test — the incident is reproduced across the seam.

### DIFF
--- a/src/lib/orchestrator/autonomy/__tests__/decide.test.ts
+++ b/src/lib/orchestrator/autonomy/__tests__/decide.test.ts
@@ -781,6 +781,41 @@ describe("decideNext — pause reasons", () => {
     expect(pauses(decideNext(makeSnapshot()))).toEqual([]);
   });
 
+  it("does not park on 'no-slots' when the only queued review belongs to a finalized run (issue #124)", () => {
+    // The LPS #135 wedge: a gated PR was merged by hand, finalizing its run,
+    // but the run's review task sat `queued`. The snapshot builder now excludes
+    // that dead-run task from the reservation counts (queuedTasksReservingSlots
+    // — unit-tested in review-tasks.test.ts), so the count the reducer sees is
+    // 0. With a slot free and a ready ticket, the frontier claims it rather
+    // than parking on `no-slots` — the ~1.5h stall this fixes.
+    const actions = decideNext(
+      makeSnapshot({
+        slots: { total: 1, occupied: 0, occupants: [] },
+        queuedReviewCount: 0,
+        candidates: [makeCandidate()],
+      })
+    );
+
+    expect(actions.some((a) => a.type === "pausePickup" && a.reason === "no-slots")).toBe(false);
+    expect(claims(actions)).toHaveLength(1);
+  });
+
+  it("still reserves a slot for a live-run queued review (the count is not blanket-zeroed, issue #124)", () => {
+    // The mirror case: a review owed by a *live* reviewing/gated run legitimately
+    // reserves the free slot, so a new claim waits. Only dead-run tasks are
+    // dropped — a genuinely owed review is not.
+    const actions = decideNext(
+      makeSnapshot({
+        slots: { total: 1, occupied: 0, occupants: [] },
+        queuedReviewCount: 1,
+        candidates: [makeCandidate()],
+      })
+    );
+
+    expect(actions).toContainEqual({ type: "pausePickup", reason: "no-slots" });
+    expect(claims(actions)).toHaveLength(0);
+  });
+
   it("pauses with 'daily-cap' and announces it once when today's spend meets the cap", () => {
     const actions = decideNext(
       makeSnapshot({ todayAutonomousSpendUsd: 500, dailyCapUsd: 500 })
--- a/src/lib/orchestrator/autonomy/__tests__/review-tasks.test.ts
+++ b/src/lib/orchestrator/autonomy/__tests__/review-tasks.test.ts
@@ -2,7 +2,13 @@ import { describe, it, expect, beforeEach } from "vitest";
 import { eq } from "drizzle-orm";
 import * as schema from "@/db/schema";
 import { createTestDb } from "@/test/create-test-db";
-import { inFlightReviewTaskId, reapDeadReviewTasks, type Db } from "../review-tasks";
+import {
+  cancelOrphanedRunTasks,
+  inFlightReviewTaskId,
+  queuedTasksReservingSlots,
+  reapDeadReviewTasks,
+  type Db,
+} from "../review-tasks";
 import { decideNext, passOutcomeSnapshot, type AutonomySnapshot } from "../decide";
 
 const ISSUE_REF = "owner/repo#141";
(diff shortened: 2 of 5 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
Approving. Reviewed with fresh context against issue #124; this is GATED (human-signoff, auto-merge off) so a human still performs the merge.

## Verified against the ticket
- **AC1 (no task outlives a finalized run):** `cancelOrphanedRunTasks(db, runId)` is wired into every run-finalization point — `executeFinalizeRun` (merged/closed), `executeExhaust` (exhausted), `executeFailAttempt` (failed) via `terminalizeFinalizedRunTasks`, plus turn-manager's `finishRun` (container error / user-cancel). I traced all four run-terminal transitions (sweep.ts:1534/1750/1838, turn-manager `finishRun`) — each now cleans up. The two init.ts terminal transitions don't need it: `finalizeDanglingRuns` only finalizes runs whose tasks are already all-terminal (it `continue`s on any queued/running/blocked task), and `markInterruptedRuns` transitions to the non-terminal `interrupted` state and already cancels its queued tasks. The invariant is made total by also sweeping `blocked`, which the AC's 'no task remains non-terminal under a terminal run' wording requires.
- **AC2 (pickup not suppressed by a dead-run queued task):** `queuedTasksReservingSlots` is applied in `gatherSnapshot` before the per-kind counts are derived, so all five reservation counts (interactive/implement/review/repair/triage) now drop tasks whose run isn't in `ACTIVE_RUN_STATUS_SET`. `interrupted` is the only non-active non-terminal run status, and its queued tasks are cancelled at boot, so the filter doesn't over-exclude a legitimately-owed reservation.
- **AC3 (regression):** covered at the DB seam (finalize → orphan cancelled → reservation counts only the live run) plus a `decideNext` claim test. Not a single end-to-end test because `gatherSnapshot` does live Docker/Octokit I/O; the implementer's self-review notes this honestly and both halves of the wedge are exercised.
- **AC4 (self-heal preserved):** `inFlightReviewTaskId` and `awaitingReview` are untouched; a live `reviewing`/`gated` run's owed review is left `queued` and still counted — asserted in both test files.

## Correctness spot-checks
- Ordering holds: `releaseImplementContainer` sets the parked implement task `completed` (teardownTaskContainer) before the cancel sweep runs, so only genuine orphans are cancelled and the implement container isn't double-removed.
- `cancelled` (not `failed`) is the right ledger choice — an abandoned pass consumes no attempt, matching user-cancel and restart-interrupt handling.
- `containerId`/`containerStatus` cleared alongside the terminal status (consistent with `reapDeadReviewTasks` and issue #46); container removal delegated to the caller, keeping the module DB-only.

## Objective signals
- No CI configured on the branch; ran locally: orchestrator suite 360 passed (incl. the two changed test files, 144 passed); `eslint` clean on all five changed files; `tsc --noEmit` clean apart from the two pre-existing `container-manager.test.ts` errors on main (untouched file).

## Non-blocking observation
- In turn-manager's `finishRun` the return of `cancelOrphanedRunTasks` is discarded, so a cancelled sibling's in-memory `activeTasks` entry / container isn't proactively dropped — it relies on the 5-minute reaper (now eligible since the run is terminal and the task cancelled). The docstring calls this out and there's rarely a live sibling container in those paths, so it's an acceptable trade-off, noted for the human merger.

The deliberate duplication with `reapDeadReviewTasks` (different WHERE filter, terminal status, and message; distinct concepts) is a reasonable judgement call, as argued in the PR body.

Issue #115, opened 2026-08-11
Review pass silently dropped under VPS slot contention — ticket parks at human-signoff with no review


lennons301:
## Summary

A ticket's **review pass can be silently dropped** when the VPS slot is taken by an explicitly-queued interactive session, leaving the PR parked at `human-signoff` with **zero reviews**. The loop does not appear to re-enqueue the dropped review pass, so the ticket looks "handed off to a human" when in fact the mandated `ticket-reviewer` stage never ran. This violates the coordinator-verifies-side-effects invariant (a review must have *actually landed* — not just been narrated/assumed).

## Observed (real run)

Repo `lennons301/last-person-standing`, ticket **#135**, PR **#159** (opened 2026-08-11 ~10:52 UTC by the implement pass).

PR #159 timeline:

```
10:52:48  ready_for_review     (interlude-agent[bot])
10:53:11  labeled human-signoff (interlude-agent[bot])   ← 23s later
10:53:40  deployed             (vercel[bot])
```

- **Reviews on #159: 0.** No `VERDICT:` comment, no `lennons301-reviewer` review, no reviewer check-run. Only CI checks ran (Build/Lint/Type/Test/Vercel — all green).
- The `human-signoff` label is *legitimate on its own terms*: #159 changes `**/*.css`, `**/components/**`, `**/app/**/page.*`, which trip the `visual-ui` gate in `standards/review-gates.yaml`. That gate governs **landing** (a human merges instead of auto-merge) — it is **not** a substitute for the review pass.
- The reviewer is otherwise healthy in this repo: the five preceding PRs each got `lennons301-reviewer` reviews (counts 1, 2, 7, 1, 2). #159 is the anomaly — uniquely missing its review — which points at *this run* being interrupted, not the reviewer being unwired.

So: the changed-paths gate step ran and applied `human-signoff`, but the **separate review-pass invocation never executed**, and nothing re-queued it.

## Hypothesis (from the operator)

Interlude is single-VPS (`self-hosted-hetzner`, per `products/interlude.yaml`). The review pass is a *separate* `claude -p` invocation under the reviewer identity, dispatched after the implement pass. An **explicitly-queued interactive session took the VPS slot**, starving/preempting the review pass; it was then **not re-enqueued** — a silent failure. The `human-signoff` label landing independently (it's applied from changed-paths, not from a review result) is exactly why the ticket *looks* complete while the review is missing.

## Why this matters

[...]

PR #128, opened 2026-08-11
Reliability: bound the memory-admission probe so a hung Docker daemon can't wedge the queue

### LINKED ISSUE #125: Reliability: the queue can't silently wedge on a hung Docker probe
## Parent

#115 — the queue loop wedged silently during the incident (root cause in the parent).

## What to build

A hung Docker daemon connection must never freeze task dispatch.

The queue's per-poll memory-admission check calls the Docker daemon. During the incident the loop stopped dispatching anything for ~1.5h with the slot reading free and no error logged — the signature of that Docker call hanging, which blocks every poll before it can dispatch and silently stalls **all** dispatch, interactive included.

Bound the Docker admission probe with a timeout; on timeout, fail open (allow the start) exactly as it already does on a probe error, so an unresponsive daemon connection can never freeze dispatch.

## Acceptance criteria

- [ ] The admission probe cannot block a queue poll indefinitely — it resolves within a bounded time.
- [ ] On probe timeout, dispatch proceeds (fail-open), matching the existing on-error behaviour, and the timeout is logged.
- [ ] Task dispatch continues to function when the Docker connection is unresponsive — covered by a test that simulates a hanging probe.

## Blocked by

lennons301:
Closes #125

## What & why

The queue's per-poll memory-admission check (`checkMemoryAdmission`) calls the Docker daemon (`docker.info()` + `listContainers()`) with **no timeout**. A hung daemon connection therefore blocks every poll before it can dispatch, silently stalling **all** dispatch — interactive included — with the slot count reading free and no error logged. This is the signature of the ~1.5h dispatch stall during the 2026-08-11 incident (parent #115).

## Change

- **Bound the probe** (`src/lib/orchestrator/capacity.ts`): race it against `ADMISSION_PROBE_TIMEOUT_MS` (5s). On timeout, **log and fail open** (allow the start), exactly as it already does on a probe *error*, so an unresponsive daemon connection can never freeze dispatch. The Docker gathering is split into `probeMemoryAdmission`; `checkMemoryAdmission` gained injectable `probe`/`timeoutMs` params (defaulted) so a test can simulate a hanging probe with no real daemon. Both call sites (`queue.ts`, `turn-manager.ts`) keep calling it argument-free and inherit the bound.
- **Reserve the slot before probing** (`src/lib/orchestrator/queue.ts`): the probe can now block up to 5s — longer than the 2s poll interval — so an overlapping poll could otherwise pick the same still-queued task and dispatch it twice (two containers for one task, the exact overcommit #93 guards against). Reserve the task in `processingTasks` before the probe and release it if the probe refuses. This also caps in-flight probes to one per free slot, since a full box short-circuits without probing.

## Acceptance criteria

- [x] The admission probe cannot block a queue poll indefinitely — it resolves within `ADMISSION_PROBE_TIMEOUT_MS`.
- [x] On probe timeout, dispatch proceeds (fail-open), matching the existing on-error behaviour, and the timeout is logged.
- [x] Task dispatch continues to function when the Docker connection is unresponsive — covered by `checkMemoryAdmission` tests that simulate a hanging probe (and a resolving/rejecting probe).

## Validation

- `pnpm test` — 560 passed (3 new).
- `pnpm lint` on changed files — clean.
- (Repo has 2 pre-existing `tsc` errors in `src/lib/docker/__tests__/container-manager.test.ts`, untouched here; and a 16-error lint baseline on `main` — neither in changed files.)

## Self-review notes (`/code-review` vs origin/main, spec #125)

Acted on:
- **Double-dispatch under overlapping polls** — fixed via the slot reservation above; this was the main risk the 5s bound introduced.
- **Abandoned/leaked hung probe per poll** — largely mitigated by the same reservation: after the first fail-open dispatch fills the slot, subsequent polls short-circuit *before* probing, so hung requests don't accumulate every 2s.
- **Test hygiene** — `console.error` spies now restored in an `afterEach`, so a failing assertion can't leave it stubbed for the rest of the suite.

Considered, not changed (rationale):
- **Boot-time `getCapacity()` outside the bound** — `initOrchestrator` derives capacity (`getCapacity`) before `startQueue`, so `_capacity` is cached before any poll. The unbounded first-call path only bites if the daemon is dead *at boot*, when init itself hangs and nothing could dispatch anyway — a different failure from the mid-operation hang this ticket targets.
- **Fail-open bypasses the overcommit guard under legitimate load** — the spec mandates fail-open on timeout; 5s is far above a healthy `info`+`listContainers` (tens of ms even under load), so this is a designed tradeoff, not a routinely-reachable one.

### DIFF
--- a/src/lib/orchestrator/__tests__/capacity.test.ts
+++ b/src/lib/orchestrator/__tests__/capacity.test.ts
@@ -1,5 +1,6 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import {
+  checkMemoryAdmission,
   createLocalCapacityProvider,
   deriveCapacity,
   wouldOvercommitMemory,
@@ -144,6 +145,40 @@ describe("wouldOvercommitMemory", () => {
     expect(
       wouldOvercommitMemory({ memTotalBytes: 4 * GiB, perAgentMemory, liveContainers: 1 })
     ).toBe(true);
+  });
+});
+
+describe("checkMemoryAdmission", () => {
+  // Issue #115: a hung Docker daemon connection must never freeze dispatch.
+  // The probe is raced against a bounded timeout and fails open on both a hang
+  // and an error, matching how it already fails open on a probe error.
+  it("fails open when the probe hangs past the timeout, and logs it", async () => {
+    const errSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    // A probe that never resolves — the signature of a hung daemon connection.
+    const hangingProbe = () => new Promise<{ ok: boolean }>(() => {});
+
+    const result = await checkMemoryAdmission(hangingProbe, 20);
+
+    expect(result.ok).toBe(true);
+    expect(errSpy).toHaveBeenCalledWith(
+      expect.stringContaining("timed out")
+    );
+    errSpy.mockRestore();
+  });
+
+  it("returns the probe's verdict when it resolves before the timeout", async () => {
+    const refused = { ok: false, reason: "2 live + one more overcommits" };
+    const result = await checkMemoryAdmission(async () => refused, 1000);
+    expect(result).toEqual(refused);
+  });
+
+  it("fails open when the probe rejects (existing on-error behaviour)", async () => {
+    const errSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    const result = await checkMemoryAdmission(async () => {
+      throw new Error("daemon unreachable");
+    }, 1000);
+    expect(result.ok).toBe(true);
+    errSpy.mockRestore();
   });
 });
 
--- a/src/lib/orchestrator/__tests__/capacity.test.ts
+++ b/src/lib/orchestrator/__tests__/capacity.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi } from "vitest";
+import { afterEach, describe, it, expect, vi } from "vitest";
 import {
   checkMemoryAdmission,
   createLocalCapacityProvider,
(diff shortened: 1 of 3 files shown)


lennons301-reviewer (approved):
Approving — verified against issue #125 with fresh eyes.

**Acceptance criteria (all met):**
- The admission probe is bounded: `checkMemoryAdmission` races `probeMemoryAdmission` (the Docker `info`+`listContainers` calls) against `ADMISSION_PROBE_TIMEOUT_MS` (5s) via `Promise.race` with a `Symbol` sentinel, so a hung daemon connection can never freeze a poll.
- On timeout it fails open (`{ ok: true }`) and logs, matching the existing on-error behaviour.
- New tests simulate a never-resolving probe (asserting fail-open + the timeout log), plus resolving-verdict and rejecting-probe cases.

**Correctness verified:**
- The double-dispatch risk the 5s bound introduces is handled: the task is reserved in `processingTasks` before probing and released only on refusal, so an overlapping 2s poll skips it via the existing `!processingTasks.has` guard — no two containers for one task (the #93 overcommit). Also caps in-flight probes to one per free slot.
- No unhandled rejection from a timed-out probe (`probePromise.catch(() => {})`); timer cleared in `finally` on all exit paths; `saturationLogged` log-once semantics preserved across the restructure.

**Objective signals:** full suite 560 passed (3 new); ESLint clean on all changed files; the only `tsc` errors are the pre-existing baseline in an untouched test file (`src/lib/docker/__tests__/container-manager.test.ts`).

**Gates:** no path in `review-gates.yaml` is touched (files are in `src/lib/orchestrator/` but not `autonomy/**`), so the armed state is correct. The fail-open-on-timeout behaviour is spec-mandated. Comfortable landing this on main.

PR #129, opened 2026-08-12
Observability: fleet surfaces & Discord-alerts stalled review + wedged pickup (#126)

### LINKED ISSUE #126: Observability: fleet surfaces & Discord-alerts stalled review + wedged pickup
## Parent

#115 — the stall was invisible (dashboard showed a free slot + ready tickets). Agreed observability design in the parent.

## What to build

Make silent pickup/review stalls loud. Add three fleet-health signals, each surfaced as a needs-you card on the fleet dashboard **and** a one-time Discord fleet ping (deduped like the existing cap/saturation announcements, so it fires once per occurrence — not every sweep):

- **(a) Owed-review stalled** — a run awaiting review (`reviewing`/`gated`) whose review pass has not started for longer than a threshold (~30 min, configurable). The card names the PR and the reason (e.g. the slot is held by an interactive session).
- **(b) Pickup wedged** — pickup is paused / there are no claimable slots while a slot is actually free (`occupied < total`), or a task has sat `queued` while a slot was free beyond a short threshold (~3 min).
- **(c) Stale queue heartbeat** — the queue poll loop records a last-progress timestamp; warn if it goes stale beyond a threshold (~2 min, ≈60 missed 2s polls).

## Acceptance criteria

- [ ] Each condition produces a needs-you card with a clear cause and actionable context, and fires a single Discord fleet ping (not repeated every sweep).
- [ ] The owed-review card clears automatically once the review lands or the run finalizes.
- [ ] The pickup-wedged signal fires for the exact incident shape (a task queued while a slot is free) within minutes, and clears when pickup resumes.
- [ ] The heartbeat detects a queue loop that has stopped making progress (and does not false-alarm on a healthy idle loop).
- [ ] Thresholds are configurable, with the stated defaults.

## Blocked by

lennons301:
Closes #126

Implements the observability leg (B + C) of the parent #115 design: make silent pickup/review stalls **loud** instead of invisible (the incident hid a wedged frontier for ~1.5h behind a dashboard that showed a free slot and ready tickets). The reliability legs (#124/#125) already shipped as #127/#128.

## What this adds

Three fleet-health signals, each surfaced as a **needs-you card** on the fleet dashboard **and** a **one-time deduped Discord fleet ping** (fires once per occurrence, not every 30s sweep, like the existing saturation/cap announcements):

- **(a) Owed-review stalled** — a `reviewing`/`gated` run whose review pass hasn't started for > threshold (~30 min). The card names the PR and the reason (e.g. "a slot is held by an interactive session"). Clears automatically when the review starts, lands, or the run finalizes.
- **(b) Pickup wedged** — a free slot (`occupied < total`) while either pickup is paused no-slots or a dispatchable task has sat queued, past a ~3 min debounce. Clears when pickup resumes.
- **(c) Stale queue heartbeat** — the 2s poll loop stamps a last-progress time at the end of each cycle; warn if stale > ~2 min. A healthy idle loop still completes a cycle every 2s so it never false-alarms; a hung dispatch path (an un-returned await, the #125 shape) never reaches the stamp.

Thresholds are configurable in minutes via `OWED_REVIEW_STALL_MINUTES` / `PICKUP_WEDGED_MINUTES` / `QUEUE_HEARTBEAT_STALE_MINUTES` (defaults 30 / 3 / 2), documented in `.env.example` and the runbook.

## Design

- The core is a **pure evaluator** (`src/lib/fleet/health.ts`, `evaluateFleetHealth`) over the current observations + prior memory (per-signal since-timers + already-announced flags) + thresholds. It returns the active signals (dashboard cards), the subset that just crossed threshold (Discord pings), and the next memory — so the card and the ping can **never disagree**. Fully unit-tested (`health.test.ts`).
- The sweep is the only caller: after it decides and acts, it gathers the input, runs the evaluator, records signals in a `globalThis` store (like the backlog / needs-human observations — the signals need live orchestrator state the DB alone can't provide), and fires pings for newly-crossed signals. Dedup memory is an in-memory module var beside the saturation/cap flags.
- `buildFleetView` stays pure — it gets `fleetHealth` in its rows and renders red cards, ordered right after the cap pause (a dead queue / wedged pickup halts the whole frontier, so it outranks any single run's question or sign-off).

## Tests

- `health.test.ts` — 15 evaluator cases: each signal's threshold, one-shot ping + dedup, auto-clear/re-arm, per-run independence, idle-loop no-false-alarm, queue-stopped/null-heartbeat guards.
- `fleet-view.test.ts` — health cards (queue-stale / pickup-wedged / review-stalled with & without PR link), and the updated cause-ordering assertion.
- `review-tasks.test.ts` — `runningReviewTaskId` (queued ≠ started, running = started, none).

`pnpm test` (594 pass) and `eslint` on changed files both green.

## Self-review (`/code-review` vs origin/main, #126 as spec)

Acted on:
- **Owed-review timer robustness** — sourced owed reviews from the DB run state rather than the GitHub-read-gated `awaitingReview`, so a transient PR-read failure can't reset the 30-min stall timer. Also a closer match to the spec wording.
- **Formatter drift** — the dashboard floored minutes while Discord rounded; unified on one shared `formatDuration`.

Noted but deliberately not changed:
- **Pickup-wedged uses one 3-min debounce for both sub-conditions** (paused-no-slots and queued-while-free). Intentional: `claimableSlots==0` with a free slot is *normal* for a single sweep when a claim is about to dispatch, so debouncing avoids flapping while still firing "within minutes" per the acceptance criteria.
- **Heartbeat also goes stale on a run of consecutive poll errors** — a genuine no-progress state, within the spirit of "the loop stopped making progress".
- **Health cards flow into the daily digest** via `buildFleetView().needsYou` (the digest renders needs-you generically) — an active stall at digest time is worth surfacing, so this is intended.

A separate `ticket-reviewer` with fresh eyes is still the approval gate.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -84,3 +84,10 @@ GITHUB_APP_INSTALLATION_ID=12345678
 # it NEVER enters an agent container. Canonical home is Doppler platform/prd,
 # mirrored into interlude/prd (rotation must update both).
 # REVIEWER_GH_TOKEN=github_pat_...
+
+# Fleet-health watchdog thresholds (issue #126), in MINUTES. Each surfaces a
+# needs-you card + one Discord fleet ping when a stall lasts past the threshold.
+# Defaults shown; a blank/invalid value keeps the default.
+# OWED_REVIEW_STALL_MINUTES=30      # a reviewing/gated run whose review never started
+# PICKUP_WEDGED_MINUTES=3           # a free slot but claimable work not dispatching
+# QUEUE_HEARTBEAT_STALE_MINUTES=2   # the 2s queue poll loop stopped making progress
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -120,16 +120,20 @@ The **dashboard is the home page** (`/`). It streams live over SSE and shows:
 
 - **slots** — total vs used, and what occupies each (autonomous / interactive).
 - **needs you** — blocked questions, `human-signoff` PRs, exhausted tickets, a
-  daily-cap pause, and failing preflights, each with a link.
+  daily-cap pause, failing preflights, and the fleet-health watchdog's stall
+  signals (issue #126: an owed review that never started, a wedged pickup, a
+  stale queue heartbeat), each with a link where one applies.
 - **running** — each active run's ticket, attempt (n/3), turn, spend vs budget,
   and phase (implement ▸ review ▸ merge).
 - **recent** — the last 7 days of completions.
 - **spend** — today's autonomous spend vs the $500/day cap.
 
 Discord is **push-only**: it tells you *when to look* (claimed, blocked question,
-sign-off needed, attempts exhausted, cap pause, slots saturated, daily digest).
-Autonomous success is deliberately silent — it shows on the dashboard. There is
-no `!status` command; the dashboard answers "what's happening".
+sign-off needed, attempts exhausted, cap pause, slots saturated, a fleet-health
+stall — owed review / wedged pickup / stale queue, issue #126 — daily digest).
+Each stall pings once per occurrence, not every sweep. Autonomous success is
+deliberately silent — it shows on the dashboard. There is no `!status` command;
+the dashboard answers "what's happening".
 
 ### 3. Pause pickup
 
@@ -263,6 +267,7 @@ Override with `CAPACITY_SLOTS`; per-agent memory with `AGENT_MEMORY_MB` (default
 | `DISCORD_FLEET_CHANNEL_ID` | Channel for fleet events + fallback for blocked questions when a project has no linked channel. |
 | `MAX_BUDGET_USD` | Per-attempt default budget ($20). |
 | `CAPACITY_SLOTS`, `AGENT_MEMORY_MB` | Override derived capacity — only when the derivation is wrong. |
+| `OWED_REVIEW_STALL_MINUTES`, `PICKUP_WEDGED_MINUTES`, `QUEUE_HEARTBEAT_STALE_MINUTES` | Fleet-health watchdog thresholds in minutes (issue #126). Defaults 30 / 3 / 2. |
 
 ### Labels
 
--- a/src/components/fleet/needs-you.tsx
+++ b/src/components/fleet/needs-you.tsx
@@ -10,6 +10,9 @@ const CAUSE_LABEL: Record<NeedsYouItem["cause"], string> = {
   exhausted: "exhausted",
   cap: "cap pause",
   preflight: "preflight",
+  "review-stalled": "review stalled",
+  "pickup-wedged": "pickup wedged",
+  "queue-stale": "queue stalled",
 };
 
 const CAUSE_TONE: Record<NeedsYouItem["cause"], "amber" | "red"> = {
@@ -20,6 +23,9 @@ const CAUSE_TONE: Record<NeedsYouItem["cause"], "amber" | "red"> = {
   exhausted: "red",
   cap: "red",
   preflight: "amber",
+  "review-stalled": "red",
+  "pickup-wedged": "red",
+  "queue-stale": "red",
 };
 
 /** Quiet confirmation sub-line: "No active runs · queue empty · autonomy on" */
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1,7 +1,20 @@
 import {
   ALLOWED_TICKET_EFFORTS,
   DEFAULT_ATTEMPT_BUDGET_USD,
+  DEFAULT_OWED_REVIEW_STALL_MS,
+  DEFAULT_PICKUP_WEDGED_MS,
+  DEFAULT_QUEUE_HEARTBEAT_STALE_MS,
 } from "./orchestrator/autonomy/budgets";
+import type { FleetHealthThresholds } from "./fleet/health";
+
+/** Parse an env value expressed in minutes into ms, falling back to a default.
+ * A blank, non-numeric or non-positive value keeps the default (a mistyped
+ * threshold must never silently become 0 and alarm every sweep). */
+function minutesEnvMs(raw: string | undefined, defaultMs: number): number {
+  if (raw == null || raw === "") return defaultMs;
+  const mins = parseFloat(raw);
+  return Number.isFinite(mins) && mins > 0 ? mins * 60_000 : defaultMs;
+}
 
 /**
  * Validate an `AGENT_EFFORT*` env value against the CLI's level set. Effort,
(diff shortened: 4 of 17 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
## Approve — gated, a human merges

Reviewed as a fresh pair of eyes against issue #126 and the repo's standards (CLAUDE.md, `docs/agents/review-gates.yaml`). This PR is **GATED** (`human-signoff`, auto-merge off): this approval only satisfies branch protection.

### Objective signals

No PR CI exists in this repo (`.github/workflows/deploy.yml` runs on push to `main` only), so I ran the checks locally on `agent/issue-126`:

- `pnpm test` — **594 pass / 27 files**, green.
- `eslint` on all 14 changed files — clean.
- `pnpm build` — succeeds (matters here: merging to `main` triggers the VPS deploy).
- `tsc --noEmit` — 2 errors, both in `src/lib/docker/__tests__/container-manager.test.ts`, a file this PR does not touch and identical to `origin/main`. **Pre-existing**, not attributable.
- Branch is 0 behind `origin/main` (merge-base = 01be7a1, i.e. #128 is included) and `MERGEABLE`.

### Acceptance criteria — all met

| AC | Verified |
|---|---|
| Each condition → needs-you card **and** a single Discord ping | One pure evaluator returns `signals` (cards) + `announce` (pings) from the same crossing, so the two can't disagree. Dedup memory (`fleetHealthState`) sits beside `saturationAnnounced` / `dailyCapAnnouncedDay`, matching the existing announce-once pattern. `health.test.ts` asserts card-persists-but-ping-is-one-shot for all three. |
| Owed-review card clears when the review lands or the run finalizes | Timers are re-seeded only from the current owed set, so a run leaving it drops both the timer and the announced flag (tested, incl. the re-arm with a fresh 30-min clock). |
| Pickup-wedged fires for the incident shape within minutes, clears on resume | `queuedWhileSlotFree` + `pausePickup(no-slots)` while `occupied < total`, 3-min debounce. I checked `decideNext`: the `no-slots` pause is only reachable after `if (eligible.length === 0) return actions`, so the signal needs genuinely armed work — it can't fire on an idle frontier. |
| Heartbeat detects a dead loop, no false alarm when idle | Stamped at the end of each cycle; `queueRunning` + null-heartbeat guards mean boot/shutdown never alarm (both tested). Idle cycles still stamp every 2s. |
| Thresholds configurable, stated defaults | `OWED_REVIEW_STALL_MINUTES` / `PICKUP_WEDGED_MINUTES` / `QUEUE_HEARTBEAT_STALE_MINUTES` (30/3/2), documented in `.env.example` + runbook env table. `minutesEnvMs` treats blank/NaN/≤0 as "keep the default", so a mistyped threshold can't become 0 and alarm every sweep. |

### What I checked most carefully

- **The owed-review derivation is sound, and the gated-PR false positive is avoided.** `gatherFleetHealthInput`'s DB filter (`reviewing|gated` + PR + `reviewResult == null` + `reviewVerdict == null` + `integrationCount < MAX_INTEGRATION_ATTEMPTS` + no *running* review task) mirrors `gatherReviewState`'s `awaitingReview` conditions minus the GitHub read. Crucially, a gated PR sitting on a human for hours has `reviewVerdict != null`, so it never reads as "review stalled"; and the spent-repairs conflict exclusion lines up exactly with `classifyGated`'s `conflict` precedence in `buildFleetView`, so no double-surfacing. Sourcing from the DB rather than `awaitingReview` is the right call — a transient PR-read failure can't reset the 30-min clock.
- **`runningReviewTaskId` vs `inFlightReviewTaskId`** is a real distinction, not a fork: the queued case must still block a duplicate (#45/#95) while counting as not-started here. Tested on both sides.
- **Read-model purity holds.** `buildFleetView` still takes everything via `FleetRows`; `health-store.ts` follows the documented `globalThis` rationale of `fleet/backlog.ts` / `fleet/needs-human.ts` verbatim, including "null until first observation".
- **Card ordering** (cap → machinery health → blocked → sign-off → exhausted → preflight) is justified in a comment and asserted in the updated ordering test; `CAUSE_LABEL`/`CAUSE_TONE` are exhaustive `Record`s so TS forced the component to keep up.
- **Failure containment:** the watchdog runs after `executeActions`, and a throw there is caught by the existing `runAutonomySweep().catch(...)` wrappers, so it cannot break the acting path. Each `notify*` swallows its own send errors.

### Non-blocking notes (suggestions, no change required to land)

1. **`src/lib/orchestrator/queue.ts`** — the heartbeat comment ("a cycle which hangs on an un-returned await (the #125 shape) never advances it") overstates the guarantee. Because `setInterval` keeps firing, the *next* cycle sees `processingTasks.has(next.id)`, short-circuits the dispatch block, and reaches the stamp — so a single hung admission probe leaves the heartbeat **fresh**. The heartbeat's real value is catching event-loop death / every-cycle-throwing; the #125 shape is covered by signal (b), or by (a) at 30 min when `CAPACITY_SLOTS=1` (the reserved probing task counts in `occupiedSlots`, so no slot reads free). Worth softening the comment so a future reader doesn't rely on a guarantee that isn't there.
2. **`repoTicket` in `fleet-view.ts`** — its comment says the card "has no run/project row to hand (the signal carries only a ref)", but `OwedReviewStall` carries `runId` and `runById` is in scope, so `runContext(run)` was available. Result is a card labelled `last-person-standing #135 · PR #159` next to siblings labelled with the *project name* + `attempt n/3`. Cosmetic inconsistency plus a third ref-formatting helper alongside `ticketLabel`/`issueUrl`.
3. **Store freezes when the sweep doesn't run.** `runAutonomySweep` returns early on the global kill switch or unconfigured GitHub, before evaluating — so a red stall card (and its frozen duration) persists until autonomy is re-enabled or the process restarts, and the queue heartbeat is unwatched while autonomy is off even though the interactive queue still runs. Same property as the backlog/needs-human stores, so consistent with convention; flagging so the runbook expectation is known.
4. **Pickup-wedged copy** tells the operator to check "a hung Docker daemon or a stuck poll loop", but a very plausible cause is the memory-admission probe refusing on low headroom (#93) — a parked container holds memory without holding a slot, which is exactly `occupied < total` + a task stuck `queued`. That reason lives in `queue.ts` and isn't surfaced; threading it through would make the card actionable in that case.
5. **Tiny:** `FleetHealthAnnounce` is field-for-field identical to `FleetHealthSignals` (could be an alias); and the DB-sourced owed-review filter is only reachable through the untested sweep — consistent with `gatherReviewState` being untested too, but a pure `owedReviewsFrom(runs, hasRunningReview)` helper would have been unit-testable alongside the evaluator.

Nothing above changes behaviour for the incident this ticket exists to make loud, and the smell baseline turns up nothing structural — the three near-identical `notify*` functions match the established shape of every other helper in `notifications.ts`, which the repo convention endorses.

Issue #130, opened 2026-08-12
Autonomy: a parked PR with failing checks is an invisible stall — the loop reads mergeability but never check status


lennons301:
## Problem

The loop never looks at a PR's **check status**. A parked run whose PR has red CI sits forever with a stale approval, no repair pass, no escalation, no needs-you card and no Discord ping — the exact invisible-stall shape #54 fixed for merge conflicts, but for checks.

`getPrState` (`src/lib/github/pull-requests.ts`) reads REST `pr.mergeable` only, and maps it to `MergeableState = mergeable | conflicting | unknown`. That field is **textual** mergeability. REST `mergeable_state` / GraphQL `mergeStateStatus` — which is `BLOCKED` when required checks fail — is never fetched, never stored, and appears nowhere in `AutonomySnapshot`. `grep` for `statusCheckRollup|checkRuns|ciStatus` across `src/lib/orchestrator/` and `src/lib/github/` returns nothing.

Consequence in `gatherReviewables` (`src/lib/orchestrator/autonomy/sweep.ts:490`): the sweep polls the PR every pass, but the only verdicts it can reach are `!pr.open` → settle, `conflicting` → repair, `mergeable` → carry on. A `mergeable` PR whose checks are failing falls through to `if (run.reviewVerdict != null) continue` — "waits on GitHub or on a human" — and nothing ever fires again.

For an **armed** run this is worse than a stall: GitHub's auto-merge silently never merges, and the run stays `reviewing` indefinitely with nobody told why.

## Observed instance

`lennons301/last-person-standing` PR #180 (issue #176), 2026-08-12:

- 09:29 implement pass pushed `d9d06fc`; 09:30 gate hit **visual-ui** → `human-signoff`, auto-merge left disarmed; 09:52 reviewer verdict **approve**.
- A human then clicked *Update branch* twice (`fc2a28b7` 10:15, `c327f5e1` 15:33) — both authored by `lennons301`, not by the loop (the issue timeline has no integration-repair comment).
- The 15:33 merge brought in a new caller from `main`, `src/components/picks/pick-table.tsx:267`, which passes the `opponentPreview` prop this PR deleted. Git merged clean — `mergeable: MERGEABLE` — but `tsc` fails:
  `error TS2322: ... is not assignable to type 'IntrinsicAttributes & TeamFormSheetProps'`
[...]

Issue #40, opened 2026-08-01
Deploy workflow needs a concurrency group — concurrent deploys took prod down


lennons301:
Incident 2026-08-01 ~20:17 UTC: PRs #39 and #38 merged 17s apart. Each push triggered its own Deploy run with no `concurrency` group on the workflow, so two `docker compose down → build → up` sequences ran simultaneously on the CX22. Two parallel Next.js builds on 2 vCPU / 4 GB starved the box: prod served nothing (HTTP 000) and even sshd stopped answering. Both runs hung in 'Deploy to VPS' until cancelled/timed out; recovery was a single manual redeploy once the box recovered.

**Fix (one-line, but `.github/workflows/**` is gated):**

```yaml
concurrency:
  group: deploy-production
  cancel-in-progress: false   # queue, don't kill a mid-flight deploy
```

**Worth considering in the same PR:** `cancel-in-progress: false` queues the second deploy rather than killing the first mid-`down` (which could strand the stack half-torn-down).

**Related but separate:** Phase 6's deploy drain mode (pause redeploys while runs are active); building on the box at all is the deeper cost — a prebuilt image registry would remove the build load entirely, but that's a bigger change than this fix needs.

**Noted during the same incident:** #25's verified-TLS health check did its job — the hung deploys were *visible* as hung rather than reporting success over a dead site.

lennons301:
Armed `ready-for-agent` under Sean's standing authorization (2026-08-02). Upstream context: #42's investigation identified this concurrency guard as option 1 of the real reboot-resilience fix — the restart-policy premise was wrong (see #42's findings). Human decision, agent applied the label.

lennons301:
🤖 Attempt 1/3 starting.

PR #133, opened 2026-08-13
Integration repair: verification misses typecheck/build, semantic merge breakage, and migration-file safety

### LINKED ISSUE #132: Integration repair: verification misses typecheck/build, semantic merge breakage, and migration-file safety
## Problem

`buildRepairPrompt` (`src/lib/orchestrator/autonomy/workflow.ts:257`) under-specifies what a clean integration repair means. Two gaps, both with a real failure behind them.

### 1. Verification stops short of the repo's actual checks

The operating rule is:

> `- Run the repo's tests and lint after resolving, and do not finish with either failing.`

Tests and lint. Not the type check, not the build. On `last-person-standing`, CI runs **Lint**, **Type Check**, **Test** and **Build** as separate jobs — so a repair pass can honestly satisfy this instruction while `tsc` fails. That is exactly the state PR #180 is in: Lint pass, Test pass, **Type Check fail** (`pick-table.tsx:267`, TS2322). The same prompt would also miss a broken `next build` on any repo whose type errors only surface there.

Related, and the actual cause of #180's breakage: the merge that broke it had **no textual conflict**. `main` had added a new caller of a prop the PR deleted. The prompt tells the agent to resolve conflicts and "not change anything the conflict did not require" — which reads as *don't look outside the conflict markers*, precisely wrong for the semantic case. A merge that compiles is the requirement; conflict-marker-free is not sufficient.

### 2. Migration files need an explicit rule (moontide#43)

Drizzle's migrator selects work by **timestamp, not filename or hash**:

```js
// drizzle-orm/pg-core/dialect.js
if (!lastDbMigration || Number(lastDbMigration.created_at) < migration.folderMillis) { /* apply */ }
```

So the natural way to resolve a migration-numbering collision — regenerate the losing migration — silently breaks every database that already applied it. This broke moontide PR #40's deploy: a migration already applied to stg as `0008` (`created_at=1786479960508`) was renumbered to `0009` with a fresh `when=1786481764114`, looked unapplied, re-ran `ADD COLUMN`, and died on Postgres `42701`.

The rule, from moontide#43:

- **Renaming the migration file is safe. Changing its `when` in `_journal.json` is not.** Preserve the original timestamp.
- The preserved timestamp must stay **greater than** the migration that now precedes it and **equal to** whatever any database already recorded.
- The opposite error is worse and quieter: a `when` stamped *earlier* than the preceding migration makes drizzle **skip the migration forever** on databases already past that point.
- Prefer idempotent DDL (`ADD COLUMN IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`) as defence in depth.
- A preview deploy of the same branch may already have applied a migration to a shared staging database, so "it's unmerged" does **not** mean "it's unapplied".

moontide#43 puts this in moontide's `AGENTS.md`, which is the right home for a human reading the repo — but the rule fires at merge-conflict time inside an autonomous pass, and the prompt is what that pass is actually steered by. Interlude uses Drizzle too, as does every other repo the loop is pointed at.

## What to build

Both changes are to the repair pass's operating rules in `buildRepairPrompt`, plus the same migration rule in the implement prompt (a fix-up cycle can regenerate a migration for the same reason).

1. **Verification = the repo's real checks.** Replace "tests and lint" with an instruction to run the checks the repo's CI runs — read `.github/workflows/` (or the `just`/`package.json` script the repo documents) and run all of them, naming type check and build explicitly as things that are commonly separate jobs. Finish only when they pass; if a check cannot be run in-container, say so in the summary rather than finishing silently.

2. **Semantic conflicts are in scope.** State that a conflict-marker-free merge is not the bar: after merging, look for callers, types and tests on the incoming side that the PR's own changes invalidate (a removed prop, a changed signature, a renamed export), and fix them in the same pass. Reword the "change nothing the conflict did not require" rule so it forbids unrelated work without forbidding this.

3. **Migration rule verbatim.** Add the five points above as their own block in the repair prompt, and the timestamp rule in the implement prompt.

## Acceptance criteria

- [ ] The repair prompt instructs the pass to run the repo's full CI-equivalent check set, with type check and build called out, and to report any check it could not run.
- [ ] The repair prompt states that semantic breakage introduced by a clean merge is in scope, with the removed-prop/changed-signature shape as the example.
- [ ] The repair prompt carries the migration rule: rename is safe, re-stamping `when` is not; preserved timestamp > preceding migration and == what any database recorded; too-early stamps are skipped forever; prefer idempotent DDL; unmerged ≠ unapplied because of preview deploys.
- [ ] The implement prompt carries the timestamp rule too.
- [ ] Prompt snapshot/unit tests assert each new rule is present.

## Related

- moontide#43 — the rule and the #40 deploy failure that produced it.
- lennons301/last-person-standing#180 — the type-check failure a clean merge produced.
- #54 — introduced the repair pass and this prompt.
- Siblings: the loop reads mergeability but never check status; a push to a reviewed PR leaves a stale approval. Both filed separately — #180 hit all three.

lennons301:
Closes #132

`buildRepairPrompt` under-specified what a clean integration repair means, in the two ways the ticket names — and last-person-standing PR #180 hit both.

## What changed

**1. Verification is the repo's real check set, not "tests and lint".**
CI on `last-person-standing` runs Lint, Type Check, Test and Build as separate jobs, so a repair pass could satisfy the old instruction honestly while `tsc` failed — exactly #180's state. The pass now reads `.github/workflows/` (or the `justfile` / `package.json` scripts the repo documents) and runs every check that gates a merge, with type check and build named as commonly-separate jobs, and reports in its summary any check it could not run rather than finishing silently on it.

**2. Semantic breakage from a clean merge is in scope.**
The merge that broke #180 had no textual conflict at all — `main` added a caller of a prop the PR deleted. "Change nothing the conflict did not require" reads as *don't look outside the conflict markers*, precisely wrong here. The prompt now states that a conflict-marker-free merge is not the bar (a tree that compiles and passes the checks is), names the shapes to hunt for (a new caller of a removed prop, a call site on a changed signature, an import of a renamed export, a test asserting replaced behaviour), and rescopes the change set to "the conflict resolutions plus the follow-on fixes the merged code needs" — forbidding unrelated work without forbidding those.

**3. The migration rule from moontide#43**, as a shared `MIGRATION_TIMESTAMP_RULE` constant carried by both the repair and implement prompts. Drizzle selects migrations by the journal's `when`, not by filename, so regenerating one to resolve a numbering collision re-runs its DDL on every database that already applied it (moontide PR #40 died on Postgres 42701 this way). The repair prompt adds the idempotent-DDL point on top, per the ticket's split.

`buildRepairPrompt` had no tests at all before this; it now covers the merge contract and each new rule.

## Self-review

Ran `/code-review` against `origin/main` with #132 as the spec. It found a **real bug in my own first commit**, now fixed in `ec7efef`:

> The implement block announced itself for a migration your work "adds", but "preserve the original timestamp" only makes sense for a rename or regeneration.

A freshly generated migration has no original timestamp and no database record pinning it — only ordering matters. This repo's journal is hand-stamped three days ahead of the wall clock (`0018` = `1786838400000` = 2026-08-16), so a migration generated today lands *below* its predecessor, and the rule as written would have told the pass to leave it there and be **skipped forever** — the exact failure its own third bullet warns about. The block now names the generated case explicitly.

Two more findings I agreed with and fixed:

- **CI discovery removed the verification floor.** This repo's `.github/workflows/` holds only `deploy.yml`, which gates no merge — so "run every check that gates a merge" alone verifies *nothing*, strictly weaker than the "tests and lint" it replaced. Tests, lint, type check and build are now the floor wherever the repo has them, with CI discovery on top.
- **"Finish only once they all pass" had no carve-out for a red base branch.** Interlude's own `main` is lint-red with a known pre-existing baseline, so a repair pass here would have been told both to make every check pass *and* not to touch unrelated files. It now judges each check against the base branch and reports pre-existing failures instead of fixing them.

Plus two test fixes (a `baseBranch` fixture of `"main"` couldn't distinguish interpolation from a hardcoded string — now `"develop"`; and four assertions duplicated verbatim across both prompts' describes now share `expectMigrationTimestampRule`), and a refresh of two comments left describing pre-#132 behaviour.

### Findings I did not act on

Each is real but outside "do what the ticket asks, all of it, and only it" — flagging rather than silently ignoring:

- **`next build` may OOM the agent container.** Verified: `DEFAULT_AGENT_MEMORY_MB = 1200` with swap disabled, and `deploy.yml` stops the app before rebuilding to free RAM on the 4 GB VPS. Worse, `turn-manager.ts` routes only `isImplementPass` to the interruption path, so a repair-pass OOM (exit 137) fails the whole run rather than being re-claimed. The ticket explicitly requires build be named, so I kept it — **but this deserves a follow-up**, either extending the interruption path to repair passes or having the pass prefer a cheaper type check over a full build.
- **The repair budget was not revisited.** `DEFAULT_REPAIR_BUDGET_USD = 5`, justified in its own comment as "a small mechanical turn". This diff adds CI discovery, four check suites and a semantic-breakage hunt. A budget-truncated repair still parks the PR for review. Worth a look, but changing a budget is its own decision.
- **"Never finish silently" has no machine-readable consumer.** The pass's summary isn't posted or parsed, so an unrunnable check changes nothing downstream. A proper fix (an `UNVERIFIED:` marker forcing `human-signoff`, mirroring `BLOCKED:`) is a feature beyond this ticket; the ticket asked only for the summary line.
- **The two halves of the preserved-timestamp rule can conflict.** If your branch generated `0008` at T1 and main later landed its own `0008` at T2 > T1, "preserve T1" and "must be greater than the migration that now precedes it" cannot both hold. The ticket asked for these five points *verbatim*, so I reproduced them — but the tension is real and moontide#43 may want to resolve it (plausibly: only a database that already recorded the migration pins the timestamp; absent that, restamp above the new predecessor).
- **"tests and lint" survives at four other sites** — the implement prompt, `workflowBlock`'s default branch, `buildFeedbackTurn`, and `docs/agents/review-pass.md`. #180's failure was actually produced by an *implement* pass, so the systemic fix is broader than the repair prompt. The ticket deliberately scoped this to `buildRepairPrompt`, and sibling #130 (the loop never reads check status) is the real fix for the class.
- **Idempotent DDL is not in the implement prompt**, which is the pass that authors the DDL. A fair argument, but the ticket explicitly split it: the five points in the repair prompt, the timestamp rule in the implement prompt.

## Validation

- `pnpm test` — 609 passed, 27 files
- `pnpm exec eslint` on the changed files — clean
- `pnpm exec tsc --noEmit` — two errors in `src/lib/docker/__tests__/container-manager.test.ts:360`, **pre-existing on `main`** and in a file this branch does not touch (diff vs `origin/main` is the two workflow files plus one comment in `turn-manager.ts`). Reporting rather than fixing it — which is the rule this PR just added to the prompt.

### DIFF
--- a/src/lib/orchestrator/autonomy/__tests__/workflow.test.ts
+++ b/src/lib/orchestrator/autonomy/__tests__/workflow.test.ts
@@ -13,6 +13,35 @@ const TICKET = {
   issueBody: "Make the frobnicator frob.\n\n## Acceptance criteria\n- it frobs",
 };
 
+/**
+ * The migration-timestamp rule (issue #132) is one shared constant carried by
+ * both the implement and repair prompts, so both assert it through here — a
+ * rewording cannot leave one prompt checked against the old text.
+ */
+function expectMigrationTimestampRule(prompt: string) {
+  // Renaming is safe; re-stamping `when` is not.
+  expect(prompt).toContain("Renaming a migration file is safe");
+  expect(prompt).toContain("`when` in `_journal.json` is not");
+  expect(prompt).toContain("preserve the original timestamp");
+
+  // The preserved timestamp is bounded on both sides.
+  expect(prompt).toContain("GREATER than the migration that now precedes it");
+  expect(prompt).toContain("EQUAL to whatever any database already recorded");
+
+  // The quieter failure: a too-early stamp is skipped forever.
+  expect(prompt).toContain("skip that migration forever");
+
+  // Unmerged does not mean unapplied — preview deploys share a staging DB.
+  expect(prompt).toContain('"It is unmerged" does not mean "it is unapplied"');
+  expect(prompt).toContain("preview deploy");
+
+  // A freshly generated migration has no timestamp to preserve — only ordering
+  // matters, and `drizzle-kit generate`'s clock stamp does not guarantee it.
+  expect(prompt).toContain("no original timestamp to preserve");
+  expect(prompt).toContain("NOT automatically greater than the entry before it");
+  expect(prompt).toContain("raise its `when` above that when it is not already");
+}
+
 describe("resolveWorkflowSkill", () => {
   it("resolves the vendored tdd workflow to its content", () => {
     const content = resolveWorkflowSkill("tdd");
@@ -173,24 +202,14 @@ describe("buildImplementPrompt", () => {
   describe("migration timestamp rule (issue #132)", () => {
     const prompt = buildImplementPrompt({ ...TICKET, workflow: { source: "default" } });
 
-    it("says renaming is safe but re-stamping `when` is not", () => {
-      expect(prompt).toContain("Renaming a migration file is safe");
-      expect(prompt).toContain("`when` in `_journal.json` is not");
-      expect(prompt).toContain("preserve the original timestamp");
+    it("carries the whole timestamp rule", () => {
+      expectMigrationTimestampRule(prompt);
     });
 
-    it("bounds the preserved timestamp on both sides", () => {
-      expect(prompt).toContain("GREATER than the migration that now precedes it");
-      expect(prompt).toContain("EQUAL to whatever any database already recorded");
-    });
-
-    it("names the quieter failure: a too-early stamp is skipped forever", () => {
-      expect(prompt).toContain("skip that migration forever");
-    });
-
-    it("warns that unmerged does not mean unapplied", () => {
-      expect(prompt).toContain('"It is unmerged" does not mean "it is unapplied"');
-      expect(prompt).toContain("preview deploy");
+    // The pass that authors a migration is the one that generates it, so the
+    // block must announce itself for that case and not only for a rename.
+    it("announces itself for a generated migration, not just a renamed one", () => {
+      expect(prompt).toContain("If your work generates, renames or regenerates");
     });
 
     it("places the rule in the pass's own instructions, before the ticket data", () => {
@@ -202,16 +221,19 @@ describe("buildImplementPrompt", () => {
 });
 
 describe("buildRepairPrompt", () => {
+  // Deliberately not "main": a fixture matching the common default cannot tell
+  // an interpolated baseBranch from one hardcoded in the prompt, and the value
+  // comes from the repo's own default_branch.
   const REPAIR = {
     repo: "acme/widgets",
     issueNumber: 7,
     prNumber: 41,
-    baseBranch: "main",
+    baseBranch: "develop",
   };
 
   it("merges the base branch — never rebase, never force-push", () => {
     const prompt = buildRepairPrompt(REPAIR);
-    expect(prompt).toContain("git merge origin/main");
+    expect(prompt).toContain("git merge origin/develop");
     expect(prompt).toContain("Never rebase and never force-push");
     expect(prompt).toContain("agent/issue-7");
   });
@@ -241,6 +263,24 @@ describe("buildRepairPrompt", () => {
     it("requires an unrunnable check to be reported, not skipped silently", () => {
       expect(prompt).toContain("cannot be run in this container");
       expect(prompt).toContain("Never finish silently");
+    });
+
+    // CI discovery is an addition to the floor, not a replacement for it: this
+    // repo's own .github/workflows/ holds only deploy.yml, so "run what CI
+    // gates" alone would verify nothing at all.
+    it("keeps an unconditional floor when CI declares no merge gate", () => {
+      expect(prompt).toContain(
+        "treat tests, lint, type check and build as the floor wherever the repo has them"
+      );
+      expect(prompt).toContain("CI only deploys can still be broken by your merge");
+    });
+
+    // Interlude's own main is lint-red; without this the pass is told both to
+    // make every check pass and not to touch unrelated files.
+    it("exempts failures the base branch already had", () => {
+      expect(prompt).toContain("Judge each check against the base branch");
+      expect(prompt).toContain("already present on origin/develop before you merged is not");
+      expect(prompt).toContain("Finish only once every check your merge broke is green again");
     });
   });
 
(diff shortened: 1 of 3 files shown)


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
Approving. GATED PR (`human-signoff`, auto-merge off) — this approval satisfies branch protection; a human still merges. Gate is correct: `src/lib/orchestrator/autonomy/**` matches the `autonomy-control` category in `docs/agents/review-gates.yaml`, and I have neither re-armed nor unlabelled it.

## Objective signals

No CI runs on PRs in this repo (`.github/workflows/` holds only `deploy.yml`), so I verified locally on `agent/issue-132`:

- `pnpm test` — 609 passed, 27 files
- `pnpm exec eslint` on the three changed files — clean (exit 0). Judged per-file against the known lint-red `main` baseline
- `pnpm exec tsc --noEmit` — two errors, both at `src/lib/docker/__tests__/container-manager.test.ts:360`. Confirmed pre-existing: `git diff --name-only origin/main...HEAD` is the two workflow files plus `turn-manager.ts`, and that file is untouched. Matches what the PR body reports
- Branch is level with `origin/main` (nothing behind), diff is 3 files / +243 −7

## Ticket acceptance criteria

All five verified against the built prompt strings, not just the diff:

1. **Full CI-equivalent check set** — the repair prompt reads `.github/workflows/` (or `justfile` / `package.json`), runs "every check that gates a merge", names `tsc --noEmit` and `next build` as commonly separate jobs, and requires an unrunnable check to be named in the summary ("Never finish silently"). ✔
2. **Semantic breakage in scope** — "A conflict-marker-free merge is NOT the bar", with the removed-prop / changed-signature / renamed-export / replaced-behaviour shapes named, and the old "Do not change anything the conflict did not require" replaced by a rule that bounds the change set to "the conflict resolutions plus the follow-on fixes the merged code needs". ✔ The test asserting `not.toContain` the old sentence is the right regression guard.
3. **Migration rule in the repair prompt** — all five moontide#43 points plus the idempotent-DDL bullet. ✔
4. **Timestamp rule in the implement prompt** — via the shared `MIGRATION_TIMESTAMP_RULE`, placed in the trusted operating-rules region ahead of the ticket markers (asserted by an index-ordering test, which is the right way to pin that). ✔
5. **Tests assert each rule** — `buildRepairPrompt` had none before; it now has 11, and the shared `expectMigrationTimestampRule` helper means a rewording cannot leave one prompt checked against stale text. ✔

Scope is clean: item 1 was ticket-scoped to `buildRepairPrompt`, and the implement prompt's own "tests and lint" line is correctly left alone. The `turn-manager.ts` hunk is comment-only.

## The one addition beyond the ticket, which I agree with

The sixth migration bullet (freshly generated migrations) is not in the ticket's five points, and I checked its premise rather than taking it on trust: this repo's `drizzle/meta/_journal.json` ends at `0018` / `1786838400000` = 2026-08-16, three days ahead of the wall clock. A migration generated today would be stamped *below* its predecessor, and the five points as written ("preserve the original timestamp") would have told the pass to leave it there — the skipped-forever failure the block's own third bullet warns about. Adding the carve-out is the correct call, and it is disclosed in the PR body rather than slipped in.

Same for the two other self-review fixes: the verification floor (this repo's CI gates no merge, so "run what CI gates" alone would have been strictly weaker than the "tests and lint" it replaced) and the base-branch carve-out (interlude's `main` is lint-red, so "make every check pass" + "touch no unrelated files" was an unsatisfiable pair). Both are reasoned deviations tied to real repo state, not taste.

Smell baseline: nothing to raise. The rule is one shared constant rather than two copies, the test duplication was already collapsed into a helper, and the `baseBranch: "develop"` fixture correctly distinguishes interpolation from a hardcoded `main`.

## Non-blocking, for the human merging

None of these should hold the PR — the first two are ticket-mandated and were disclosed by the implementer; I verified them and agree with the disclosure.

- **`next build` in a repair container.** Confirmed: `DEFAULT_AGENT_MEMORY_MB = 1200` with `MemorySwap = Memory` (swap off), and `turn-manager.ts:425` routes only `isImplementPass` to `interruptImplementPass` — so a repair-pass OOM (exit 137) falls through to the generic failure tail rather than being re-claimed, even though `isImplementShaped` exists two lines above at :150. That asymmetry is pre-existing, but this prompt makes it more likely to fire. Worth a follow-up ticket: extend the interruption path to repair passes, or let the pass prefer a type check when a full build is out of reach.
- **`DEFAULT_REPAIR_BUDGET_USD = 5`** is still justified in its own comment as "a small mechanical turn", which this diff meaningfully enlarges. Its own decision, but it should be taken.
- **Suggestion (mine, not tied to a standard):** "Judge each check against the base branch" tells the pass *what* to conclude but not *how* to establish the baseline, so a budget-pressured pass could assert "pre-existing" without checking. A half-sentence naming the mechanism — the base is already fetched, so a stash or a scratch worktree makes it cheap — would close that.

Issue #116, opened 2026-08-11
Spec: Fleet-parity UI upgrade + modern live chat interface (PRD)


lennons301:
> **PRD / spec** — produced by `/to-spec` from a grilling session. Not a single executable ticket; `/to-tickets` decomposes this into the implementable issues. Left unlabeled deliberately (no `ready-for-agent`).

## Problem Statement

Interlude's UI has split in two. The fleet dashboard (`/`) and the session-entry screen (`/tasks/new`) speak the "fleet" design language — IBM Plex Sans/Mono, the warm `fl-*` token palette, an instrument-panel vocabulary (eyebrows, chips, gauges, attempt pips, hairlines), lowercase mono navigation, and a system/dark/light theme. Every other screen is still the original shadcn/zinc scaffolding: `/settings`, the `/tasks` list, the global header, and — most importantly — the live task view at `/tasks/[id]`.

That live view is the sharpest problem. It renders agent output as plain text inside fixed-width zinc/purple chat bubbles with **no markdown**. Now that interactive generation sessions (`/grill-me`, `/to-spec`, `/to-tickets`, `triage`, `wayfinder`) run through this same view, the agent routinely emits long, structured, question-heavy responses — headings, numbered option lists with recommendations, code blocks, tables — and the current view mangles all of it. It is actively hostile to the one interaction the owner most needs to do well: read a grilling agent's structured questions and answer them.

The owner therefore experiences Interlude as two different products, and the half they increasingly live in (interactive sessions) is the least legible.

## Solution

Bring every remaining screen to parity with the fleet design system, and redesign the live task view into a genuinely modern, legible chat interface for interactive sessions.

Concretely, from the owner's perspective:

- Every page looks and behaves like one product: the same wordmark, navigation, type, palette, and theme control, flash-free on load, in system/dark/light.
- The live task view renders the agent's responses as a proper document — full-width markdown with real headings, lists, code and tables — while the owner's own messages stay as compact, scannable bubbles. Reading a grilling agent's options and typing an answer is comfortable on a phone and on desktop.
- The composer is a first-class input: multiline, slash-command aware, with clear "queued / working / idle" feedback and obvious ways to continue or complete the session.
- Tool activity is present but quiet — collapsed by default so it never drowns the conversation, expandable when the owner wants the detail.
- The `/tasks` list becomes a legible archive of every session and run, filterable, with interactive sessions kept visually distinct from autonomous runs.
[...]

Issue #123, opened 2026-08-11
Live view: preview pane fleet restyle


lennons301:
## Parent

#116

## What to build

The live preview pane restyled to the fleet palette, with its behavior otherwise unchanged. It reads as part of the fleet system (tokens, type) while keeping today's mobile-tabs / desktop-split layout and the subdomain proxying exactly as they are.

**Scope extended after the #117 review (see the comment below).** #117 moved the fleet theme into the root layout, so the live view now resolves to the light ground on a light-preference OS — but the live-view *container* chrome in `src/components/task-chat.tsx` is still hardcoded dark, and no other ticket in this milestone claims it. This ticket owns it: the mobile tab strip, the bottom bar, the desktop split divider, and the status dot / container label, alongside the preview pane itself. Styling only — the tabs/split behavior stays exactly as it is.

## Acceptance criteria

- [ ] The preview pane and its chrome render in the fleet palette/type
- [ ] The live-view container chrome renders in the fleet palette: mobile tab strip (`task-chat.tsx:156-172`), bottom bar (`:222-223`), desktop split divider (`:186`), status dot and container label (`:110`, `:236-246`) — no `zinc-*` class survives in `task-chat.tsx` or `preview-pane.tsx`
- [ ] The preview pane and the container chrome are verified in **both** dark and light themes; in particular the active mobile tab is legible on the light fleet ground (it renders `text-zinc-100` on `#eae5dc` today — roughly 1.1:1, effectively invisible)
- [ ] Mobile-tabs / desktop-split behavior is unchanged
- [ ] Subdomain proxying and hot-reload behavior are unchanged
- [ ] `pnpm build` and `pnpm lint` pass for changed files

## Blocked by

- #117 (shared shell + global theme)

lennons301:
### Carried over from the #117 review (PR #138)

#117 moved the fleet theme into the root layout, so the live view is no longer pinned dark — a light-preference OS now resolves it to the light fleet ground (`#eae5dc`) with no user action. The live view's hardcoded dark islands did not move with it.

**In scope for this ticket** — the preview chrome is `bg-zinc-950`, a dark island on parchment. The criterion "the preview pane and its chrome render in the fleet palette/type" covers it; check it in **both** themes.

[...]

Issue #121, opened 2026-08-11
Live view: markdown transcript read path (PoC gate)


lennons301:
## Parent

#116

## What to build

The read side of a modern live task view — the priority of this phase — so interactive sessions (`/grill-me` and the rest) are genuinely legible. A new pure `toChatView(messages)` mapper turns parsed stream messages into a discriminated view-model (`user-chip` / `agent-markdown` / `tool-event` / `system-note`); the React pieces are dumb renderers. Owner messages render as compact right-aligned chips; agent messages render full-width as a transcript with GFM markdown — real headings, lists, tables, and syntax-highlighted fenced code — sanitized so no raw HTML/script can execute, using a self-hosted pipeline (no CDN). Tool events render as compact, quiet rows collapsed by default, expandable to input/output, with a diff for edits. A subtle working pulse shows while the agent is thinking before text appears, and agent output streams in as the parser emits it. The view header shows task identity, branch, linked issue/PR and cost/status in fleet styling.

This ticket carries a proof-of-concept sign-off gate: first produce a small self-contained visual mock of the hybrid render (a realistic grilling transcript — heading, numbered options, a recommendation, a code block, and a short owner reply) in the real fleet tokens/fonts, in dark and light, and get owner sign-off. Build the full view only after sign-off — run this ticket as a supervised/checkpoint run for that reason.

## Acceptance criteria

- [ ] A visual PoC of the hybrid render (dark + light) is produced and signed off before the full build
- [ ] `toChatView` is a pure mapper with table-driven unit tests (user text, agent markdown, `tool_use` for Read/Bash/Edit, system, legacy plain-text)
- [ ] Owner messages render as compact right-aligned chips; agent messages render full-width with GFM markdown
- [ ] Fenced code is syntax-highlighted and scrolls horizontally without breaking layout; tables and nested lists render
- [ ] Raw HTML/script in agent output is rendered inert (sanitization test)
- [ ] The markdown/highlight pipeline is self-hosted (no CDN at build or runtime)
- [ ] Tool events are compact, collapsed by default, expandable; edits show a diff
- [ ] A working pulse shows before first output; agent output streams in
- [ ] The view header shows identity, branch, issue/PR, cost/status in fleet styling
[...]

PR #138, opened 2026-08-14
Foundation: shared AppShell + global fleet theme

### LINKED ISSUE #117: Foundation: shared AppShell + global fleet theme
## Parent

#116

## What to build

One shared application shell and a single global theme, so every screen reads as the same product. A shell component renders the mono `interlude / <section>` wordmark, the lowercase mono nav (fleet · new · tasks · settings), and the system/dark/light theme toggle (reusing the theme store currently local to the dashboard). The fleet fonts (IBM Plex Sans/Mono), the `.fleet` scope, the `fl-*` tokens, and the pre-paint theme script move into the root layout so the whole app is fleet-themed with no flash on first paint. The old shadcn `Header` is removed; the dashboard and the session-entry screen adopt the shared shell rather than hand-rolling their chrome. The full-height live view uses a slim shell variant (compact bar, back-to-fleet + task identity, no full nav). The change is additive: existing shadcn tokens remain available so screens not yet reskinned keep working until their own ticket lands.

## Acceptance criteria

- [ ] Every page renders the shared shell with the mono wordmark, nav, and a working theme toggle
- [ ] Nav indicates the current section
- [ ] Theme (system/dark/light) applies on every route, persists across navigation and reload, with no flash of the wrong theme or font on first paint
- [ ] The shadcn `Header` component is deleted and no longer referenced
- [ ] The dashboard and `/tasks/new` use the shared shell (no duplicated hand-rolled chrome)
- [ ] The live view uses the slim shell variant
- [ ] Not-yet-reskinned pages (`/settings`, `/tasks`) still render without visual breakage

## Blocked by

None — can start immediately

lennons301:
Closes #117

The app had split in two: the fleet dashboard and `/tasks/new` spoke the fleet design language and hand-rolled their own chrome, while `/settings`, `/tasks` and the live view wore the shadcn/zinc scaffolding under a separate global header. This is ticket 1 of the milestone — the foundation the reskin tickets build on.

## What changed

**The design system is now the app's ground, not one screen's.** The self-hosted Plex fonts, the `.fleet` token scope and the pre-paint theme script move into the root layout (`src/app/layout.tsx`), so every route is fleet-themed from the first frame. Geist is gone, and the shadcn `--font-sans`/`--font-mono` tokens point at Plex so the not-yet-reskinned screens stop rendering in a second typeface.

**One shell.** `AppShell` (`src/components/app-shell.tsx`) renders the mono `interlude / <section>` wordmark, the lowercase mono nav (`fleet · new · tasks · settings`) marking the current section with `aria-current="page"` plus a hairline underline, and the theme toggle. `/`, `/tasks/new`, `/tasks` and `/settings` all wear it; the dashboard and `/tasks/new` drop their duplicated chrome, and the shadcn `Header` is deleted. A page states its own `section` rather than the shell sniffing the pathname, which keeps the shell stateless (the theme toggle is the one client island).

**Slim variant for the live view.** `SlimShell` gives `/tasks/[id]` a compact bar — back-to-fleet, the task's identity, its live container status — and no nav, so chat and preview keep the vertical space. It owns the viewport height (`h-dvh`), replacing the old `h-[calc(100vh-4rem)]` guess at the removed header's height.

**Theme is a shared contract.** The system/dark/light store that lived inside the dashboard component (with its pre-paint script copy-pasted into two pages) is now `src/lib/fleet-theme.ts` + a `ThemeToggle` island, with a test that runs the pre-paint script the way a browser does.

## Additive, as the ticket asks

`/settings` and `/tasks` keep the shadcn tokens they render with today; only their chrome and container change. Their reskins (and the live view's) are the later tickets in the milestone.

## Findings from self-review I acted on

Ran `/code-review` against `origin/main` with #117 as the spec (commit `83c2e85`):

- **The theme didn't actually reach the un-reskinned screens.** The fleet palette follows `prefers-color-scheme`, but the shadcn tokens are chosen by a `dark` class that I'd pinned on — so on a light-preference OS (or with the toggle set to light) `/settings`, `/tasks` and the live view showed dark zinc cards on the light parchment ground. Those screens were coherent before, so that was breakage the shell introduced rather than the sanctioned transitional mismatch. The resolved theme now lands on both systems — `data-fleet-theme` and the `dark` class — in the pre-paint script, on every toggle press, and when the OS flips under a "system" choice.
- **Unguarded `localStorage`** in the toggle: it sits on every screen now, so a private-mode throw during render would take the whole client tree. Reads and writes are guarded; the toggle keeps working for the session when persistence is refused.
- **`<nav>` nested inside a `<span>`** — flow content in a phrasing-content element. Now a `div`.
- **PWA manifest colours** — I'd changed them to the fleet ground; reverted, since the ticket didn't ask for it and a static manifest can't follow the theme anyway.

## Findings I disagree with

- **"`FleetTheme`/`fleet-theme.ts` are misnamed now that the theme is app-wide."** The fleet design system *is* the app's design system as of this ticket (#116: "Adopt the existing fleet design system … across all remaining surfaces"). `fleet` names the language, not the dashboard, and the storage key has to stay `fleet-theme` regardless — renaming the module around a key that keeps the name would obscure more than it clarifies.
- **Duplicated `{accessory}<ThemeToggle />` tail in both shells** — one line in two sibling components; extracting a component for it would cost more than it saves.
- **Toggle server-renders as "system", so a pinned user sees the glyph correct itself on hydration.** Pre-existing behaviour of the dashboard's toggle and unavoidable without a cookie; the *theme* never flashes (the pre-paint script settles it), only the control's own label. Same for cross-tab drift (no `storage` listener) — neither is in this ticket's scope.

## Verification

- `pnpm build` passes; `pnpm test` 671 pass (17 new).
- `pnpm lint` is clean on every changed file. The repo-wide run still fails on the 16 pre-existing baseline errors on `main` (`custom-server.js`, `preview-pane.tsx`, `output-parser.test.ts`) — untouched here.
- Dev server checked on `/`, `/tasks/new`, `/tasks`, `/settings` and a live task page: each renders the shell with the right section marker, `aria-current`, the toggle, and the pre-paint script.
- Not automated (and not automatable within this repo's test culture — no `@testing-library`/Playwright, per #116): the visual read across surfaces, dark/light, and mobile ergonomics. Worth a look on the phone before merge.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/*.css` | `src/app/globals.css` |
| visual-ui | `**/components/**` | `src/components/app-shell.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/theme-toggle.tsx` |
| visual-ui | `**/components/**` | `src/components/header.tsx` |
| visual-ui | `**/app/**/page.*` | `src/app/tasks/new/page.tsx` |
| visual-ui | `**/app/**/page.*` | `src/app/tasks/page.tsx` |

lennons301-reviewer (approved):
**Approving.** This PR is gated (`human-signoff`, auto-merge off), so this only satisfies branch protection — a human still merges, and there is a visual call below worth their eyes.

## Objective signals

No CI is configured for PRs on this repo, so I ran everything locally on `agent/issue-117` (`83c2e85`):

- `pnpm test` — 30 files, **671 passed**, exit 0. Matches the PR's claim, including the 17 new theme tests.
- `pnpm build` — exit 0, all 5 routes compile.
- `npx eslint` on every changed file — **0 errors** (`globals.css` is simply outside the ESLint config). Consistent with the repo's documented practice of linting changed files against the 16-error `main` baseline.

## Acceptance criteria — verified

- **Every page renders the shared shell.** All five routes are covered: `/`, `/tasks/new`, `/tasks`, `/settings` directly, and `/tasks/[id]` via `TaskChat` → `SlimShell`. No page is left without chrome.
- **Nav marks the current section** — `aria-current="page"` plus a hairline underline, driven by the page-declared `section` rather than pathname sniffing. That is the better call: it keeps the shell out of the client graph and a new route can't silently inherit the wrong marker.
- **Theme applies everywhere, no flash.** I traced all four states through `globals.css` and the pre-paint script. `system`+dark OS → base `.fleet` + `dark` on; `system`+light OS → media-query light + `dark` off; pinned `dark` → `[data-fleet-theme=dark]` (and the light media block correctly excluded by its `:not(...)`); pinned `light` → `[data-fleet-theme=light]` + `dark` off. All four are self-consistent across **both** token systems. The second commit's fix — landing the resolved theme on the `dark` class as well as the attribute — was the right catch; without it the un-reskinned screens would have shown dark zinc on light parchment.
- **`Header` deleted** — confirmed gone, `grep` finds no remaining reference. Geist is likewise fully removed.
- **Dashboard and `/tasks/new` drop their duplicated chrome**, and the copy-pasted pre-paint script collapses from two inline string literals into one shared contract.
- **Live view uses the slim variant**, and `h-dvh` correctly replaces the old `h-[calc(100vh-4rem)]` guess at the now-removed header's height.

## What I liked

`src/lib/fleet-theme.ts` is the right seam: pure functions plus one small DOM write, so a server component can embed the script and the client island can share the vocabulary. The test that runs `THEME_PRE_PAINT_SCRIPT` through `new Function` with injected fakes deserves a mention — this repo has no jsdom and no testing-library, and rather than treat that as a reason to leave the no-flash contract untested, it found a way to exercise the real shipped string in a node environment. That is the honest version of testing here.

I also agree with all three of the self-review findings you pushed back on. `fleet` names the design language, not the dashboard; extracting a component for a one-line `{accessory}<ThemeToggle />` tail would cost more than it saves; and the toggle's glyph settling on hydration is `useSyncExternalStore` working as designed — the *theme* never flashes, only the control's own label, which is a different and much smaller thing.

## One finding — for the human's visual check, and for #121–#123

**The live view is now incoherent in light mode.** The body sits on `bg-fl-ground`, which follows the resolved theme, but the live view's remaining hardcoded islands don't follow anything. The sharpest case is the mobile preview tab strip in `src/components/task-chat.tsx`:

```
activeTab === "chat" ? "text-zinc-100 border-b-2 border-purple-500" : "text-zinc-500"
```

`text-zinc-100` (`#f4f4f5`) on the light fleet ground (`#eae5dc`) is about **1.1:1** — the active tab label is effectively invisible. The `bg-zinc-900 / text-zinc-100` composer in `message-input.tsx` and the `bg-zinc-950` preview chrome become dark islands on parchment, and the `text-blue-400` issue/PR links land at poor contrast.

Before this PR the live view was pinned dark, so this state was unreachable; now any user whose OS prefers light hits it with no action taken. I am **not blocking** on it: the ticket scopes its "no visual breakage" promise to `/settings` and `/tasks`, and I verified both of those are genuinely fine — they're token-based (`project-list.tsx`'s three `text-green-400` are the only hardcoded colours across either, and they read on both grounds). The live view's reskin is #121/#122/#123, next in this milestone. But this should be an explicit line item in those tickets rather than something discovered later, and it's worth a human glancing at a live task page in light mode before merging.

Minor, and only a suggestion: the `AppShell` header now carries wordmark + four nav items + toggle on one row. The toggle already hides its text label below `sm`, but the wordmark has `min-w-0` without `truncate`, so a long section name on a narrow phone could wrap. Worth including in the phone check the PR description already asks for.

## Smells

Nothing to raise. The `WIDTHS` map has three variants and all three are used, so it isn't speculative; `Section` is a proper union rather than a bare string; and the duplicated light palette in `globals.css` is pre-existing and already carries a keep-in-sync comment.

Issue #122, opened 2026-08-11
Live view: first-class composer


lennons301:
## Parent

#116

## What to build

The write side of the live task view — a first-class composer so answering a grilling agent is comfortable on phone and desktop. The composer is multiline and autosizes, uses Plex, sends on Enter and inserts a newline on Shift+Enter, and is slash-command aware (building on the existing follow-on slash routing). It gives clear feedback on whether the agent is working, is idle, or the message is queued, and offers obvious controls to continue the session or mark the task complete.

## Acceptance criteria

- [ ] The composer is multiline and autosizes as you type
- [ ] Enter sends; Shift+Enter inserts a newline
- [ ] Slash commands are usable/discoverable from the composer
- [ ] The composer shows clear queued / working / idle feedback
- [ ] Continue and complete controls are obvious and work
- [ ] Controls are keyboard-operable with visible focus
- [ ] The composer renders in the fleet palette/type — no `zinc-*` class survives in `message-input.tsx` — and is verified in **both** dark and light themes (it is `bg-zinc-900` / `text-zinc-100` today, a dark island on the light fleet ground after #117)
- [ ] `pnpm build` and `pnpm lint` pass for changed files

## Blocked by

- #121 (live view read path)

lennons301:
🤖 Attempt 1/3 starting.

Issue #119, opened 2026-08-11
Settings: fleet reskin + per-project autonomy control


lennons301:
## Parent

#116

## What to build

The settings screen brought to fleet parity and turned into the fleet's control room. Docker status and the project list are restyled to the fleet system (tokens, Plex, instrument-panel language). Each project shows whether autonomy is armed and its preflight verdict (passing/failing plus the reason when failing). The owner can arm or disarm a project's autonomy from here — arming requires a deliberate confirmation, and a project whose preflight is failing is not armable without an explicit override, with the failing reason shown. The per-project toggle uses the existing project `PATCH` endpoint (which already runs preflight on enable, fail-closed). The global kill switch is surfaced here too, using the endpoint from its ticket.

## Acceptance criteria

- [ ] Settings, docker-status and project-list render in the fleet system
- [ ] Each project card shows armed state and preflight verdict + reason
- [ ] Arming a project requires an explicit confirmation
- [ ] A failing-preflight project is not armable from the UI without an explicit override; the reason is shown
- [ ] Disarming works and takes effect
- [ ] The global kill switch is toggleable from settings
- [ ] The `canArm` predicate is unit-tested
- [ ] `pnpm build` and `pnpm lint` pass for changed files

## Blocked by

- #117 (shared shell + global theme)
- #118 (kill-switch endpoint)

lennons301:
🤖 Attempt 1/3 starting.

PR #140, opened 2026-08-15
Live view: preview pane fleet restyle

### LINKED ISSUE #123: Live view: preview pane fleet restyle

lennons301:
Closes #123

#117 moved the fleet theme into the root layout, so a light-preference OS now resolves the live view to the light ground (`#eae5dc`) with no user action — but the live view's hardcoded dark islands did not move with it. This restyles the preview pane and the live-view container chrome into fleet tokens and type. Behavior is untouched.

The headline case is the one the ticket calls out: the active mobile tab rendered `text-zinc-100` on the light ground, about **1.1:1** — effectively invisible, on the control you use to switch between chat and preview on a phone. It is now `text-fl-ink`: **12.9:1** light, **15.6:1** dark. The idle tab moves from `zinc-500` to `fl-ink-2` (4.6:1 / 7.3:1).

## What changed

**`preview-pane.tsx`** — toolbar, empty state, and the over-iframe loading/error messages. The toolbar controls are plain buttons rather than shadcn `Button`s: the ghost/outline variants resolve to the *greyscale shadcn* tokens, which don't track the fleet theme, so a shadcn `outline` button sits on parchment as a grey card. A message over the iframe gets its own opaque ground (`StatusNote`) instead of relying on a translucent scrim — whatever the dev server painted underneath isn't a background this palette controls.

**`task-chat.tsx`** — mobile tab strip, references block, desktop split divider, terminal footer, container label and status dot.

Selection on the tab strip is marked in ink, not colour. Colour in this system is semantic and which pane you're looking at says nothing about the run, so the old `purple-500` had no meaning to carry. The rule is heavier than the shell nav's hairline because this is a thumb-sized control, not a header link.

The running label loses `animate-pulse` and the dot gains `fleet-dot-live` — the fleet's one ambient animation, which unlike `animate-pulse` honours `prefers-reduced-motion`.

## Verification

`pnpm build` passes; 693 tests pass; `pnpm lint` adds no new errors (see the one caveat below). No `zinc-*` survives in either file, and no other hardcoded palette class either — the remaining `zinc` in the rendered live view comes from `task-stream.tsx` and `message-input.tsx`, which belong to #121 and #122.

**On "verified in both themes":** Chromium's system libraries (`libnspr4`/`libnss3`) aren't installed in this environment and sudo needs a password, so I could not take screenshots. Instead I verified the two things a screenshot would have caught:

1. **Every token class actually compiles.** All 14 `text-/bg-/border-fl-*` utilities used across both files resolve to a real rule against the right custom property in the built CSS, and each referenced `--fl-*` var is defined in all three theme scopes (`.fleet`, the `prefers-color-scheme: light` block, and `[data-fleet-theme="light"]`). A typo'd token — the realistic failure, since it renders as *nothing* rather than erroring — would have shown here.
2. **Contrast, computed from the tokens in `globals.css`** for every foreground/background pair in the diff, in both themes. Figures quoted above.

I also rendered `PreviewPane` through `react-dom/server` and diffed the markup, and seeded task rows to check the container chrome in the real SSR output.

A reviewer with a working browser should still eyeball it; the above is not a substitute for looking at it on a phone.

## Self-review

The `/code-review` pass against `origin/main` found a **real regression I'd introduced**, fixed in f3e2b82: I moved the running cue from the label's `animate-pulse` (keyed to `containerStatus`) onto the dot (keyed to task `status`), and those part company — `turn-manager.ts:349` sets `containerStatus: "idle"` at end of turn while the task stays `running`. An agent waiting on a follow-up, the exact "needs you" state, would have breathed forever. `live` is now its own prop off `containerStatus`. Verified both ways in SSR output.

Also acted on: `noopener` on the preview's `window.open` (it's agent-authored code); scrim back to 50% (I'd raised it to 80%, forgetting this overlay fires on *every* agent write via the 500ms reload — it would blink the preview out on each save); explicit heights on the toolbar/retry buttons (11px mono in a bare button is ~18px, and these are thumb targets); the running label to neutral ink (`text-fl-green` on the bare light ground is 3.8:1, under AA at 11px — the system's other green text sits on a wash or card, so precedent didn't cover it); `aria-current` on the tabs; and the system's `ActionLink` underline idiom on the issue/PR links.

**Findings I did not act on, deliberately:**

- **`FOCUS_RING` is 2.44:1 on the light ground**, under the 3:1 non-text minimum. This is real, but `FOCUS_RING` is a shared constant from #117 used app-wide — retuning it from a preview-pane ticket would silently restyle every screen. It deserves its own ticket, and I'd support one.
- **`StatusDot` duplicates `LiveDot`/`DOT_COLOR` in fleet-bits.** Reusing `LiveDot` would either change the header's appearance (it renders a text label beside every non-live state) or require extending its state union with task statuses — a shared-component change affecting the dashboard. The duplication is a 5-entry record and a span.
- **`ToolbarButton`/`StatusNote` could live in fleet-bits.** Fair, and #119/#120/#122 will likely each want the same quiet mono button — but promoting an atom on one usage is speculative. Better promoted when the second caller appears.
- **The `!devPort` empty state is unreachable**, since `task-chat.tsx` guards with `devPort &&`. True. Deleting it is a structural change beyond a styling ticket, and it keeps the component coherent standalone.
- **`pnpm lint` reports one error on `preview-pane.tsx:92`** (`react-hooks/set-state-in-effect`). It is **pre-existing** — byte-identical on `main`, same line and rule, and the diff doesn't touch that effect. Fixing it means restructuring an effect that drives the TLS pre-warm, which is real behavior risk in a styling ticket. Flagging rather than smuggling it in.

### DIFF
--- a/src/components/preview-pane.tsx
+++ b/src/components/preview-pane.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useCallback, useEffect, useRef, useState } from "react";
-import { Button } from "@/components/ui/button";
+import { FOCUS_RING } from "@/components/fleet/fleet-bits";
 
 interface PreviewPaneProps {
   taskId: string;
@@ -113,7 +113,7 @@ export function PreviewPane({
 
   if (!devPort) {
     return (
-      <div className="flex items-center justify-center h-full text-zinc-500 text-sm">
+      <div className="flex h-full items-center justify-center bg-fl-ground font-plex-mono text-[11px] text-fl-ink-2">
         {status === "stopped"
           ? "Dev server stopped"
           : "No dev server running"}
@@ -122,48 +122,40 @@ export function PreviewPane({
   }
 
   return (
-    <div className="flex flex-col h-full">
+    <div className="flex h-full flex-col bg-fl-ground">
       {/* Toolbar */}
-      <div className="flex items-center gap-2 px-3 py-2 border-b border-zinc-800 shrink-0">
-        <span className="text-xs text-zinc-500 font-mono">:{devPort}</span>
+      <div className="flex shrink-0 items-center gap-2 border-b border-fl-line bg-fl-surface px-3 py-2">
+        <span className="font-plex-mono text-[11px] tabular-nums text-fl-ink-2">
+          :{devPort}
+        </span>
         <div className="flex-1" />
-        <Button
-          onClick={reload}
-          size="sm"
-          variant="ghost"
-          className="h-6 px-2 text-xs text-zinc-400"
-        >
-          Reload
-        </Button>
-        <Button
-          onClick={() => window.open(previewUrl, "_blank")}
-          size="sm"
-          variant="ghost"
-          className="h-6 px-2 text-xs text-zinc-400"
-        >
-          Open
-        </Button>
+        <ToolbarButton onClick={reload}>reload</ToolbarButton>
+        <ToolbarButton onClick={() => window.open(previewUrl, "_blank")}>
+          open
+        </ToolbarButton>
       </div>
 
       {/* iframe */}
       <div className="flex-1 relative">
         {(status === "loading" || status === "provisioning") && (
-          <div className="absolute inset-0 flex items-center justify-center bg-zinc-950/50 z-10">
-            <span className="text-zinc-500 text-sm">
+          <div className="absolute inset-0 z-10 flex items-center justify-center bg-fl-ground/80">
+            <StatusNote>
               {status === "provisioning"
                 ? "Provisioning preview certificate..."
                 : "Connecting to dev server..."}
-            </span>
+            </StatusNote>
           </div>
         )}
         {status === "error" && (
-          <div className="absolute inset-0 flex flex-col items-center justify-center gap-2 z-10">
-            <span className="text-zinc-500 text-sm">
-              Could not connect to dev server
-            </span>
-            <Button onClick={reload} size="sm" variant="outline" className="text-xs">
-              Retry
-            </Button>
+          <div className="absolute inset-0 z-10 flex flex-col items-center justify-center gap-2.5 bg-fl-ground">
+            <StatusNote>Could not connect to dev server</StatusNote>
+            <button
+              type="button"
+              onClick={reload}
+              className={`rounded-[4px] border border-fl-line-strong px-2.5 py-1 font-plex-mono text-[11px] lowercase text-fl-ink hover:bg-fl-card ${FOCUS_RING}`}
+            >
+              retry
+            </button>
           </div>
         )}
         <iframe
@@ -180,5 +172,38 @@ export function PreviewPane({
         />
       </div>
     </div>
+  );
+}
+
+/** The pane's chrome controls speak the fleet's quiet lowercase mono voice, the
+ * same one the shell's nav and theme toggle use. They are deliberately not
+ * shadcn `Button`s: those are painted from the greyscale shadcn tokens, which
+ * are a different palette from the fleet's parchment/ink one. */
+function ToolbarButton({
+  onClick,
+  children,
+}: {
+  onClick: () => void;
+  children: React.ReactNode;
+}) {
+  return (
+    <button
+      type="button"
+      onClick={onClick}
+      className={`rounded-[4px] px-1.5 py-px font-plex-mono text-[11px] lowercase text-fl-ink-2 hover:text-fl-ink ${FOCUS_RING}`}
+    >
+      {children}
+    </button>
+  );
+}
+
+/** A message shown over the iframe carries its own opaque ground: what the dev
+ * server has painted underneath is not a background this palette controls, so
+ * ink-on-scrim alone can't be relied on to stay legible. */
+function StatusNote({ children }: { children: React.ReactNode }) {
+  return (
+    <span className="rounded-[4px] border border-fl-line bg-fl-card px-2.5 py-1 font-plex-mono text-[11px] text-fl-ink-2">
+      {children}
+    </span>
   );
 }
--- a/src/components/preview-pane.tsx
+++ b/src/components/preview-pane.tsx
@@ -130,15 +130,23 @@ export function PreviewPane({
         </span>
         <div className="flex-1" />
         <ToolbarButton onClick={reload}>reload</ToolbarButton>
-        <ToolbarButton onClick={() => window.open(previewUrl, "_blank")}>
+        {/* noopener: the preview is agent-authored code, and a bare
+            window.open leaves it holding window.opener on this tab. */}
+        <ToolbarButton
+          onClick={() => window.open(previewUrl, "_blank", "noopener")}
+        >
           open
         </ToolbarButton>
       </div>
 
       {/* iframe */}
       <div className="flex-1 relative">
+        {/* Held at half opacity: this overlay is not rare. Every agent write
+            reloads the iframe 500ms later, so a heavier scrim would blink the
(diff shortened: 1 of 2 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/preview-pane.tsx` |

lennons301-reviewer (approved):
Approving. The ticket's substance is delivered, the objective signals are green, and every load-bearing claim in the PR body held up when I checked it independently.

## Objective signals (no CI on this repo, so run locally)

- `pnpm build` — passes, all 24 routes.
- `pnpm test` — 693 passed / 32 files.
- `pnpm lint` — 16 errors, 1 warning. The one error in a changed file (`preview-pane.tsx:92`, `react-hooks/set-state-in-effect`) is **pre-existing**, as claimed: I diffed lines 1–112 of the file against `origin/main` and the only difference in that whole region is the `Button` → `FOCUS_RING` import swap. The effect is byte-identical. Flagging rather than restructuring the TLS pre-warm in a styling ticket was the right call — the ticket explicitly requires the pre-warm behaviour to be unchanged.

## Acceptance criteria

| Criterion | |
|---|---|
| Preview pane + chrome in fleet palette/type | met |
| Container chrome in fleet palette; no `zinc-*` in either file | met — `grep` for `zinc\|purple\|blue-\|green-4\|red-4\|slate-\|gray-` across both files returns nothing. All four named sites (tab strip, bottom bar, split divider, dot + label) are converted |
| Legible in both themes, active mobile tab in particular | **arithmetically** met; visually unverified (see below) |
| Mobile-tabs / desktop-split behaviour unchanged | met — the strip became a `.map` over `["chat","preview"]`, same `setActiveTab`, same `hidden lg:flex` visibility logic |
| Subdomain proxying / hot-reload unchanged | met — cert pre-warm, retry ladder, debounced activity reload, `previewUrl` derivation and the `sandbox` attribute are untouched. One delta, below |
| `pnpm build` / `pnpm lint` pass for changed files | build yes; lint carries one pre-existing error, verified above |

## What I verified independently

- **Contrast.** Recomputed WCAG ratios from the raw tokens in `globals.css` for every foreground/background pair in the diff. Every figure in the PR body is accurate to 2dp: active tab `fl-ink` on ground **15.55** dark / **12.85** light; idle tab `fl-ink-2` **7.30** / **4.62**; `StatusNote` on `fl-card` 6.62 / 5.56; links `fl-cool` 8.30 / 4.79; and the headline — the old `zinc-100` active tab on the light ground computes to **1.14:1**, so the "effectively invisible" framing is not rhetorical. The declined `FOCUS_RING` finding (2.44 light) and the `fl-green` text finding (3.76 light) also reproduce exactly.
- **Tokens compile.** Grepped the production CSS chunk: all 15 `fl-*` utilities used across both files emit a rule against the correct custom property, including `focus-visible:outline-fl-ink-3`, `decoration-fl-cool/45` and `bg-fl-ground/50`. A typo'd token renders as nothing rather than erroring, so this was worth confirming.
- **Scope resolves.** `--fl-*` is defined under `.fleet`, which `layout.tsx:35` puts on `<body>` — so the tokens resolve on this route, in all three theme scopes.
- **The self-review fix is correct.** `turn-manager.ts:349` does set `containerStatus: "idle"` at end of turn while the task stays `running`, so keying the breath to task status would indeed have left the dot animating through the "agent is waiting on you" state. `live` off `containerStatus === "running"` restores the original signal, and `fleet-dot-live` honouring `prefers-reduced-motion` is a real improvement over `animate-pulse`.
- **The environment excuse is true.** The Playwright chromium in the cache is missing `libnspr4`, `libnss3`, `libnssutil3` and `libasound.so.2`; no system browser is installed. I can't take screenshots either.
- **Sibling zinc is correctly deferred** — `task-stream.tsx` (#121, transcript read path) and `message-input.tsx` (#122, composer).

## For the human merging this

**The "verified in both themes" criterion is met on paper, not on glass.** Neither the implementer nor I can run a browser here, and this repo has no component test harness at all, so nothing but an eyeball closes it. The two things worth a look on a phone: the mobile tab strip in light mode (the whole point of the ticket), and the loading scrim over the iframe, whose 50% opacity was deliberately kept low because it fires on every agent write.

**One behavioural delta in a styling-only ticket:** `window.open(previewUrl, "_blank")` → `window.open(previewUrl, "_blank", "noopener")`. The motivation is sound — the preview is agent-authored code and a bare `window.open` hands it `window.opener` on the operator's tab. The HTML spec removes `noopener`/`noreferrer` from the tokenised feature map before the popup-vs-tab check, so a features string containing only `noopener` should still open a tab, but since it's the single behaviour change in the diff, it's worth one click on **open**.

## Findings

Non-blocking. Nothing here should hold the merge.

1. **Duplicated Code — the link idiom is now in three places.** The anchor class string at `task-chat.tsx:136` and `:146` is byte-identical to itself and, apart from `text-[12px]` vs `text-[11px]`, to `ActionLink` in `src/components/fleet/needs-you.tsx:102`. The PR's stated rule for `ToolbarButton`/`StatusNote` — "promoted when the second caller appears" — is the right rule, and for this atom the second and third callers have now appeared. Worth promoting `ActionLink` into `fleet-bits` with a size variant in a follow-up; the two local copies are also a candidate for the same `.map` treatment the tab strip got.
2. **Nit, same token as the deferred `FOCUS_RING` finding:** the `cancelled` dot is `fl-ink-3` on the ground — 2.44:1 in light, under the 3:1 non-text minimum. It matches `LiveDot`'s `offline` precedent and the state is also stated in words in the terminal footer, so it carries no information alone. Mentioning it only because it's the same token and the same shortfall as `FOCUS_RING`, so the follow-up ticket the PR offers to write could cover both.

Where the PR declined a self-review finding, I agree with each call: `FOCUS_RING` is genuinely app-wide (4 files) and retuning it from a preview-pane ticket would silently restyle every screen; `StatusDot` vs `LiveDot` are different state domains and coupling them to save a 5-entry record would drag the dashboard into this diff; and the unreachable `!devPort` branch is a structural edit beyond a restyle.

The commit messages, the inline rationale comments, and the honest "here is what I could not verify and why" in the PR body are all well above the bar.

PR #141, opened 2026-08-16
Tasks list: fleet reskin + organizeTasks selector (#120)

### LINKED ISSUE #120: Tasks list: fleet reskin + organizeTasks selector
## Parent

#116

## What to build

The `/tasks` list brought to fleet parity and made a legible archive that complements the live dashboard. A pure `organizeTasks(tasks, filter)` selector separates interactive sessions from autonomous runs and applies a single kind filter. Rows render as fleet cards showing status, a kind chip (chat / grill / spec / tickets / implement / review / triage), project, cost, and relative time. Tapping a row opens the task's live view. A clear empty state tells a new user how to begin.

## Acceptance criteria

- [ ] The list renders as fleet cards with status, kind chip, project, cost and relative time
- [ ] Interactive sessions are visually separated from autonomous runs
- [ ] A single kind filter narrows the list
- [ ] `organizeTasks` is a pure selector with unit tests covering the split and the filter
- [ ] Tapping a task opens its live view
- [ ] A clear empty state is shown when there are no tasks

## Blocked by

lennons301:
Closes #120

The `/tasks` archive brought to fleet parity, and its read path fixed — the ticket's field finding folded the two together, because the screen was rendering "No tasks yet. Create one to get started." over a database holding 222 tasks.

## What's here

**`organizeTasks(rows, filter)` — the pure selector** (`src/lib/tasks/organize-tasks.ts`)

The session/run split, the one kind filter, and the chip vocabulary in one pure function, the same seam `buildFleetView` gives the dashboard. A session is named by what it *is* — a grilling reads as `grill`, never as the bare `interactive` — and anything a run owns is loop bookkeeping whatever its kind, so the sessions worth revisiting are never buried under it. The filter offers only kinds that exist (with counts), plus the active one if the data moves beneath it, so it is never a dead end.

The status/kind/skill unions are derived from the schema through a **type-only** import: the column enums stay the one source of truth and the chip maps are genuinely exhaustive (adding a kind to the schema fails the type check here until it gets a chip), while the build confirms no drizzle runtime reaches the client bundle.

**The list** (`task-feed.tsx`, `task-card.tsx`, `/tasks/page.tsx`)

Fleet cards shaped like the dashboard's running cards — project and ticket, kind chip, title, then status, spend and last activity — in two labelled sections, one filter row, and the shadcn status badge retired with the last screen that used it. Tapping a card opens the live view, as before. The empty state names the two ways to begin.

**The read path** (`GET /api/tasks`)

- Projects only the columns a card renders, and joins the project name. On prod the old `select().from(tasks)` shipped 1035 KB of `description` (the full implement prompt, never rendered) inside a 1219 KB response; the same list is now a few hundred bytes a row. Bounded at 200 (raisable to 500), where it was unbounded and growing with every task.
- Polling is sequential — the next request is scheduled once the previous settles, and it aborts on unmount — so a slow connection can't pile polls on top of each other. Cadence is 10s: the dashboard is the live surface, this is the record.
- Failure is no longer indistinguishable from emptiness. A never-loaded list renders loading; a failed one renders the failure with a retry; a failure with rows already on screen is a staleness banner over them; the empty state is reachable only from a confirmed empty result that nothing has failed since. A failing load backs off to a minute rather than hammering a server that is already unwell.
- Two fixes fell out of touching it: a status outside the enum is a 400 rather than a silently empty list, and `status` + `projectId` now `and` together — drizzle's builder was replacing the first predicate with the second, so the status filter was being dropped.

## Tests

`organizeTasks` and `taskTicket` are table-tested (split, ordering, both grilling skills folding into one chip, filter, chip counts, purity); the route's projection, bound, both-filters-together case and empty-parameter handling are tested against a from-migrations DB. `pnpm test` (720 passing), `pnpm build`, and `pnpm lint` on the changed files all pass. `tsc --noEmit` is clean apart from the pre-existing errors in `src/lib/docker/__tests__/container-manager.test.ts`.

Also smoke-tested against a running dev server: the list payload, the empty-parameter contract and the 400.

## Self-review notes

`/code-review` against `origin/main` raised 13 findings; the ones I acted on are in the last commit. Three I disagreed with, stated here rather than silently dropped:

- **"Use SSE like the fleet dashboard rather than polling."** Out of scope — a new stream route with server-side poll-and-diff is a bigger change than the ticket's read-path fix, and the archive doesn't need second-level freshness. The 10s cadence over a payload ~28× smaller is the proportionate version.
- **"The 200-row cap is applied by global recency, before the split, so sessions can fall outside the window."** True, and inherent to a bound without pagination — which #116 puts explicitly out of scope ("no search/grouping/saved views"). The list says when it is showing a capped view. Worth a follow-up ticket if the archive outgrows it.
- **"`taskChip` is computed up to three times per row."** ~1000 map lookups per poll on a list that changes a handful of rows an hour; the selector's clarity is worth more than carrying a denormalised chip through the row type.

Deliberately not touched: the `Caddyfile`'s missing `encode` directive, noted in the same field finding. It is real, but it's an edge concern rather than this ticket's read path, and `Caddyfile` is a `human-signoff` gated path — it belongs in its own ticket rather than gating this one.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/status-badge.tsx` |
| visual-ui | `**/components/**` | `src/components/task-card.tsx` |
| visual-ui | `**/app/**/page.*` | `src/app/tasks/page.tsx` |

lennons301-reviewer (approved):
Approving. Gated for human sign-off (`human-signoff`, auto-merge off), so this only satisfies branch protection — a human still merges after looking.

## What I verified

CI reports no checks on this branch, so I ran the signals locally at b1a6258:

- `pnpm test` — 33 files, **720 passing**, exit 0
- `npx eslint` on the seven changed files — clean (the repo-wide `pnpm lint` failure on `main` is the known pre-existing baseline, untouched here)
- `pnpm build` — exit 0, `/tasks` still prerendered
- `npx tsc --noEmit` — only the two pre-existing `src/lib/docker/__tests__/container-manager.test.ts` errors, which this PR doesn't touch
- The PR's type-only-import claim: `grep -rl drizzle .next/static/chunks/` returns nothing, so the schema import into a client-bundled module really is erased

Against issue #120's acceptance criteria: fleet cards carry status, kind chip, project, cost and relative time; sessions and autonomous runs are separate labelled sections; one kind filter narrows both; `organizeTasks` is pure and table-tested for the split, ordering, purity, filter and chip counts; a card is a `Link` to `/tasks/[id]`; the empty state names both ways to begin. Against the folded-in field finding: the projection drops `description`, the bound is real and clamped (`parseLimit` tested at every edge), polling schedules only after a request settles and aborts on unmount, a failure never renders as the empty state, and the empty state is reachable only from a confirmed empty result with no outstanding error. The read-path state machine reads correctly to me on the four paths (never-loaded, failed-with-nothing, failed-with-rows, confirmed-empty).

Two things I checked for breakage rather than took on trust:

- `src/components/status-badge.tsx` is deleted and nothing else references it or `TaskCard` outside `task-feed.tsx`.
- The list route's new contract (projection, 400 on an out-of-enum status) has exactly one caller — `task-feed.tsx`. Nothing in the app builds a `?status=` query, so the new 400 can't break an existing client.

The `and(...)` fix is a genuine latent bug (two `.where()` calls, the second replacing the first) and the test that covers it constructs the only case that could distinguish the two — same project, differing status. Good test.

Against #116: the card layout matches `running-list.tsx`'s run card element for element, the split follows the spec's "no run / `kind: interactive`" wording, and the tests stay inside the spec's stated culture — pure-function plus API, no `@testing-library`, no Playwright. Fowler baseline: nothing I'd act on. `CHIP_TONE` here and `MODE_TONE` on the dashboard are near-identical shapes over deliberately different vocabularies, which I read as parallel structure, not duplication.

## Suggestions, none blocking

1. **The 200 cap bites today, not eventually.** The bound is applied by global recency before the split and before the client-side filter, and prod already holds 222 tasks — so #116's story 29 ("find a past grilling session without scrolling past every autonomous pass") is already partly out of reach, since the filter can only narrow what fell inside the window, and the loop mints implement/review rows far faster than sessions. The PR argues this is inherent to a bound without pagination, and I agree pagination is out of scope; but a server-side kind filter would fit the existing route with no new UI. Worth a follow-up ticket rather than a change here.
2. **The state machine that was the bug is the untested part.** `organizeTasks` is tested thoroughly; the loading/failed/stale/confirmed-empty logic — precisely what made the screen lie in production — is verified only by reading it, because the spec rules out component tests. A tiny pure `listState(rows, error)` selector beside `organizeTasks` would have put it at the same seam the repo says to test at. Cheap follow-up if this area is touched again.
3. **`TaskListRow` and the route's projection aren't tied at compile time.** They're two hand-written lists that must agree; the key-set assertion catches an added or dropped column at runtime, which mitigates most of it, but a rename would type-check on both sides. Deriving the interface from the query (allowing for `updatedAt` being `Date` server-side and `string` after JSON) would close it.
4. **The Caddyfile `encode` deferral is well argued but untracked.** Agreed it doesn't belong in this PR (gated path, different concern), but there's no open issue for it — worth filing so the other half of the field finding isn't lost.
5. Trivia: `interactive-flow.test.ts` now reads `sessionSkill`/`sessionIssue` through `GET /api/tasks/[id]` even though both are still in the list projection, so that detour isn't needed; and the card no longer surfaces `containerStatus` for a running task, which looks deliberate (the dashboard is the live surface) but is a small loss worth a human's eye.

The self-review notes disagreeing with three `/code-review` findings are all reasonable, and stating them in the PR body rather than dropping them silently is the right call.

PR #145, opened 2026-08-16
Settings: fleet reskin + per-project autonomy control (#119)

### LINKED ISSUE #119: Settings: fleet reskin + per-project autonomy control

lennons301:
Closes #119

`/settings` becomes the fleet's control room. Read top to bottom it answers the three questions the owner arrives with: may the fleet pick up work at all, for which projects, and is the box under it healthy.

## What's here

- **`canArm` (pure, tested)** — `src/lib/projects/autonomy.ts`. One thing blocks arming: a preflight that ran and failed, because that is the one state `decideNext` already fails closed on. A never-checked project *is* armable — arming is what runs preflight. 12 cases cover `canArm` / `armBlocker` / `preflightVerdict`.
- **Project cards** — name, repo, `armed`/`disarmed`, `preflight passing|failing|unchecked`, and the reason when there is one. An armed project sitting on a non-passing preflight says so ("armed, but nothing is claimed until preflight passes"), because armed-and-idle otherwise looks exactly like armed-and-quiet.
- **Arming asks twice.** `arm…` opens a confirmation naming the spend it authorises ($20/attempt, 3 attempts). A failing preflight removes that affordance entirely and leaves `arm anyway…` beside the failing reason — the explicit override — which opens the same confirmation in amber. Disarming is single-press: reversible things shouldn't feel dangerous.
- **Global kill switch** on `PATCH /api/settings/autonomy`, deliberately asymmetric: **stop the fleet** is one press, lifting takes a confirmation (it re-arms every armed project at once). `AUTONOMY_ENABLED` being off is named on the panel, so "I lifted it and nothing happened" has an answer on screen.
- **Docker status** restyled to chips and a hairline panel, with the failure path it never had.
- Runbook updated: the two autonomy controls are presses now, curl kept for when the UI is what's broken. It also called `AUTONOMY_ENABLED` the kill switch, which is the one thing it isn't.

## Self-review (`/code-review` vs `origin/main`, spec #119)

Acted on: the three panels each hand-rolled the same GET (now one `useLoad` hook) and the same card geometry (now `PANEL`/`PANEL_PLAIN`); the two confirmations asked in two shapes (now `ConfirmStrip`); `preflightVerdict` returned a colour, which belongs in the card, as `needs-you.tsx` does with its causes. It also caught two real defects, both fixed:

- a reload that failed *after* the list was on screen set an error nothing rendered — add a project, let the follow-on reload fail, and the form clears as if it worked while the list never shows it. Now the staleness banner the archive already uses;
- the Add button was the one new control without a focus ring, because it hand-rolled the filled skin. Both filled buttons now share `PRIMARY_BUTTON`.

Where I disagreed, and why:

- **Docker status gaining an error state** was flagged as beyond "restyle it". Keeping it: the restyle rewrote that fetch, and a rejected probe used to leave "Checking…" on screen forever — which reads exactly like a hung daemon, the one answer this panel must never give by accident.
- **Lifting the kill switch takes a confirmation** though #116 asks for one only on arming. Lifting arms every armed project at once; engaging stays one press.
- **`new-task-form.tsx` / `task-feed.tsx` touched** — both only adopt the skins extracted here, no behaviour change. #116's "consistency dividend" is the spirit; the alternative was a second copy of each skin.
- **Doppler placeholder `dp.st.stg.` → `dp.st.dev.`** — agent containers use the Doppler `dev` config (CLAUDE.md), so the old placeholder pointed at the wrong environment.

## Scope

`GET /api/projects` still ships `dopplerToken` in cleartext to the browser. Real, known, and explicitly **out of scope** per #116 ("gets its own ticket") — this screen reads presence only (`!== null`), so it is unaffected by whichever way that ticket redacts it.

## Verification

`pnpm build` ✅ · `pnpm test` 732/732 ✅ · `eslint` clean on every changed file (repo-wide lint keeps its 16 pre-existing errors, none in this diff). Arm → preflight re-run → disarm, the kill switch both ways, and a non-boolean rejection (400) exercised against a local server.

Not verified: pixels. No browser is installable here (missing system libs), so the visual pass is the reviewer's — this is a `human-signoff` PR anyway: it touches `components/**` and `app/**/page.tsx`, both gated.

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -28,11 +28,14 @@ or on the `AUTONOMY_ALLOWED_AUTHORS` allow-list. Priority is expressed purely by
 
 Autonomy is off by default at two levels, and **both** must be on for pickup:
 
-1. **Global kill switch** — `AUTONOMY_ENABLED` (env, from Doppler `interlude/prd`).
+1. **Boot master** — `AUTONOMY_ENABLED` (env, from Doppler `interlude/prd`).
+   False and no sweep ever starts; read once, so changing it means a restart.
 2. **Per-project toggle** — `projects.autonomyEnabled` (off by default; enabled
-   deliberately, per project, never in bulk).
+   deliberately, per project, never in bulk). Arm and disarm it on `/settings`,
+   where each project also shows its preflight verdict.
 
-A project is only claimable when both are on **and** its preflight passes.
+A project is only claimable when both are on, the runtime kill switch is lifted
+(*Pause pickup* below), **and** its preflight passes.
 
 ### Preflight: is a repo safe to run unattended?
 
@@ -145,8 +148,9 @@ the dashboard answers "what's happening".
 Pausing only stops autonomous **pickup**. In-flight runs finish, and interactive
 chat/preview is never affected.
 
-- **Globally, right now (kill switch):** flip the durable runtime switch — no
-  restart, effective at the next sweep tick (≤30s).
+- **Globally, right now (kill switch):** press **stop the fleet** on `/settings`,
+  or flip the durable runtime switch by hand — either way no restart, effective
+  at the next sweep tick (≤30s).
 
   ```bash
   # Engage: nothing new is claimed anywhere (implement or triage)
@@ -167,7 +171,8 @@ chat/preview is never affected.
 - **Globally, hard off (boot master):** set `AUTONOMY_ENABLED=false` in Doppler
   `interlude/prd` and restart. Sweeps never start at all. Use this to stand the
   fleet down for a while; use the kill switch to stop it now.
-- **Per project:** disable the toggle —
+- **Per project:** press **disarm** on the project's card in `/settings`, or
+  disable the toggle by hand —
 
   ```bash
   curl -s -X PATCH http://localhost:3000/api/projects/<id> \
--- a/src/app/settings/page.tsx
+++ b/src/app/settings/page.tsx
@@ -1,22 +1,40 @@
 import { AppShell } from "@/components/app-shell";
+import { Eyebrow } from "@/components/fleet/fleet-bits";
 import { ProjectList } from "@/components/project-list";
 import { DockerStatus } from "@/components/docker-status";
+import { KillSwitch } from "@/components/kill-switch";
 
-// Settings is reskinned (and gains the autonomy controls) by its own ticket;
-// here it just moves inside the shared shell (issue #117).
+/**
+ * The fleet's control room (issue #119). Read top to bottom it answers the
+ * three questions the owner actually arrives with: is the fleet allowed to pick
+ * up work at all, which projects is it allowed to pick it up for, and is the
+ * box underneath healthy enough to run it.
+ */
 export default function SettingsPage() {
   return (
     <AppShell section="settings">
-      <div className="space-y-6 py-2">
-        <h1 className="text-2xl font-bold">Settings</h1>
-        <div className="space-y-4">
-          <h2 className="text-lg font-semibold">Docker</h2>
-          <DockerStatus />
+      <div className="space-y-10">
+        <div className="mb-6 mt-2">
+          <h1 className="text-lg text-fl-ink">Settings</h1>
+          <p className="mt-0.5 text-[13px] text-fl-ink-3">
+            What the fleet may do while you&apos;re not watching.
+          </p>
         </div>
-        <div className="space-y-4">
-          <h2 className="text-lg font-semibold">Projects</h2>
+
+        <section aria-label="Autonomy" className="space-y-3">
+          <Eyebrow>Autonomy</Eyebrow>
+          <KillSwitch />
+        </section>
+
+        <section aria-label="Projects" className="space-y-3">
+          <Eyebrow>Projects</Eyebrow>
           <ProjectList />
-        </div>
+        </section>
+
+        <section aria-label="Environment" className="space-y-3">
+          <Eyebrow>Environment</Eyebrow>
+          <DockerStatus />
+        </section>
       </div>
     </AppShell>
   );
--- a/src/components/confirm-strip.tsx
+++ b/src/components/confirm-strip.tsx
@@ -1,0 +1,63 @@
+"use client";
+
+import { ControlButton, PANEL, TONES } from "@/components/fleet/fleet-bits";
+
+/**
+ * The pause between pressing a control and it happening, for the two controls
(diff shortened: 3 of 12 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/confirm-strip.tsx` |
| visual-ui | `**/components/**` | `src/components/docker-status.tsx` |
| visual-ui | `**/components/**` | `src/components/new-task-form.tsx` |
| visual-ui | `**/components/**` | `src/components/project-list.tsx` |

lennons301-reviewer (approved):
Approving. Every acceptance criterion on #119 is met, the objective signals are green, and the domain claims the code makes in its comments hold up when traced. This is a gated PR (`human-signoff`, auto-merge off) — my approval satisfies branch protection only; the visual pass and the merge are yours. Findings below are all non-blocking.

## Verified

- **Objective:** `pnpm build` ✓ (`/settings` still prerenders static, so nothing server-only leaked into the client tree) · `pnpm test` 732/732 ✓ · `eslint` clean on all 12 changed files; repo-wide lint is 16 errors — exactly the documented baseline, no regression.
- **Acceptance criteria:** all eight. Fleet skin on settings/docker-status/project-list; armed + preflight chips with the reason line; arming always routed through `ConfirmStrip` (there is no path to `setAutonomy(true)` that skips it); failing preflight replaces `arm…` with `arm anyway…` beside the reason; disarm is a single press through the existing PATCH; kill switch both directions; `canArm` unit-tested with 12 cases.
- **Contract alignment** against the three endpoints it drives, read rather than assumed: `PATCH /api/projects/[id]` returns the whole re-preflighted row (so `replace`-from-response is sound), and it masks `dopplerToken` to `"••••••••"` — the card reads presence only (`!== null`), so swapping a raw list row for a masked PATCH row keeps `doppler token · set` correct. `PATCH /api/settings/autonomy` returns `{globalAutonomyPaused, updatedAt, envMaster}`, matching `AutonomySettings` field-for-field.
- **The reasoning behind `canArm` checks out.** `decide.ts:1256` fails closed on any non-passing preflight, which is what makes both the "unchecked is still armable" call (arming is what runs preflight) and the "Armed, but nothing is claimed until preflight passes" line on `verdict.state !== "passing"` accurate rather than decorative. `refreshProjectPreflight` returns `null` when GitHub isn't configured, so the "unchecked" copy ("arming runs it, once the GitHub App is configured") is the honest answer on a bare install, not a fudge.
- **No behaviour lost** vs the old `project-list.tsx`: create, name/gitUrl/githubRepo/dopplerToken edit and the linked-Discord readout all survive, and the card header stops being a `div onClick` (keyboard-reachable now). `formatChanged` runs only after the client fetch resolves, so no locale hydration mismatch.
- **Scope of the two extra files** is as the PR body claims: `new-task-form.tsx` loses a byte-identical local `FIELD` and gains `FOCUS_RING` via `PRIMARY_BUTTON`; `task-feed.tsx`'s `RetryButton` becomes a `ControlButton` with equivalent classes. `FOCUS_RING` is still used there, so no dangling import.
- **Smell baseline:** this round is mostly deduplication (`useLoad`, `PANEL`/`PANEL_PLAIN`, `FIELD`, `PRIMARY_BUTTON`, `ConfirmStrip`), and I found no new Duplicated Code, Speculative Generality (every new export has ≥2 call sites) or Repeated Switches — the tone maps are one map each, shared.

## Findings (non-blocking)

1. **The arm confirmation names a budget that may not be the one in force.** It renders `DEFAULT_ATTEMPT_BUDGET_USD` (20), but the effective per-attempt budget is `config.maxBudgetUsd` = `MAX_BUDGET_USD ?? 20` (`src/lib/config.ts:154`), and `sweep.ts:619` puts *that* into the snapshot. On an install where Doppler sets `MAX_BUDGET_USD`, the strip understates or overstates the spend it exists to authorise; a ticket `budget:` directive can also raise one attempt to $75. `/settings/page.tsx` is a server component, so `getConfig().maxBudgetUsd` is one prop away — or soften the copy to name the default and the $75 ceiling. (`MAX_ATTEMPTS` is a true constant, so "3 attempts a ticket" is exact.)
2. **Opening either confirmation drops keyboard focus.** `{!confirmingLift && <ControlButton …>}` and the `intent === null` branch both unmount the trigger, so a keyboard user who presses `arm…` or `lift…` lands on `<body>` and tabs from the top of the page; cancel doesn't restore focus either. Everything is reachable and ringed (#116 story 43 is met in the letter), but autofocusing the confirm button and returning focus on cancel is the finishing touch on the app's two most deliberate controls.
3. **Nit — two error components, near-identical names, one screen.** `Failure` (project-list, an action that didn't take) vs `LoadFailure` (fleet-bits, a resource that wouldn't load). Possible Mysterious Name; `ActionFailure` would separate them at a glance.
4. **Nit — error visibility is asymmetric on cancel.** In `AutonomyControl` the `canArm` branch renders no `<Failure>`, so cancelling out of a failed arm silently drops the message, while the override branch keeps it. Defensible as dismissal, but the two halves of one control tell different stories.
5. **Nit, outside this ticket:** `src/components/ui/input.tsx` and `ui/card.tsx` now have zero consumers outside `src/components/ui/`. Worth sweeping when the last live-view ticket retires `ui/button`.

Not verified by me, deliberately: pixels, dark/light parity and mobile ergonomics — that is the human gate this PR is labelled for.

PR #144, opened 2026-08-16
Live view: markdown transcript read path

lennons301:
Closes #121

The read side of the live task view: interactive sessions (`/grill-me` and the rest) now render as a legible document instead of chat bubbles.

The PoC gate on this ticket was signed off on the issue (2026-08-15); this builds against the five decisions recorded there.

## What's here

- **`toChatView`** (`src/lib/chat/chat-view.ts`) — a pure mapper from stored message rows to a discriminated view-model (`user-chip` / `agent-markdown` / `tool-event` / `system-note`). Every "what is this row" decision lives here; the components switch on `kind` and nothing else. Table-driven tests cover user text, agent markdown, `tool_use` for Read/Bash/Edit/Write/MultiEdit, system rows and legacy plain-text content from Phase 2a.
- **Markdown pipeline** (`src/lib/chat/markdown.ts`) — unified + remark-gfm + rehype-sanitize + rehype-highlight, all npm dependencies, **no CDN at build or runtime**. Syntax colour is fleet CSS variables in `globals.css`; no highlighter theme is imported.
- **Renderers** — owner turns are compact right-aligned chips, agent turns full-width documents, tool calls one quiet row (verb, argument, one right-aligned metric) that expands to input/output and, for an edit, a diff.
- **Header** — branch, issue/PR and spend against this task's actual ceiling (its run's budget, or the per-attempt default for an interactive session) as `Money` + `Gauge`, the dashboard's own metering language. Cost is no longer repeated in the terminal footer.
- **Working pulse** — 6px green dot, ~1.4s breathe, suppressed under `prefers-reduced-motion`, shown in the gap between asking and the first output.

## Safety

Agent output is semi-trusted text, so the pipeline is inert in two independent layers:

1. Raw HTML becomes literal text before it can be a node — `<script>` is shown, never run, and prose about `<Component>` survives instead of silently vanishing.
2. `rehype-sanitize`'s allowlist then runs anyway, because markdown produces dangerous nodes without any raw HTML (`[x](javascript:…)` is a link node).

Plugin **order is the argument**: only trusted plugins (highlighting, the recommendation marker) run after sanitizing, which is why their classes survive and nothing from agent text can spoof them. There's a test for that.

## Verification

`pnpm build`, `pnpm test` (750 pass) and `pnpm lint` on changed files all pass. `pnpm lint` across the repo still reports the 16 pre-existing errors on `main` (`custom-server.js`, `preview-pane.tsx`, `output-parser.test.ts`) — none in files this PR touches.

[...]

PR #146, opened 2026-08-16
Live view: first-class composer

### LINKED ISSUE #122: Live view: first-class composer

lennons301:
Closes #122

The write side of the live task view. The composer was a bare `bg-zinc-900` textarea with two shadcn buttons — the last dark island on the light fleet ground, and the thing you have to use to answer a grilling agent. It is now a first-class input.

## What it does

- **Multiline and autosizing.** The field grows with the draft and scrolls at its cap (`max-h-40`), and re-measures on a width change so rotating a phone or opening the preview pane doesn't leave it the wrong height.
- **Enter sends, Shift+Enter writes a newline** — with a stated deviation on touch, below.
- **Slash commands are discoverable.** Typing `/` in a generation session opens a menu of the six session skills with the same blurbs `/tasks/new` uses (now shared from `src/lib/sessions/skills.ts` rather than restated). Arrow keys move, Enter/Tab accepts, Escape dismisses; the rows are ordinary buttons, so Shift+Tab walks into them and tap/click works. It builds on the existing follow-on slash routing (#63), so it is offered exactly where the orchestrator re-frames a typed slash — a generation session — and not on a plain chat task, where the framing (and the `gh` token a publishing skill needs) isn't there.
- **Queued / working / idle feedback**, driven by real data rather than optimism (below).
- **Continue and complete.** With an empty draft on an idle agent the primary button reads `continue` and sends that word — the same move as replying to Discord's idle notification; as soon as you type it becomes `send`. Completing asks first.
- **Keyboard-operable with visible focus:** the system's `FOCUS_RING` on every control, `focus-within` on the field, `role="status"` on the status line, `aria-label`/`aria-describedby` on the textarea.
- **Fleet palette only** — the palette guard in `chat-render.test.tsx` now covers `message-input.tsx`, and it is verified in both themes (see below).

## Design and structure

Two pure seams carry the decisions, per the repo's culture (and #121's shape):

- `composerState({taskStatus, containerStatus, queued})` → the phase, status label, dot tone, placeholder, whether a message is accepted, whether an empty draft may be "continue", and whether completion is offered. `message-input.tsx` draws it.
- `slash.ts` → when a draft is a command being typed, what matches it, and what the draft becomes when one is accepted.

Both are table-tested (`composer.test.ts`, `slash.test.ts`); the component is rendered through `react-dom/server` in `composer-render.test.tsx` for the states and control enabled-ness. No `@testing-library`, no Playwright — as the parent spec's testing decisions require.

## The queue is real, not optimistic

`queued` is derived from the transcript's own rows (`role === "user" && deliveredAt == null`), so it survives a reload and counts messages sent from Discord too. For it to *drain*, delivery now stamps `updatedAt` alongside `deliveredAt` in the turn manager — that column is what the SSE stream re-emits an already-seen row on. One line, and the only backend change in the PR.

Verified end to end in a browser: with two undelivered rows the composer reads `agent working · 2 queued`; marking them delivered the way the turn manager does drains it to `agent working` live, with no reload.

## Two deliberate deviations, both flagged

1. **Enter on touch.** On a coarse-pointer device the return key writes a newline and the send button sends. A phone keyboard has no Shift+Enter, so Enter-to-send everywhere would make the multiline composer the ticket asks for unusable on the device sessions are read on. The hint under the field always says which contract you have; on any keyboard device it is exactly the AC's "Enter sends; Shift+Enter inserts a newline".
2. **The composer accepts a message on a `blocked` task.** A run parked on a `BLOCKED:` question could previously only be answered from Discord — the composer hid itself. It uses the identical mechanism Discord does (an undelivered user message; `processQueuedMessages` already resumes a blocked task), so this is parity, not new autonomy semantics. `complete` stays disabled there, because the API only completes a `running` task.

Also slightly beyond the letter of the ticket: completion asks "end this session?" before it fires. An accidental thumb on a phone shouldn't end a session and mark its PR ready.

## Verification

`pnpm build`, `pnpm lint` (changed files) and `pnpm test` (832 tests, incl. 56 new) all pass. Checked in a real browser in **dark and light**, at phone (430px) and desktop (1280px) widths: idle, working with a queue, the slash menu open and filtered, a long multiline draft, the completion confirmation, and a focused control. That pass turned up four things now fixed — a `useId` hydration mismatch on the hint id, the hint truncating on a phone exactly where it explains newlines, a slash highlight marked with a shade instead of the app's cool selection hue, and the confirmation fighting the status line for one row.

Not in scope, deliberately: structured/tappable option buttons (Interactivity Level 2 is explicitly deferred by #116), and any change to what `/tasks` or the header show.

## Merged `main` (#145) in

`origin/main` moved under this branch, so `main` is merged in (merge, not rebase). The one conflict was `new-task-form.tsx`: #145 hoisted `FIELD`/`PRIMARY_BUTTON` into `fleet-bits`, this branch lifted `SESSION_BLURBS` out into `lib/sessions/skills.ts` — resolved by taking both.

Two things from #145 I deliberately did **not** adopt in the composer: `ConfirmStrip` (a full-width tinted `PANEL`, scoped by its own doc to the settings controls that authorise unattended spend — dropping one into the composer dock would shove the field down the screen on a phone, where the confirmation is a single 11px row), and `FIELD` (it styles one control and keys off `:focus`; the composer's field is a flex box holding the textarea *and* the send button, so it keys off `focus-within`).

## Self-review

Ran the two-axis review (standards + spec) against `origin/main` with #122 as the spec before pushing. Findings I agreed with are fixed in `61f7ead`: a dangling `aria-describedby` when a send fails, a completion confirmation that could fire after the agent restarted, autosize that only re-measured on a *window* resize (now a width-only `ResizeObserver`, which the preview pane opening also trips), a menu label on a role-less div, and four duplication/naming cleanups.

One finding I did not act on: both passes flagged that the slash menu is offered only in a generation session, while a typed slash still expands natively on a plain chat task (#59) — so it works there but is undiscoverable. I've left it, because the ticket asks for a composer "building on the existing follow-on slash routing", and that routing (`composeSessionTurn`, #63) only applies where `task.sessionSkill` is set; a chat container also has no `gh` token, so a publishing skill offered there would half-work. Advertising a command whose framing isn't applied seemed worse than not advertising it. Happy to widen it if you'd rather.

### DIFF
--- a/src/app/tasks/[id]/page.tsx
+++ b/src/app/tasks/[id]/page.tsx
@@ -36,6 +36,7 @@ export default async function TaskDetailPage({
         githubIssue: task.githubIssue ?? null,
         pullRequestNumber: task.pullRequestNumber ?? null,
         pullRequestUrl: task.pullRequestUrl ?? null,
+        sessionSkill: task.sessionSkill ?? null,
       }}
       domain={process.env.DOMAIN ?? null}
     />
--- a/src/components/__tests__/chat-render.test.tsx
+++ b/src/components/__tests__/chat-render.test.tsx
@@ -116,6 +116,8 @@ describe("fleet palette", () => {
     "src/components/tool-card.tsx",
     "src/components/task-stream.tsx",
     "src/components/task-chat.tsx",
+    // The composer was the last dark island on the light ground (issue #122).
+    "src/components/message-input.tsx",
   ];
 
   const OFF_PALETTE =
--- a/src/components/__tests__/composer-render.test.tsx
+++ b/src/components/__tests__/composer-render.test.tsx
@@ -109,8 +109,9 @@ describe("composer — controls", () => {
     expect(html).toContain("blocked on a question");
     expect(tag(html, /<textarea[^>]*>/)).not.toMatch(DISABLED);
     // Nothing to continue — the agent asked something, so the primary control
-    // waits for an answer to send.
-    expect(tag(html, /<button[^>]*>continue<\/button>/)).toMatch(DISABLED);
+    // waits, as a send, for an answer to send.
+    expect(html).not.toContain(">continue</button>");
+    expect(tag(html, /<button[^>]*>send<\/button>/)).toMatch(DISABLED);
   });
 });
 
--- a/src/components/__tests__/composer-render.test.tsx
+++ b/src/components/__tests__/composer-render.test.tsx
@@ -1,0 +1,125 @@
+import { renderToStaticMarkup } from "react-dom/server";
+import { describe, expect, it } from "vitest";
+import { MessageInput } from "../message-input";
+
+/**
+ * The composer's decisions live in `composerState` / `slash` and are tested
+ * there. What these cover is that the component actually wires them to the
+ * field: the states reach the markup, the controls carry the right enabled-ness
+ * and labels, and the keyboard contract is stated where you can read it.
+ *
+ * Effects do not run in a static render, so this sees the composer's
+ * first-paint state — which is the desktop one (Enter sends). The touch
+ * variant is resolved on mount from a pointer media query.
+ */
+/** The rendered `disabled` attribute, not the `disabled:` utility classes the
+ * same markup is full of. */
+const DISABLED = /disabled=""/;
+
+function tag(html: string, pattern: RegExp): string {
+  return pattern.exec(html)?.[0] ?? "";
+}
+
+function render(props: Partial<React.ComponentProps<typeof MessageInput>> = {}) {
+  return renderToStaticMarkup(
+    <MessageInput
+      taskId="01TASK"
+      taskStatus="running"
+      containerStatus="idle"
+      queued={0}
+      sessionSkill={null}
+      {...props}
+    />
+  );
+}
+
+describe("composer — field", () => {
+  it("is a multiline field that autosizes rather than being dragged", () => {
+    const html = render();
+
+    expect(html).toContain("<textarea");
+    expect(html).toContain('rows="1"');
+    expect(html).toContain("resize-none");
+    // The growth is capped in CSS so a long answer scrolls instead of eating
+    // the transcript.
+    expect(html).toContain("max-h-40");
+  });
+
+  it("labels itself for a screen reader", () => {
+    expect(render()).toContain('aria-label="Message the agent"');
+  });
+
+  it("states the keyboard contract under the field", () => {
+    expect(render()).toContain("enter to send · shift+enter for a newline");
+  });
+});
+
+describe("composer — feedback", () => {
+  it("says the agent is idle and awaiting you", () => {
+    const html = render({ containerStatus: "idle" });
+
+    expect(html).toContain("agent idle");
+    expect(html).toContain("Message the agent…");
+  });
+
+  it("says the agent is working, and that a message would queue", () => {
+    const html = render({ containerStatus: "running" });
+
+    expect(html).toContain("agent working");
+    expect(html).toContain("your message will be queued");
+  });
+
+  it("counts messages the agent has not been handed yet", () => {
+    expect(render({ containerStatus: "running", queued: 2 })).toContain("2 queued");
+    expect(render({ containerStatus: "running", queued: 0 })).not.toContain("queued</span>");
+  });
+
+  it("announces its status rather than leaving it to be noticed", () => {
+    expect(render()).toContain('role="status"');
+  });
+
+  it("takes nothing before the task has a container, and says why", () => {
+    const html = render({ taskStatus: "queued", containerStatus: null });
+
+    expect(html).toContain("queued for a slot");
+    expect(tag(html, /<textarea[^>]*>/)).toMatch(DISABLED);
+  });
+});
+
+describe("composer — controls", () => {
+  it("offers continue on an idle agent with an empty draft", () => {
+    expect(render({ containerStatus: "idle" })).toContain(">continue</button>");
+  });
+
+  it("offers completion between turns", () => {
+    const html = render({ containerStatus: "idle" });
+
+    expect(tag(html, /<button[^>]*>complete<\/button>/)).not.toMatch(DISABLED);
+  });
+
+  it("withholds completion mid-turn, when the API would refuse it anyway", () => {
+    const html = render({ containerStatus: "running" });
+
+    expect(tag(html, /<button[^>]*>complete<\/button>/)).toMatch(DISABLED);
+  });
+
+  it("takes an answer to a blocked run, but offers no bare continue", () => {
+    const html = render({ taskStatus: "blocked", containerStatus: null });
+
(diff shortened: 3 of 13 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/chat-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/__tests__/composer-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/new-task-form.tsx` |
| visual-ui | `**/components/**` | `src/components/task-stream.tsx` |
| visual-ui | `**/app/**/page.*` | `src/app/tasks/[id]/page.tsx` |

lennons301-reviewer (approved):
**Approve.** Gated for human sign-off (`human-signoff`, auto-merge off) — this approval satisfies branch protection only; a human merges.

## Objective signals

No CI is configured on this repo, so I ran everything locally at the PR head (`847e3e4`, matching my checkout):

- `pnpm test` — 844 passed, 40 files
- `pnpm build` — clean
- `npx eslint` on all 13 changed files — clean. `pnpm lint` repo-wide still fails at the known 16-error baseline, all of them in `preview-pane.tsx` and `output-parser.test.ts`, which this PR does not touch.

## Acceptance criteria

All eight met, with one deviation on AC 2 called out below.

- **Multiline + autosize** — `rows=1` + `resize-none`, measured from zero so it shrinks back, capped by `max-h-40` (verified the class is actually emitted in the built CSS). The width-only `ResizeObserver` is the right guard: the chat pane is hidden with CSS rather than unmounted when the preview tab is active, so it re-measures on the way back as well as on rotation.
- **Enter sends / Shift+Enter newline** — on a fine pointer. See deviation 1.
- **Slash commands discoverable** — and I checked the gating claim rather than taking it: `slashMenu` is offered only when `task.sessionSkill` is set, which is exactly the condition `turn-manager.ts:667` puts on `composeSessionTurn`. The menu therefore advertises precisely what the server re-frames. The trigger regex matches `composeSessionTurn`'s own (`/^\s*\/(\S+)/`) on leading whitespace, as the comment claims.
- **Queued / working / idle feedback** — derived from the transcript's own undelivered `user` rows, so it survives a reload and counts Discord follow-ups and delivered review feedback too. I confirmed the initial prompt is stored on `tasks`, not as a message row, so there is no phantom permanently-queued message.
- **Continue and complete** — `canComplete` is `idle` only, which is what `POST /complete` will actually accept (it 400s on anything but `running`, and mid-turn completion would abandon work).
- **Keyboard-operable with visible focus** — `FOCUS_RING` on every control; the field's `focus-within` border shift mirrors the system's own `FIELD` atom.
- **Fleet palette, both themes** — the palette guard now covers `message-input.tsx` and passes. Every token used exists in `globals.css` in both grounds, and `bg-fl-cool/13` does generate CSS (with the `color-mix` variant).
- **build/lint** — above.

## The backend line

The one-line `updatedAt` stamp in `processQueuedMessages` is necessary and correctly scoped: `stream/route.ts` re-emits an already-seen row only on `isNotNull(updatedAt) && gt(updatedAt, lastPollTime)`, so without it the queue could grow but never drain live. `messages.updatedAt` has exactly one other writer (`output-parser` updating a tool row), so nothing else is disturbed.

## The two deviations

1. **Enter on touch** is the one AC not met to the letter, and worth the human merger's eye. I think it is right in spirit — the ticket asks for a composer that is comfortable on a phone, and a touch keyboard has no Shift+Enter, so Enter-to-send everywhere would make the multiline field it also asks for unusable there. It is stated in-product on the hint line either way, and on any keyboard device the contract is exactly the AC's.
2. **Accepting a message on a `blocked` task** — verified as genuine parity, not new autonomy semantics: `processQueuedMessages` already admits `blocked` (`turn-manager.ts:578`), `resumeParkedContainer` restarts the stopped container, and the answer un-parks the run (`:641`) — the identical path a Discord reply takes. `complete` stays correctly disabled there. Nothing under `src/lib/orchestrator/autonomy/**` is touched.

## Minor findings (none blocking)

1. **`CLAUDE.md` not extended.** The sibling ticket #121 (PR #144) added a Key Conventions bullet for `chat-view.ts`; this PR adds two more pure seams in that family (`composer.ts`, `slash.ts`) plus a new shared source of truth in `src/lib/sessions/skills.ts`, and nothing in the context file points at any of them — a reader editing `new-task-form.tsx` has no signal not to restate the blurbs. Nothing existing became *wrong*, which is why this isn't a block, but `~/code/platform/standards/agent-context.md` §2 asks for context to move with the code. Worth a one-line bullet on merge or a follow-up.
2. **The complete-confirmation race dead-ends silently.** `handleComplete` re-checks `state.canComplete` and returns early — right call — but leaves the confirmation row open and says nothing, and while it is open the status line is hidden, so the owner cannot see that the agent started a turn. Narrow (needs a delivery to land in the seconds the confirmation is up) and it fails closed, but a `setError` on that early return would close the loop.
3. **Typo**, `turn-manager.ts:629`: "re-emits a already-sent row" → "an already-sent row".
4. **Suggestion only, no standard behind it:** `onQueuedChange` lifts derived state to the parent through an effect — the shape `react-hooks/set-state-in-effect` exists to catch, which only passes here because the rule can't see through a prop callback. Benign at this size, and the alternative (hoisting the message list out of `TaskStream`) would reopen #121's shape, so I'd leave it.

## What I liked

The two pure seams are the right cut, and the tests are tables of the cases a reader would actually doubt (`blocked/idle` vs `blocked/null`, an idle agent still holding a queued message, prefix-before-substring ranking). `isTerminalTaskStatus` shared with the view — with a test asserting the two cannot drift — is a nice touch, as is the `SESSION_BLURBS` de-duplication. The self-review's one non-action (not offering the menu on a plain chat task) is correct and well argued: advertising a command whose framing isn't applied, in a container with no `gh` token, would be worse than not advertising it.

Issue #118, opened 2026-08-11
Global autonomy kill switch (backend + dashboard)


lennons301:
## Parent

#116

## What to build

A global autonomy kill switch that pauses all autonomous pickup instantly, without restarting the app. A durable single-row `settings` table stores `globalAutonomyPaused`. The autonomy sweep reads it fresh each tick and passes it into the snapshot; the pure `decideNext` reducer emits no pickup actions when it is engaged, exactly mirroring the existing daily-cap pause — in-flight runs, gating and review are unaffected. A small settings endpoint flips the flag. The env `AUTONOMY_ENABLED` remains the hard boot master (when false, sweeps never start); the DB flag is the runtime operational pause layered on top, taking effect at the next sweep tick. The fleet dashboard's existing paused surface (the live dot + banner, today driven by the daily cap) is extended to show when the kill switch is engaged, with its own reason, so a paused fleet is never mistaken for an idle one.

## Acceptance criteria

- [ ] A durable `settings` record holds the global pause flag (schema migration follows the documented drizzle gotchas — bump the journal `when`, clear `local.db*` before tests)
- [ ] Engaging the switch stops all new autonomous pickup at the next sweep tick; disengaging resumes it — no app restart
- [ ] In-flight runs, gating and review are unaffected while paused
- [ ] `decideNext` returns no pickup actions when `globalPaused` is true, covered by unit tests alongside the existing daily-cap pause cases
- [ ] An endpoint toggles the flag
- [ ] The dashboard's live dot + banner clearly reflect the kill-switch state with a distinct reason
- [ ] `env AUTONOMY_ENABLED=false` still prevents sweeps from starting at all
- [ ] `pnpm build` and `pnpm lint` pass for changed files

## Blocked by

None — can start immediately (runs in parallel with #117)

lennons301:
🤖 Attempt 1/3 starting.

PR #147, opened 2026-08-16
Autonomy observability: a kill-switch hold is invisible in the daily digest

### LINKED ISSUE #143: Autonomy observability: a kill-switch hold is invisible in the daily digest
## Parent

#118 — follow-up from its review pass.

## What to build

#118 shipped a global autonomy kill switch: a durable flag that stops all new autonomous pickup at the next sweep tick, surfaced on the dashboard as an amber `held` dot with its own banner. The dashboard tells the truth. The **daily Discord digest does not** — `src/lib/fleet/digest.ts` reports only the daily-cap pause, so a fleet you held all day reads there as merely quiet.

That is the failure mode the kill switch was supposed to remove: not knowing whether nothing happened because there was nothing to do, or because you stopped it and forgot. The dashboard is pull; the digest is the one push surface, and it is the one that goes silent.

Fix the digest, and correct two statements in the runbook that no longer match the code:

- It says the live dot "goes amber-paused" when the switch is engaged. The dot's label reads `held` — `paused` is the daily cap, a deliberately different word for a deliberately different state.
- It documents a `Pickup paused (kill-switch)` log line as the confirmation that the switch took effect. The pause action sits *after* the `eligible.length === 0` early return in the sweep, so engaging the switch against an empty queue logs nothing at all. An operator following the runbook to verify the switch engaged can reasonably conclude it failed. The dashboard banner is unaffected — it reads the row directly rather than the action.

## Acceptance criteria

- [ ] The daily digest states when autonomous pickup is held by the kill switch, distinctly from the daily-cap pause, with its own wording
- [ ] A digest generated while the switch is engaged never reads as an ordinarily quiet day
- [ ] Digest tests cover both pause reasons and the unpaused case
- [ ] `docs/runbook.md` says `held`, not "amber-paused"
- [ ] `docs/runbook.md` states how to confirm the switch engaged in a way that holds when the queue is empty — either by describing the dashboard banner instead of the log line, or by moving the log so it always fires

## Blocked by

None — `digest.ts` and `docs/runbook.md` are untouched by anything in flight.

lennons301:
Closes #143

The daily digest reported only the daily-cap pause, so a fleet held all day by the kill switch arrived in Discord reading as merely quiet — the exact failure mode #118 set out to remove, on the one surface that pushes.

## Digest

New **Autonomous pickup** section, rendered first and every morning, off the read model's own `pickupPaused`:

| state | line |
| --- | --- |
| kill switch | `⏸ Held right now — the kill switch is engaged: … · [Lift it](…/settings)` |
| daily cap | `⏸ Paused — yesterday's spend reached the $500.00 daily cap … the pause lifted at midnight.` |
| neither | `No fleet-wide hold — the kill switch is lifted and the day stayed inside the cap.` |
| no project armed | `No project has autonomy enabled — nothing is claimed unattended.` |

The two hold reasons carry their own copy **and their own tense**, because they are not about the same moment: the cap was breached inside the covered day and lifted itself at midnight (Spend still reports the breach, so when both hold neither fact is lost), while the kill-switch flag is read live off the settings row and therefore describes the fleet as the digest is written. When both hold the switch leads — the view's own precedence, it being the one a human can lift.

Tests cover both pause reasons, both holds at once, and the two unpaused cases; the "never reads as an ordinarily quiet day" criterion is tested by rendering an identical empty day with and without the hold on a genuinely *armed* fleet.

## Runbook

- The live dot reads `held`, not "amber-paused" — `paused` is the daily cap.
- `Pickup paused (kill-switch)` is no longer offered as the confirmation. The confirmation is now the settings row itself: the dashboard dot + banner, or `GET /api/settings/autonomy` headless. The runbook says why the log line isn't reliable — the hold is evaluated after the sweep's empty-queue early return, and a breached cap returns before it and logs its own reason.

I chose the "describe the dashboard banner" option over "move the log so it always fires": moving it above the `eligible.length === 0` return would emit a line every 30s for as long as the fleet is held, and it would touch `src/lib/orchestrator/autonomy/**` (a gated path) for a docs-accuracy fix.

## Self-review

Ran `/code-review` against `origin/main` with #143 as the spec. Acted on:

- the unheld line originally said "Running — no fleet-wide hold", which is false with `AUTONOMY_ENABLED=false` or a failing preflight — it now claims only what the read model knows;
- the kill-switch line now says "held right now", since the flag has no history and shouldn't imply anything about the covered day;
- a future third `PickupPause` reason now fails the build (a `never`) instead of silently reading as "no hold";
- the quiet-day baseline in the test had no project armed, so it would have passed with the kill-switch branch deleted;
- the runbook now names the cap-returns-first case as a second reason the log line can be absent.

Not acted on, with reasons:

- **Use `view.pickupPaused.body` instead of digest copy.** The banner's body is present-tense one-liner copy for a pull surface; the digest needs a different tense (the cap's pause has lapsed by the time the digest is read), a push framing, and a link. The AC asks for the digest's own wording. Both surfaces still branch off the same `pickupPaused` field, so they can't disagree about *whether* the fleet is held.
- **A switch engaged all day yesterday but lifted before the 08:00 send still reads quiet.** True, and inherent: the flag is a boolean with no history. `settings.updatedAt` would make "held since X" derivable — worth a follow-up ticket, not this one.
- **The digest can't see `AUTONOMY_ENABLED` or preflight.** Also true, and shared with the dashboard's live dot (it reads `live` with the boot master off). Fixing it properly means putting the boot master into the shared read model, which changes the dashboard too — a #118 follow-up rather than scope for a digest/docs ticket. This PR's wording is careful not to assert anything it can't back up.
- **`${appBaseUrl}/settings` hardcodes a route the preflight card also emits.** One literal, no behaviour risk; a shared route constant is a bigger cleanup than this diff warrants.

## Checks

`pnpm test` 795 passing, `pnpm build` clean, `pnpm lint` clean on the changed files. (`tsc --noEmit` reports two errors in `src/lib/docker/__tests__/container-manager.test.ts` — pre-existing on `main`, untouched here.)


lennons301-reviewer (approved):
Approving. Reviewed with fresh eyes against #143, `CLAUDE.md`, `docs/runbook.md` and the gate config.

## Gate check

Changed paths — `CLAUDE.md`, `docs/runbook.md`, `src/lib/fleet/digest.ts`, `src/lib/fleet/__tests__/digest.test.ts` — match nothing in `docs/agents/review-gates.yaml` or the estate default (`standards/review-gates.yaml`). No `components/**`, no `app/**/page.*`, no autonomy-control path: the reducer and sweep are untouched, which was the stated reason for choosing the "describe the dashboard" option over "move the log line". Armed state is correct, and I judge the change unremarkable enough for it — one added Discord embed field plus docs.

## Objective signals

No CI is wired to this branch, so I ran everything locally at 9b10abc:
- `pnpm test` — 795 passed, 37 files
- `pnpm build` — clean
- `npx eslint src/lib/fleet/digest.ts src/lib/fleet/__tests__/digest.test.ts` — clean (the repo's lint baseline on `main` is not clean, so changed-files-only is the right scope)

## Acceptance criteria

1. **Kill-switch hold stated distinctly, own wording** — `pickupLines` branches on `view.pickupPaused.reason`, with separate copy per reason. Verified against `fleet-view.ts:385`: the switch outranks the cap in the view itself, so the digest inherits precedence rather than re-deciding it.
2. **Never reads as an ordinarily quiet day** — the strongest test in the set: identical *armed* fleets with and without the hold, asserting the other sections are byte-equal and only "Autonomous pickup" differs. That baseline change (from an unarmed fleet) matters — as noted in the PR body, the earlier version would have passed with the kill-switch branch deleted.
3. **Tests cover both reasons and unpaused** — both holds, both-at-once (with Spend still naming the breach), and both unpaused shapes (armed / nothing armed).
4. **Runbook says `held`** — matches `fleet-bits.tsx:167-174` (`held` -> `bg-fl-amber`) and `fleet-dashboard.tsx:15-23`; the dot literally renders the state word, so "turns amber and reads `held`" is accurate, as is "banner above the panels" (it renders before `PulseStrip`).
5. **Confirmation that holds on an empty queue** — verified in `decide.ts`: `if (eligible.length === 0) return actions;` precedes both gates, and the daily-cap gate `return`s before the kill-switch gate. So both runbook claims (empty queue logs nothing; a breached cap logs `daily-cap` instead) are true of the code as written. `GET /api/settings/autonomy` does return the row plus `envMaster`, so the headless confirmation is real.
6. **Build and lint** — above.

## Correctness spot-checks

- The digest's view is built at `window.end - 1ms` (`digest-schedule.ts:78`), but `loadFleetRows` reads `globalAutonomyPaused` live via `isGlobalAutonomyPaused()` (`rows.ts:91`) — the flag is not time-scoped. The "Held right now" tense is therefore the only honest one, and the code comment says exactly that rather than leaving a reader to discover it.
- Conversely `capPaused` derives from spend since the covered day's local midnight, so "yesterday's spend ... the pause lifted at midnight" is accurate, and `sendIfDue` only ever covers the previous day, so "yesterday" can't drift.
- The unheld copy claims only "no fleet-wide hold" — correct restraint, since `AUTONOMY_ENABLED` and preflight are outside this read model. The three deferrals in the PR body (boot master in the view, `settings.updatedAt` as history, a shared `/settings` route constant) are each argued to a real boundary rather than waved off; I agree none belong in this ticket.
- `String(unhandled)` on a `never` compiles and lints, and the default branch fails the build if `PickupPause` grows a reason — while still reading as *held* at runtime rather than reassuring. Right default for this surface.
- Section ordering: nothing else indexes `sections[]`; `postDailyDigest` maps sections to embed fields generically and already truncates at 1024. Six fields is well inside Discord's limits at realistic line counts.

## Smell baseline

Nothing to raise. The one duplication candidate — digest copy vs `pickupPaused.body` — is deliberate and argued (different tense, push framing, a link), and both surfaces still branch off the same field, so they cannot disagree about *whether* the fleet is held. `pickupLines` sits alongside `needsYouLine`/`backlogLines` with the same shape.

## Nit (non-blocking, no action needed)

`docs/runbook.md` line 181-182 has a stray short wrap ("The next" / "morning's Discord digest ..."). Renders identically; only visible in the source.

PR #153, opened 2026-08-18
Fleet UI + tasks archive: follow-ups deferred by the milestone #3 reviews

### LINKED ISSUE #142: Fleet UI + tasks archive: follow-ups deferred by the milestone #3 reviews
## Parent

#116 — follow-ups the milestone #3 review passes deferred out of their own tickets.

## What to build

Five items the `ticket-reviewer` raised on #120 and #123 and deliberately left out of those PRs, because each was either out of the ticket's scope or app-wide and would have silently restyled screens the ticket didn't own. They are small, unrelated to each other, and cheap together.

**1. Server-side kind filter for the tasks archive.** `GET /api/tasks` bounds to the 200 most recent rows by global recency — *before* the interactive/autonomous split and before the client's kind filter. Production already holds 222 tasks, and the loop mints implement/review rows far faster than a human creates sessions, so #116's story 29 ("find a past grilling session without scrolling past every autonomous pass") is already partly unreachable: the filter can only narrow what fell inside the window. Pagination is out of scope; a `kind` parameter on the existing route is not, and needs no new UI.

**2. `listState(rows, error)` as a pure selector.** The loading / failed / stale / confirmed-empty logic is exactly what made `/tasks` lie in production (an empty state shown over a failed load), and it is currently verified only by reading it, because the spec rules out component tests. Extracting it beside `organizeTasks` puts it at the seam this repo says to test at.

**3. Tie `TaskListRow` to the route's projection.** They are two hand-written column lists that must agree. The key-set assertion catches an added or dropped column at runtime, but a rename type-checks on both sides and ships broken. Derive the interface from the query, allowing for `updatedAt` being `Date` server-side and `string` after JSON.

**4. Contrast floor on two shared tokens.** The `cancelled` status dot and `FOCUS_RING` both resolve to `--fl-ink-3`, which is 2.44:1 against `--fl-ground` in the light theme — under the 3:1 WCAG floor for non-text. #123 declined to retune it from a preview-pane ticket, correctly: `FOCUS_RING` is used in four files and fixing it there would have restyled every screen from a ticket that owned one pane. It needs its own change.

**5. `encode` on the Caddy reverse proxy.** The other half of a field finding whose first half #120 fixed. `GET /api/tasks` was returning 1219 KB in production (1035 KB of it `description`); the bounded projection fixed the payload, but responses are still served uncompressed because the Caddyfile has no `encode` directive. Deploy config is a gated path, so this lands with human sign-off by construction.

Also promote the duplicated link atom: the anchor class string at `task-chat.tsx:136` and `:146` is byte-identical to itself and, apart from `text-[12px]` vs `text-[11px]`, to `ActionLink` in `needs-you.tsx:102`. Three call sites is past the repo's own "promote on the second caller" rule.

**6. The arm confirmation states a budget that may not be in force.** Added after the #119 review. The settings arm confirmation renders `DEFAULT_ATTEMPT_BUDGET_USD` (20), but the effective per-attempt budget is `config.maxBudgetUsd` (`src/lib/config.ts:154`), which is what `sweep.ts:619` puts into the snapshot. Wherever Doppler sets `MAX_BUDGET_USD`, the strip mis-states the very spend it exists to authorise — and a ticket `budget:` directive can raise one attempt to $75 on top. `src/app/settings/page.tsx` is a server component, so the real value is one prop away. (`MAX_ATTEMPTS` is a true constant, so "3 attempts a ticket" is already exact.)

**7. Both confirmations drop keyboard focus when opened.** Also from the #119 review. `kill-switch.tsx` and `project-list.tsx` unmount the trigger when the confirmation opens, so a keyboard user lands on `<body>` and tabs from the top; cancelling does not restore focus. Everything stays reachable and ringed, so the accessibility story is met in the letter — but these are the app's two most deliberate controls, and they are exactly where focus should be handled properly.

## Acceptance criteria

- [ ] `GET /api/tasks` accepts a kind filter, applied **before** the row bound, so narrowing by kind reaches rows older than the newest 200; `/tasks` passes its active filter to the route rather than filtering the window client-side
- [ ] A pure `listState(rows, error)` selector lives beside `organizeTasks` with unit tests covering loading, failed, stale and confirmed-empty; the page renders from it
- [ ] `TaskListRow` is derived from the route's projection rather than hand-maintained, so a column rename fails to type-check
- [ ] The `cancelled` dot and `FOCUS_RING` meet 3:1 against `--fl-ground` in **both** themes, changed once at the token/atom level rather than per screen
- [ ] `ActionLink` is promoted into `src/components/fleet/fleet-bits.tsx` with a size variant; `needs-you.tsx` and both `task-chat.tsx` call sites use it; no copy of the class string remains
- [ ] The Caddyfile enables response compression, and a request to `/api/tasks` on the VPS comes back compressed
- [ ] The arm confirmation shows the **effective** per-attempt budget (`config.maxBudgetUsd`), not the compiled-in default, so it matches what the sweep would actually authorise
- [ ] Opening either confirmation moves focus into it, and cancelling returns focus to the trigger

## Blocked by

- #121 (live view read path) — rewrites `task-chat.tsx`, which holds two of the `ActionLink` call sites

lennons301:
Closes #142

Seven unrelated follow-ups the milestone #3 reviews deferred, plus the two the #119 review added. Each is small; they're together because they're cheap together.

## What changed

**1. Kind filter in SQL.** `GET /api/tasks` takes a `kind` parameter and puts it in the `where`, before the row bound. That's the point: the archive is bounded to the 200 most recent rows by global recency, so a filter applied after the bound could only narrow the window — with 222 tasks on prod and the loop minting implement/review rows far faster than a human opens sessions, the grilling session you were looking for had already fallen out of it. The parameter speaks the list's own vocabulary (`TaskFilter` — a chip, or `all`), so `/tasks` hands its filter over as it stands and an unknown value is a 400 rather than a silently unfiltered list.

The chip vocabulary reads two columns (a grilling is `interactive` *plus* a skill; `triage` is both a pass and a session), so `chipColumns` derives the column shapes from the same `SESSION_CHIP`/`KIND_CHIP` maps `taskChip` classifies with — not a second hand-written list. A new `SessionSkill` must be given a chip to compile, and becomes filterable in the same stroke. Guarded at both layers: `chipColumns` vs `taskChip` over every `(kind, skill)` pair the schema allows, and the generated SQL vs `taskChip` row by row.

**2. `listState(rows, error, loadedUnder)`** is the loading / failed / stale / confirmed-empty decision, out where it's a table. This is the logic that made `/tasks` lie in production — an empty state rendered over a failed load, so 222 tasks read as "No tasks yet" — and the branch order is the argument, so the tests state it: `empty` is reachable only from a load that succeeded, and never from a narrowed one. Nothing of this kind is not nothing at all.

**3. `TaskListRow` derives from the query.** The projection moved into `readTaskList`, and the row type is now `Omit<TaskListQueryRow, "updatedAt">` — verified: renaming a projection key produces a type error at the cards. Drizzle carries the left join's nullability through, so `projectName` stays honestly `string | null`; `updatedAt` is the one patched field, `Date` in SQLite and an ISO string after JSON.

**4. Contrast floor.** New `--fl-mark` token, tuned per theme and checked against ground, surface and card in both. `--fl-ink-3` keeps every text use it had, so no screen is restyled beyond the ring and the dots.

**5. `encode` in the Caddyfile** — the app block only (see below).

**6. `ActionLink` promoted** into `fleet-bits.tsx` with a size variant; all three call sites use it and no copy of the class string remains. The atom owns `Link`-vs-`<a>`, `rel`, and the focus ring — which `needs-you.tsx`'s copy had silently lost.

**7. The arm confirmation states `config.maxBudgetUsd`**, the ceiling the sweep actually enforces, not the compiled-in default.

**8. Focus follows the decision** into either confirmation and back out. `useReturnFocus(open)` fires on the edge back to closed, when refs are already attached, so it works whether React remounts the replacement control or reconciles it in place.

## Things worth a reviewer's attention

- **`/settings` had to become `ƒ`.** Caught by the build output: it was prerendered as static, because reading `process.env` isn't a dynamic signal Next tracks. The image is built by `pnpm build` in a Docker stage with no Doppler env, so the prerender would have frozen `MAX_BUDGET_USD`'s default into the HTML and quietly reintroduced the defect item 7 exists to fix.
- **The Caddyfile change is only the app block.** I checked caddy's encoder before compressing a proxy that streams: an SSE response gets its header written immediately (caddy #6293), compression starts only once a body write exceeds the 512-byte minimum and is never buffered to reach it, and a 101 upgrade passes through — so the live transcript and the preview pane's websockets are unaffected. I first added `encode` to the wildcard preview block too and took it back off: it fronts agent containers' dev servers, nobody asked to compress that, and this is a gated file where the smaller change is the better one.
- **AC "a request to `/api/tasks` on the VPS comes back compressed" can't be verified pre-deploy.** Everything up to the config is done; the check itself needs the deploy.
- **The filter row grew a little.** Moving the filter server-side means the rows on screen are one chip's worth, so the vocabulary can't be read off them — otherwise narrowing to `grill` leaves `grill` as the only way out of `grill`. `filterOptions` draws the row from the last unfiltered load. Not in the ACs, but forced by AC 1.
- **The client-side filter stays**, no longer as the mechanism: between pressing a chip and its answer arriving the rows in hand are the previous query's, and narrowing them is what keeps the list correct rather than briefly wrong. It's a provable no-op on fresh rows, since the SQL and `taskChip` agree by test.

## Self-review findings I disagreed with

- **Chip counts can sum above `all`.** Under an active filter, the active chip states the count the narrowed query found (which can exceed what the bounded window showed — reaching those rows is the whole point), while the other chips still show window counts. The review read this as looking like a bug. I kept it: the number next to the chip you're on should match the list under it, and the footer says when the view is bounded. Happy to drop the active count instead if you'd rather the row stayed internally consistent.
- **The strip doesn't mention the $75 directive ceiling.** The issue notes a ticket `budget:` directive can raise one attempt on top, but the AC asked for the effective per-attempt budget, and inventing extra confirmation copy is a design change I didn't want to make on your behalf.
- **`confirm-strip-render.test.tsx` is a component test**, which item 2's rationale says the spec rules out. It renders with `renderToStaticMarkup` (no DOM), matching `chat-render.test.tsx`, and asserts only the precondition that silently defeats the whole mechanism: `focus()` on a non-focusable `div` does nothing.
- **`import "server-only"` in the query module** was suggested to enforce the type-only-import invariant. It isn't a dependency here and there's no precedent, and the real failure is already caught — a value import would pull `better-sqlite3` into a client bundle and fail `pnpm build` in CI.

Two adjacent defects I noticed and left alone: pressing **disarm** also drops focus (a single-press control swapping for its opposite, not a confirmation), and the app has a second hover-underline link idiom (`+ new task`, the ledger rows) that `ActionLink` doesn't cover. Both want their own tickets.

## Verification

`pnpm test` — 894 passing (43 files, +31 from this branch). `pnpm build` clean, `/settings` confirmed `ƒ`. `pnpm lint` clean on every changed file (the repo's 16-error baseline on `main` is untouched; the 2 pre-existing `tsc` errors in `container-manager.test.ts` are likewise untouched). Both new tripwires mutation-checked — reintroducing `bg-fl-ink-3` and breaking `chipColumns` each fail.

Gated paths: `Caddyfile` (infrastructure) and the visual-UI estate default (`src/components/**`, `src/app/settings/page.tsx`, `globals.css`), so this needs `human-signoff` and can't auto-merge.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/*.css` | `src/app/globals.css` |
| visual-ui | `**/components/**` | `src/components/__tests__/confirm-strip-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/__tests__/fleet-tokens.test.ts` |
| visual-ui | `**/components/**` | `src/components/confirm-strip.tsx` |
| visual-ui | `**/components/**` | `src/components/project-list.tsx` |
| infrastructure | `Caddyfile` | `Caddyfile` |

lennons301-reviewer (approved):
Approving. Every acceptance criterion is met, and I re-derived the load-bearing claims rather than taking the PR body's word for them. Gated (`human-signoff`, auto-merge off) — correctly, for `Caddyfile` (repo `infrastructure`) and the estate `visual-ui` default. I have not re-armed anything.

## Objective signals

This repo has **no PR CI** — `.github/workflows/` holds only `deploy.yml` on push to `main` — so I ran everything locally:

- `pnpm test` — 894 passing, 43 files.
- `npx tsc --noEmit` — only the 2 pre-existing errors in `container-manager.test.ts`, which this PR doesn't touch.
- `npx eslint` over all 20 changed files — clean.
- `pnpm build` — clean, and `/settings` does render `ƒ` as claimed.
- Branch is 0 behind `main` (`#146` merged), `MERGEABLE`.

## Acceptance criteria

All met bar AC 6's second half ("comes back compressed on the VPS"), which is unverifiable pre-deploy and was disclosed as such. Notes below on what a human should check at merge.

## What I verified independently

**AC 3, the rename tripwire actually bites.** Renaming `costUsd` → `costUsdRenamed` in `TASK_LIST_COLUMNS` produces `TS2339: Property 'costUsd' does not exist on type 'TaskListRow'` at `src/components/task-card.tsx:68` and `:69`, plus the selector tests. Type-probed the derived row too: `projectName` is honestly `string | null` (drizzle carries the left join through), `status`/`kind` stay narrow enums, `updatedAt` is `string`.

**AC 4, the contrast numbers, recomputed from scratch.** `--fl-mark` against ground/surface/card: dark 3.85 / 3.67 / 3.50, light 3.22 / 3.56 / 3.88. The old ink-3 on light: 2.44 / 2.69 / 2.93 — the ticket's 2.44:1 reproduces exactly. No non-text ink-3 use survives anywhere in `src/` (`bg-`, `outline-`, `border-`), and `text-fl-ink-3` is untouched as intended.

**Both new tripwires are real, not decorative.** Mutating light `--fl-mark` back to ink-3 and reverting `FOCUS_RING`/`offline` fails 6 tests; mutating `chipColumns` (`skills.length > 0` → `> 1`) fails 5 across *both* the selector test and the SQL-agreement test. The two-layer guard on the chip vocabulary is the strongest thing in this PR.

**AC 5** — no copy of the link class string remains (`decoration-fl-cool/45` appears once, in the atom).

**The `or(...alternatives) as SQL` cast in `chipFilter`** is safe: all nine `TASK_CHIPS` are in the image of `SESSION_CHIP`/`KIND_CHIP`, with `chat` covered by the `NO_SKILL_CHIP` branch. A future orphan chip would degrade to *no* `where` (silently unfiltered), but the per-chip exhaustive route test catches exactly that, and it derives its space from `schema.tasks.kind.enumValues` / `SESSION_SKILLS`.

**The type-only import's safety net.** Converting it to a value import fails `pnpm build` hard (`the chunking context does not support external modules (request: node:fs)`, traced through `./src/db/index.ts`), and `.next/static/` carries no `better-sqlite3` reference. So the argument against `import "server-only"` holds — with one correction below.

## Caddy `encode` — checked against the source, not the comment

This is the gated line, so I read `modules/caddyhttp/encode/encode.go` rather than trusting the rationale. The change is safe, but by **two** mechanisms, neither of which is the one the comment states:

1. `responseWriter.WriteHeader` has an `isSSE(...)` branch that writes headers immediately (the #6293 fix) — so the transcript is never held back. This half of the comment is right.
2. `init()` gates on `isEncodeAllowed(rw.Header())`, which refuses to encode a **response** carrying `Cache-Control: … no-transform`. `src/lib/sse.ts:40` sets exactly that, so Interlude's SSE is never compressed at all. Belt and braces.

The stated mechanism — "only compresses once a body write exceeds `minimum_length` (512 bytes)" — does **not** apply to SSE: the `isSSE` branch calls `rw.init()` at header time, bypassing the `MinLength` check entirely. The conclusion is correct; the reasoning as written isn't. Since this is a gated file where the comment is the safety argument a future reader will lean on, I'd suggest amending it to cite `no-transform` + #6293. Not blocking, and not worth another round-trip on its own.

Version note for the merger: `docker-compose.yml` pins the floating `caddy:2-alpine`. The #6293 fix landed in Caddy 2.9, so any current image is fine — just make sure the deploy pulls rather than reusing a stale cached layer.

## Non-blocking, for the human at sign-off

1. **The chip-count disagreement.** I'd keep it as the implementer argues — the number beside the chip you're on should match the list under it. But one premise of their rebuttal doesn't hold: `bounded` is computed from the *narrowed* row count (`loaded.rows.length >= TASK_LIST_LIMIT`), so in the exact case that looks inconsistent — `all 200 | grill 12` over a 12-row list — the "showing the 200 most recently active…" footer doesn't render at all. The mitigation they cite isn't there. Still their call; no AC covers it.
2. **AC 8 has no behavioural test, and can't have one here.** The repo has no DOM environment whatsoever — no jsdom, happy-dom or testing-library installed — so `confirm-strip-render.test.tsx` asserting only the static `tabindex="-1"` precondition is the right call, not a shortcut. Worth one manual keyboard pass at merge: Tab to `arm…`, Enter, check focus lands on the strip and that cancel returns it to the replacement control. (`ref`-as-a-plain-prop needs React 19 — confirmed 19.2.3, so the wiring is valid.)
3. **A sub-second transient on filter change**: the footer text and `organizeTasks` read the *current* `filter` while `loaded.rows` are still the previous query's, so the footer can briefly name the new chip over old rows. `listState` correctly reads `loaded.filter`, which is the part that matters. Cosmetic.
4. **Estate-level, not this ticket**: `standards/testing.md` requires "CI runs tests on every PR" and this repo has none. That makes the PR body's "would fail `pnpm build` in CI" optimistic — the failure surfaces at *deploy*, after merge. Worth its own ticket.

## Smell baseline

Clean. The one duplication — the client-side filter alongside the SQL one — is a deliberate transient-correctness measure, disclosed, and proven a no-op on fresh rows by the row-by-row SQL/`taskChip` agreement test. `chipColumns` derives from the classifier's own maps rather than restating them (the Repeated Switches fix, done properly); `Loaded { rows, filter }` bundles a genuine data clump; `ActionLink`'s `size` variant has both values in use; nothing speculative. `organize-tasks.ts` has grown to hold the row type, chips, `chipColumns`, `organizeTasks`, `listState` and `filterOptions` — brushing Divergent Change, but it is coherently "the tasks-list read model" and I wouldn't split it yet.

PR #154, opened 2026-08-18
Fleet surfaces are blind to the boot master: autonomy off still reads as running

### LINKED ISSUE #148: Fleet surfaces are blind to the boot master: autonomy off still reads as running
## Parent

#118 — raised by its own review, and again by #143's, which fixed the sibling half.

## What to build

The fleet's two "is pickup happening?" surfaces — the dashboard live dot and the daily Discord digest — know about the two *runtime* holds (the kill switch, the daily cap) and nothing about the **boot master**. With `AUTONOMY_ENABLED=false`, or a project whose preflight is failing, both still read as running.

So the fleet can look completely healthy while no ticket can possibly be picked up. That is the exact failure the kill-switch surfacing existed to remove — #118 closed it for the runtime flag and #143 closed it for the digest, but both were built on a snapshot that has never carried the boot state. A green dot currently means "nothing is holding pickup *that this view knows about*", which is not what anyone reads it as.

`decideNext` already returns no pickup actions when autonomy is off at boot; the gap is purely that the read model can't see why. Both surfaces branch off the same field today (`pickupPause`), so the fix belongs there rather than in two places — a third reason, not a second mechanism, so the two surfaces cannot disagree.

Preflight is per-project rather than fleet-wide, so it needs a decision the other two didn't: a fleet of six armed projects where one fails preflight is not "held". Sinking that into the same fleet-wide field would over-claim. Treat the boot master as fleet-wide and preflight as per-project, and say the per-project part where projects are already listed.

## Acceptance criteria

- [ ] The live dot and the digest both reflect `AUTONOMY_ENABLED=false` as a distinct hold, with its own wording — not as "running", and not conflated with the kill switch (one is a deploy-time master, the other a runtime pause a human can lift from the UI)
- [ ] The reason is carried in the shared read model, so the dot and digest cannot disagree
- [ ] A project whose preflight is failing is visibly not pickable, stated per project rather than as a fleet-wide hold
- [ ] Adding a further pause reason fails the build rather than silently reading as unheld (the guard #143 added stays intact)
- [ ] `decideNext`'s existing behaviour is unchanged — this is a read-model and presentation change only
- [ ] Unit tests cover boot-master-off, preflight-failing, and both-with-a-runtime-hold

## Blocked by

None.

lennons301:
Closes #148

The dashboard's live dot and the daily digest knew about the two *runtime* holds on pickup — the kill switch and the daily cap — and nothing about the **boot master**. With `AUTONOMY_ENABLED=false` the fleet rendered green and the digest said "No fleet-wide hold", while no ticket could possibly be claimed.

Both surfaces branch off the same field, so the fix is there: a third reason, not a second mechanism.

## What changed

**Read model** (`fleet-view.ts`, `rows.ts`)
- `FleetRows` gains `autonomyEnabledAtBoot`, read from the same config the sweep gates itself on.
- `PickupPause.reason` gains `autonomy-off-at-boot`, ranked **above** the kill switch and the cap. Precedence is by what a reader must act on: with no sweep running at all, naming the kill switch would send the owner to press a control that changes nothing — the "I lifted it and nothing happened" failure the settings panel already guards against in prose.
- New `ProjectPickupHold` (`autonomy-off` / `preflight-failing` / `preflight-unchecked`) on the per-project backlog rows, deliberately **not** in the fleet-wide field: six armed projects with one failing preflight is not a held fleet, and saying so would over-claim. It mirrors the reducer's fail-closed rule (`decide.ts` treats anything but `passing` as ineligible), so a preflight that never ran is a stated hold rather than a silent one.

**Surfaces**
- Dot reads `off` — amber like `held`, but its own word, because the dot prints its state and `held` would point at the wrong control. Banner carries the matching copy.
- Digest's *Autonomous pickup* section leads with the boot master, in the present tense, and never offers `/settings` as the remedy. Its unheld line no longer disclaims the boot master (it can see it now) and still doesn't promise every project was picking up — preflight is per-project, and *Backlog* is where that is said: `moontide: 3 — not picked up: preflight is failing`.
- The per-project preflight needs-you card now states the consequence ("none of its tickets are picked up"), not just the state.

`decideNext` is untouched — the diff contains no `src/lib/orchestrator/` changes.

## The build guard (AC 4), verified rather than asserted

I temporarily added a fourth `PickupPause` reason and a new `preflightStatus` value and ran `tsc`. Four errors fire: both `Record<PickupPause["reason"], …>` maps in `fleet-bits.tsx`, the `never` in `digest.ts`'s `pickupLines`, and now `projectHold` in `fleet-view.ts`. Nothing reads as unheld by omission.

## Checks

`pnpm test` — 862 pass (41 files). `pnpm build` compiles. `pnpm lint` clean on every changed file. Two pre-existing `tsc` errors in `src/lib/docker/__tests__/container-manager.test.ts` are on `main` too and untouched here.

## Self-review

Ran `/code-review` against `origin/main` with #148 as the spec. Standards found the runbook quoting copy the last commit had changed, `projectHold` using a ternary where the rest of the diff argues for exhaustive maps, and a non-null assertion in the backlog map — all fixed. Spec found the same doc drift, and that `quietSubline` still printed "autonomy on" under a hold; that one is the ticket's headline failure one panel below the dot, so it now names the hold in the dot's own word, off a map both components share.

Two scope notes the spec reviewer raised, where I made a deliberate call rather than silently ignoring it:

- **The never-run preflight card is new, not just a re-word.** The ticket asks that a project failing preflight be *visibly not pickable*; the reducer refuses a never-checked preflight identically, so covering only `failing` would leave the same blindness one state over. It can fire transiently between arming a project and the first preflight run — amber, on a card that names the fix.
- **`ProjectPickupHold` includes `autonomy-off`, which the ticket only implied.** The field's own claim has to be complete: a dormant project's backlog depth printed with nothing beside it reads as work about to start. Omitting it would have made `hold: null` mean "pickable" in one case and "not, but we didn't say" in another.

Both are small and both are removable if a reviewer disagrees.


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (dismissed):
Reviewed with fresh context against #148, CLAUDE.md, `docs/runbook.md`, `docs/agents/review-gates.yaml` and the touched modules. Approving: gated for human sign-off (`human-signoff` label, auto-merge off), so this satisfies branch protection and a human merges.

## Objective signals

This repo has no PR CI (`.github/workflows/` is deploy-on-main only), so I ran everything locally on `a870113` — which contains `origin/main` (`3b32d5a`), so the base is current:

- `pnpm test` — 862 passed / 41 files. The new `src/components/__tests__/needs-you-render.test.tsx` runs (3 tests) and follows existing precedent (`chat-render`, `composer-render`).
- `pnpm build` — compiles.
- `npx eslint` on all 9 changed source/test files — clean.
- `npx tsc --noEmit` — only the two pre-existing `src/lib/docker/__tests__/container-manager.test.ts` errors, which I confirmed are identical on `origin/main` and untouched here.

## Acceptance criteria

1. **Boot master as a distinct hold, own wording, not the switch's** — met. `pickupPaused.reason: "autonomy-off-at-boot"`, dot `off` (amber, its own word), banner "Autonomy is off on this install (AUTONOMY_ENABLED) … the kill switch cannot start one", and the digest's lead line explicitly says the switch can't lift it. `fleet-view.test.ts` asserts the body does *not* match `/kill switch engaged/i` and `digest.test.ts` asserts the line contains no `/settings` link — the right shape of negative assertion for a "don't send the reader to a dead control" requirement.
2. **Carried in the shared read model** — met, and slightly better than asked: `FleetView.pickupPaused` is the single field for dot, banner *and* digest, and the last commit routed `needs-you.tsx`'s quiet sub-line through the same `PAUSE_DOT` map, closing the one place the panel could still have printed "autonomy on" under a hold. `grep autonomyOn` shows the only remaining readers are the digest's unheld line and that sub-line's fallback, both correct.
3. **Failing preflight visibly not pickable, per project not fleet-wide** — met. `ProjectPickupHold` rides `queue.byProject[].hold` (digest Backlog: `broken: 4 — not picked up: preflight is failing`) and the needs-you card now names the consequence. `buildFleetView` is tested to keep `pickupPaused` null with one project failing preflight, which is the over-claiming this AC was guarding against.
4. **A further pause reason fails the build** — met. Both `Record<PickupPause["reason"], …>` maps moved into `fleet-bits.tsx` (so the dashboard and needs-you share one keyed map rather than drifting), and `pickupLines`'s `const unhandled: never = reason` still stands. I verified this by reading rather than by mutating the tree — TS rejects a missing key in a `Record` literal and rejects the `never` assignment once the union widens, which is exactly the #143 mechanism. `projectHold` adds the same guard for `preflightStatus`.
5. **`decideNext` unchanged** — met. No `src/lib/orchestrator/**` file in the diff. I checked the claim the new field mirrors the reducer: `sweep.ts:223` gates on `config.autonomyEnabled` and `decide.ts:1256` refuses any `preflightStatus !== "passing"`, so `autonomyEnabledAtBoot` and `projectHold`'s fail-closed default both match the behaviour they describe.
6. **Tests for boot-master-off, preflight-failing, and both-with-a-runtime-hold** — met, including the case I'd have asked for: with the master off *and* the switch engaged *and* the cap breached, the reason is the master while `spend.capPaused` and the `cap` needs-you card survive being outranked, and the digest's Spend section still reports `$512.34 of $500.00`. Nothing is lost to precedence.
7. **Build + lint on changed files** — met (above).

## Judgement on the two scope calls flagged in the PR body

Both are sound and I would not strip either.

- Raising a card for a **never-run** preflight is the same fail-closed rule the reducer applies (`!== "passing"`); covering only `failing` would have left the identical blindness one state over. It self-clears — `startPreflightRefresh()` runs whenever GitHub is configured, independent of autonomy.
- `autonomy-off` in `ProjectPickupHold` is used and tested (a project disarmed after the sweep recorded its depth keeps that depth in the in-memory store, and would otherwise print bare). Not speculative generality.

## Non-blocking notes

None of these block; the first two are worth the human merger's eye since the change is gated on `components/**` anyway.

- **`off` sits next to `offline` in the dot's vocabulary** (`fleet-bits.tsx`). Amber-with-banner vs grey-with-none distinguishes them, and the reasoning for not reusing `held` is right, but they are one glance apart and mean unrelated things (boot master vs SSE stream down). A design-language call for the owner, not a defect.
- **One residual route of the ticket's own shape.** `runAutonomySweep` has *two* boot gates — `!config.autonomyEnabled` and `!isGitHubConfigured()` — and only the first is modelled. So `AUTONOMY_ENABLED=true` with the App unconfigured still renders a green dot and a digest reading "No fleet-wide hold — autonomy is on, …", while no sweep runs. The old digest comment disclaimed unmodelled routes; the new one narrows that disclaimer to preflight. Out of this ticket's ACs (which name `AUTONOMY_ENABLED`) and unreachable on an install where the loop functions at all — but it is the same failure mode, so worth a follow-up issue rather than a silent gap.
- **The one non-exhaustive branch in the diff** is the needs-you card body in `fleet-view.ts`: `hold === "preflight-failing" ? … : "Preflight has never run, …"`. `hold` is typed as the full `ProjectPickupHold` there, so a fourth member would silently render as "never run" — the one spot that doesn't get the build guard the rest of the diff argues for. A keyed map or a `switch` would match `projectHold` above it.
- **Wording drift, minor:** the digest says `preflight has never passed` where the card and runbook say "never run". "Never passed" is also true of a failing preflight, which is the neighbouring hold.
- **Suggestion only:** `pickupPaused` is now a three-deep nested ternary. A first-match list or a small function would read closer to the precedence comment above it.
- Docs match the surfaces: the CLAUDE.md `settings` bullet, the runbook's preflight/boot-master/"nothing is being claimed" edits, and the `off`/`held`/`paused` triple all describe what the code actually prints, and `/settings`' own `!envMaster` prose stays consistent with the new banner.

Issue #151, opened 2026-08-18
A stalled outbound call permanently leaks a queue slot — one phantom wedges the whole box


lennons301:
## What happened in production (2026-08-18)

An interactive grilling session on Moontide (`01M09TACWZ31M9KQ5KZQER8V6V`) was completed at 07:04:38. The next task (`01M09V20FC15BBT7JNSDPM9QXS`, 07:06:02) never dispatched: it sat `queued` with `branch: null` and `containerName: null` — proof `startTask` was never called, since its first DB write is `{status: "running", branch, containerStatus: "setup"}` (`turn-manager.ts:201`). Meanwhile the box had **no agent containers**, no task carried a `containerStatus`, and the fleet dashboard read `slots: {total: 1, used: 0, saturated: false}` with `needsYou: []`.

The DB said idle; the queue said full. Only a restart cleared it.

## Root cause

`occupiedSlots()` (`queue.ts:36-45`) counts live `activeTasks` entries **plus** `processingTasks` reservations for tasks not in `activeTasks`. A step-2 reservation is released only by `.finally()` on the `processQueuedMessages` promise (`queue.ts:158`) — so a promise that never *settles* holds a slot forever.

Reconstructed from the message ledger:

| time | event |
|---|---|
| 06:58:56, 07:01:39 | follow-ups **delivered** — proves the reservation set was clear at both moments (step 2 skips a task already in it, `queue.ts:135`) |
| 07:01:37 | last reservation added for delivery |
| 07:02:06 | turn ends; 07:02:07 `Branch … pushed.` |
| — | `runPostTurnCommitAndPush` then awaits `createDraftPr` (`turn-manager.ts:1407`) — **never returns**: no PR on the task, no `Push warning` message, no further output |
| 07:04:38 | `completeTask`'s `finally` deletes the `activeTasks` entry → the still-held reservation becomes a **phantom occupant** of the box's only slot |

The unbounded await is the outbound GitHub call. `src/lib/github/client.ts` builds every client as `new Octokit({ auth })` with **no `request.timeout`**. Every call site catches errors and returns `null`, but a *stall* is not an error — nothing bounds it. The same applies to Discord REST calls awaited inside reserved promises.

`CAPACITY_SLOTS=1` in `interlude/prd` (the stopgap from the #93 OOM) makes one phantom fatal to the whole box rather than costing one slot.

### Why grilling sessions are the worst-exposed shape

[...]

PR #155, opened 2026-08-18
Completing a task while the agent starts a turn dead-ends silently

### LINKED ISSUE #149: Completing a task while the agent starts a turn dead-ends silently
## Parent

#122 — raised by its review pass.

## What to build

`handleComplete` re-checks `state.canComplete` and returns early when the task is no longer completable. That check is right and the failure is safe — nothing wrong happens. But the early return says nothing and leaves the confirmation row open, and while that row is open the status line is hidden. So the owner is left looking at a confirmation that does nothing when clicked, with the one surface that would explain why deliberately covered.

The window is narrow: a delivery has to land in the seconds the confirmation is up. It is exactly the case a person hits when they decide to complete a task at the same moment the agent starts a turn — which is not a rare coincidence, it is the normal end of a session.

The fix is small: report the reason on that early return rather than returning silently, and let the owner see the state that changed under them.

## Acceptance criteria

- [ ] The early return in `handleComplete` surfaces why completion was refused, rather than failing silently
- [ ] The confirmation row does not stay open in a state where confirming does nothing
- [ ] The status line is visible, or the reason is otherwise legible, when completion is refused this way
- [ ] A unit test covers the refusal path — the agent becoming non-completable while the confirmation is open

## Also in scope

Typo at `src/lib/orchestrator/turn-manager.ts:629`: "re-emits a already-sent row" → "an already-sent row".

## Blocked by

None.

lennons301:
Closes #149

## What changed

`handleComplete` re-checked `canComplete` and returned silently when the agent had started a turn under the open confirmation — so the owner was left looking at a question that did nothing when answered, with the status line that would explain it covered by that very row.

The reason is now a value. `completionRefusal` (`src/lib/chat/composer.ts`) names, per phase, why ending the session is not on offer; the composer closes the confirmation the moment completion stops being possible — giving the status line back — and says what changed on the line under the field. Same sentence on the disabled complete button's tooltip, so the two surfaces can't drift.

A refusal is toned as a notice, not an error: nothing went wrong, the session moved on. It steps aside after 8s and hands the line back to the keyboard hint, which the status line beside it keeps saying anyway.

Also in scope, as the ticket asks: the `a already-sent row` typo at `turn-manager.ts:629`.

## Acceptance criteria

- [x] The early return surfaces why completion was refused — `completionRefusal(state)`, shown under the field
- [x] The confirmation row does not stay open where confirming does nothing — a layout effect closes it before paint, so not even a frame of it survives
- [x] The reason is legible — the status line is restored *and* the reason is written out; the disabled button repeats it
- [x] A unit test covers the refusal path — `src/components/__tests__/composer-refusal.test.tsx`, plus the per-phase table in `composer.test.ts`
- [x] `pnpm build` and `pnpm lint` pass (lint clean on all changed files; `tsc --noEmit` reports only the two pre-existing errors in `src/lib/docker/__tests__/container-manager.test.ts`, untouched here)

## Self-review

Ran `/code-review` against `origin/main` with #149 as the spec before pushing; the second commit acts on six findings — the untested effect, the after-paint close, dropped focus, the sub-AA amber, the contradictory tooltip, and the `null`-permitting `REFUSALS` type. Details in that commit message.

**One judgement call worth flagging:** the AC asks for a unit test of "the agent becoming non-completable while the confirmation is open", which lives in a component effect, and this repo had no DOM test capability — `composer-render.test.tsx` is `renderToStaticMarkup`, where effects never run. I added `jsdom` and `@testing-library/react` as devDependencies rather than settle for testing the pure lookup alone, because without them deleting the effect body left the whole suite green. They are dev-only: CI here is deploy-only, and the runtime image copies the standalone build, so neither ships. Four of the new file's seven tests fail if the effect is removed (verified). If you'd rather not carry the test infra, the alternative is dropping that file and accepting the pure-function coverage.

**Not fixed, deliberately:** the review also flagged that a successful complete POST latches `completing` forever — `/api/tasks/[id]/complete` is fire-and-forget, so a `completeTask` that throws or an SSE drop leaves "completing…" with no note. That is pre-existing and is the same defect class #151 owns (a completed session holding a slot until restart), so it stays out of this ticket's UI half.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,8 @@
   },
   "devDependencies": {
     "@tailwindcss/postcss": "^4",
+    "@testing-library/dom": "^10.4.1",
+    "@testing-library/react": "^16.3.2",
     "@types/better-sqlite3": "^7.6.13",
     "@types/dockerode": "^4.0.1",
     "@types/hast": "^3.0.5",
@@ -53,6 +55,7 @@
     "drizzle-kit": "^0.31.9",
     "eslint": "^9",
     "eslint-config-next": "16.1.6",
+    "jsdom": "^30.0.1",
     "tailwindcss": "^4",
     "typescript": "^5",
     "vitest": "^4.1.0"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -96,6 +96,12 @@ importers:
       '@tailwindcss/postcss':
         specifier: ^4
         version: 4.2.1
+      '@testing-library/dom':
+        specifier: ^10.4.1
+        version: 10.4.1
+      '@testing-library/react':
+        specifier: ^16.3.2
+        version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.14))(@types/react@19.2.14)(react-dom@19.2.3(react@19.2.3))(react@19.2.3)
       '@types/better-sqlite3':
         specifier: ^7.6.13
         version: 7.6.13
@@ -132,6 +138,9 @@ importers:
       eslint-config-next:
         specifier: 16.1.6
         version: 16.1.6(@typescript-eslint/parser@8.57.0(eslint@9.39.4(jiti@2.6.1))(typescript@5.9.3))(eslint@9.39.4(jiti@2.6.1))(typescript@5.9.3)
+      jsdom:
+        specifier: ^30.0.1
+        version: 30.0.1(@noble/hashes@1.8.0)
       tailwindcss:
         specifier: ^4
         version: 4.2.1
@@ -140,7 +149,7 @@ importers:
         version: 5.9.3
       vitest:
         specifier: ^4.1.0
-        version: 4.1.0(@types/node@20.19.37)(msw@2.12.10(@types/node@20.19.37)(typescript@5.9.3))(vite@8.0.0(@types/node@20.19.37)(esbuild@0.25.12)(jiti@2.6.1)(yaml@2.9.0))
+        version: 4.1.0(@types/node@20.19.37)(jsdom@30.0.1(@noble/hashes@1.8.0))(msw@2.12.10(@types/node@20.19.37)(typescript@5.9.3))(vite@8.0.0(@types/node@20.19.37)(esbuild@0.25.12)(jiti@2.6.1)(yaml@2.9.0))
 
 packages:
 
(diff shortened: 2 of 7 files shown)


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/composer-refusal.test.tsx` |

lennons301-reviewer (approved):
Approving. The change does what #149 asked, all of it, and its objective signals are clean when run locally (this repo's CI is deploy-only, so there were no PR checks to read).

**Verified locally at `de176a2`**
- `pnpm vitest run` — 41 files, 865 tests, all passing, including the new `composer-refusal.test.tsx` under the per-file jsdom pragma.
- `npx eslint` on all five changed source/test files — clean.
- `npx tsc --noEmit` — only the two pre-existing errors in `src/lib/docker/__tests__/container-manager.test.ts`, a file this PR does not touch.
- `pnpm build` — succeeds.
- `pnpm-lock.yaml` is in sync with `package.json` (importer specs `^10.4.1` / `^16.3.2` / `^30.0.1` all resolve), so the deploy's `pnpm install --frozen-lockfile` will not trip; the additions are purely additive plus a vitest peer-key rewrite.

**Against the ticket**
- *Surfaces why completion was refused* — `completionRefusal` (`src/lib/chat/composer.ts`) replaces the bare early return; `handleComplete` writes the reason to the note line instead of returning silently. Keying `REFUSALS` over `Exclude<ComposerPhase, "idle">` is the right shape: a wordless refusal is unrepresentable, and the exhaustive table means a new phase is a type error rather than a silent `undefined`.
- *The row does not stay open where confirming does nothing* — the `useIsomorphicLayoutEffect` closes it before paint, and the deps (`confirmingComplete`, `refusal` — a string, so referentially stable) make it a one-shot with no loop: the closing render short-circuits on `!confirmingComplete`. The re-check kept inside `handleComplete` is genuinely belt-and-braces for a click landing in the same commit, and is documented as such.
- *The reason is legible* — the status line comes back with the row, the sentence lands under the field, and the disabled button's `title` now reads the same value, which removes the old contradiction where a blocked task was told "once the agent is idle".
- *A unit test covers the refusal path* — I traced the coverage claim rather than taking it: tests 1–5 in `composer-refusal.test.tsx` all assert state that only the layout effect produces (row gone, note present, description text, focus on the textarea, timed hand-back), so deleting the effect body fails them. The per-phase table in `composer.test.ts` pins each sentence to the state it explains, and the `completionRefusal(s) === null` ↔ `s.canComplete` assertion keeps the two predicates from drifting.
- *Typo* — `turn-manager.ts:629` fixed as asked, nothing else in that file.

**On the new dev dependencies.** The judgement call flagged in the PR body is the right one. Without a DOM the effect this ticket exists to add was untestable by construction (`composer-render.test.tsx` is `renderToStaticMarkup`, where effects never run), and an AC asking for the confirmation-open transition can't be met by testing the pure lookup alone. They are devDependencies added through pnpm with the lockfile committed (`choices/dependency-management.md` §Conventions), the run stage copies only the standalone build so nothing reaches the runtime image, and RTL's auto-cleanup is already satisfied by `globals: true` in `vitest.config.ts`. The one cost a human merging this should be aware of: the Dockerfile's `deps` stage installs devDependencies, so the VPS deploy build now also fetches jsdom and its tree — disk and install time, not the build-stage heap that has OOM'd before.

**Gating.** Correct as it stands: `src/components/**` matches the estate `visual-ui` gate, the PR carries `human-signoff`, and auto-merge is off. I have not changed either.

Suggestions, none blocking and none of them a documented-standard breach:

1. **Error messages went from assertive to polite.** Folding the two lines into one element means one role, and `role="status"` is the coherent choice for an element that normally holds the keyboard hint — but the repo's convention for error text elsewhere (`project-list.tsx`, `task-feed.tsx`, `confirm-strip.tsx`, `kill-switch.tsx`, `fleet-bits.tsx`) is `role="alert"`. Worth a line in the module comment noting the deliberate trade, so the next reader doesn't "fix" it back.
2. **The live region now announces the hint too.** Because the region is permanent, the swap back after `NOTICE_MS` re-announces "enter to send · shift+enter for a newline", and on touch the mount-time `enterSends` flip changes it once more. Harmless, but it is new noise that arrived with the role.
3. **`canComplete` has no remaining consumer.** `completionRefusal` is now the only thing the composer reads; `canComplete` survives on `ComposerState` used by tests alone. Two spellings of one predicate is a small drift surface — deriving one from the other (or dropping the field) would close it. The `=== null` assertion means drift would at least be caught.
4. **`useReturnFocus` landed on `main` after this branch forked** (`ab4853d`, `src/lib/use-return-focus.ts`) for exactly the "confirmation strip unmounts its trigger" shape. It is *not* a drop-in here — the complete button is disabled in the refusal case, so focusing it would drop focus anyway, and the textarea is the better target — but the confirmation's *cancel* path still drops focus to `<body>`, which is what that hook exists to fix. A follow-up ticket, not this one.

The deliberate non-fix is also right: a latched `completing…` after a successful POST is pre-existing and is the same defect class #151 owns.

PR #156, opened 2026-08-18
Two highlight.js cores resolve in the live view — a runtime-only failure

### LINKED ISSUE #150: Two highlight.js cores resolve in the live view — a runtime-only failure
## Parent

#121 — raised by its review pass.

## What to build

The live view's syntax highlighting resolves **two different highlight.js cores**. The direct grammar imports pull 11.12.0; `rehype-highlight` → `lowlight` carries 11.11.2.

It works today, and it works for a reason that will not hold forever: grammar modules are standalone, so a grammar compiled against one core runs fine on the other. The moment a grammar uses a helper introduced in a newer core, it fails **at runtime, in the browser, on a rendered transcript** — with no compile-time signal and no test that would catch it, because the failure depends on which grammar a particular agent message happens to contain.

That is a bad failure shape for a screen whose whole job is to be readable: a transcript renders fine for months and then one message with a Rust block blanks the code area.

Pin to the version lowlight resolves, so there is one core.

## Acceptance criteria

- [ ] The dependency tree resolves exactly one `highlight.js` version (`pnpm why highlight.js` shows no skew)
- [ ] The version chosen is the one `lowlight` depends on, so `rehype-highlight` is not forced onto an unexpected core
- [ ] Highlighting still renders for every grammar the live view registers, covered by the existing render tests

## Blocked by

- #142 — touches the same live-view dependency surface; sequencing avoids a lockfile conflict

lennons301:
Closes #150

The live view resolved two highlight.js cores: the grammars are imported directly (`^11.12.0` → 11.12.0) while `rehype-highlight` ran them on the 11.11.2 its own `lowlight` resolves.

## The fix

`lowlight@3.3.0` declares `highlight.js: ~11.11.0`, so 11.12.0 can never satisfy it — the skew was **structural**, not a stale lockfile. `package.json` now tracks the same range (`~11.11.0`), which resolves 11.11.2 for both importers and keeps them deduping through future patches rather than drifting apart again on the next `pnpm install`.

```
$ pnpm why highlight.js
highlight.js@11.11.2
├── interlude-scaffold@0.1.0 (dependencies)
└─┬ lowlight@3.3.0
  └─┬ rehype-highlight@7.0.2
    └── interlude-scaffold@0.1.0 (dependencies)

Found 1 version of highlight.js
```

## Tests

Two gaps the pin alone leaves open, both closed here:

- **Every registered grammar now runs.** The render tests covered 3 of the 14 grammars the pipeline registers, so AC 3 ("covered by the existing render tests") was not actually true. The snippet table is keyed off the module's own list (`REGISTERED_LANGUAGES`, exported for this), so a grammar added without coverage fails. The load-bearing assertion is a token span — `hljs` alone proves nothing, since `rehype-highlight` adds it *before* it tries the language.
- **The lockfile invariant is asserted.** The skew is recorded in `pnpm-lock.yaml`, so that is where it is checked: our resolution, lowlight's resolved edge, and the installed package set must all name one core. Verified to fail on the pre-fix lockfile with `expected '11.11.2' to be '11.12.0'`. This is the test that would have caught the original defect, and the one that catches a future `lowlight` bump moving its range out from under ours.

Rationale is recorded at the site (`markdown.ts` header) and in the `CLAUDE.md` convention line, since a routine `pnpm update` is the obvious way to undo this.

## Checks

`pnpm test` 910 passed (43 files) · `pnpm lint` clean on changed files · `pnpm build` exits 0.

`tsc --noEmit` reports 2 errors in `src/lib/docker/__tests__/container-manager.test.ts` — pre-existing on `main`, unrelated, left alone.

## Self-review — findings I did not take

Ran `/code-review` against `origin/main` with #150 as the spec. Acted on the lockfile-parsing, class-ordering and masked-skew findings (third commit). Declining these, with reasons:

- **"Use `pnpm.overrides` for a structural guarantee."** Overrides would *force* `lowlight` onto a core it may not declare — precisely the hazard this ticket describes, inverted — and AC 2 explicitly asks that `rehype-highlight` is "not forced onto an unexpected core". Mirroring the range and failing a test on drift surfaces the problem; an override hides it. Related sub-point taken: the guard now asserts *which* core, not just that there is one.
- **"The runtime coverage assertion is dead weight; the `Record` type already enforces it."** Doesn't hold in this repo — nothing that runs typechecks test files (`pnpm build` doesn't, and `tsc --noEmit` is already red on `main`), so the type is editor-only. I removed the assertion, verified the premise was wrong, and put it back with a comment saying why.
- **`process.cwd()` for the lockfile path.** Matches the adjacent `globals.css` test in the same block; diverging for one of two neighbours reads worse than the marginal robustness gain. `pnpm test` runs from the repo root.

Worth separate tickets, deliberately out of scope here:

- **`shell` and `console` fences render unhighlighted.** They are separate highlight.js grammars, not `bash` aliases, and agents fence shell output that way constantly. A real gap in the #121 registry, but a registry/bundle-weight decision rather than a core-skew one.
- **CI never runs `pnpm test`.** `deploy.yml` is the only workflow; it builds and deploys on push to `main`, so this new guard (and every other test) depends on a human or agent running the suite. That is a repo-wide gap, not something #150 should decide.
- **No `dependabot`/`renovate` config,** so nothing will surface that 11.12.x exists or that `lowlight` has moved on. The tilde ceiling is what the ticket asked for; noticing when to lift it is unautomated.
- **`package.json`/`pnpm-lock.yaml` match no review gate,** so a dependency-tree change — including this downgrade — is auto-mergeable unattended.
- **`CLAUDE.md` vs `AGENTS.md`.** The estate standard puts technical context in `AGENTS.md`; this repo has none and keeps every convention in `CLAUDE.md`. I followed the local pattern rather than starting a migration inside this ticket.


lennons301-reviewer (approved):
## Approve

Verified independently against #150 (no CI runs on this repo, so all signals were reproduced locally in the worktree).

**Acceptance criteria — all four confirmed:**

1. *One `highlight.js`* — `pnpm why highlight.js` → `Found 1 version of highlight.js` (11.11.2). `node_modules/highlight.js` resolves to `.pnpm/highlight.js@11.11.2`, and the lockfile's `packages`/`snapshots` now carry a single `highlight.js@` entry.
2. *The version lowlight resolves* — `lowlight@3.3.0` declares `highlight.js: ~11.11.0`, so 11.11.2 satisfies both importers and `rehype-highlight` is not forced onto anything. Mirroring lowlight's own range rather than using `pnpm.overrides` is the right call here and matches AC 2's wording; the PR body's reasoning stands.
3. *Every registered grammar renders* — the ticket said this was already covered by existing tests; it wasn't (3 of 14). The new per-grammar loop is keyed off `REGISTERED_LANGUAGES` exported from the module itself, so the list can't drift, and the `<span class="hljs-` assertion is the right load-bearing one (`hljs` alone is added before the grammar is attempted). All 14 pass.
4. *Build and lint* — `pnpm build` exits 0. `npx eslint src/lib/chat/markdown.ts src/lib/chat/__tests__/markdown.test.ts` is clean; the 16 `pnpm lint` errors are the known pre-existing baseline on `main`, untouched here.

**Extra verification:**

- `pnpm test` → 910 passed / 43 files.
- The lockfile guard is a genuine regression test, not a tautology. I replayed its logic against `git show origin/main:pnpm-lock.yaml`: `ours: 11.12.0`, lowlight edge `11.11.2`, installed `[11.11.2, 11.12.0]` → fails. It also fails closed if either edge stops being recorded (`ours` would be `undefined`/throw; a missing lowlight snapshot makes the `toBe` comparison fail).
- `pnpm install --frozen-lockfile` succeeds with a clean tree, so the `Dockerfile`'s frozen install won't break on the downgrade.
- The downgrade doesn't step back across a security release — `pnpm audit` reports no `highlight.js` advisory (the advisories it does report are all pre-existing on `main` and unrelated).
- Changed paths (`package.json`, `pnpm-lock.yaml`, `CLAUDE.md`, `src/lib/chat/**`) match no glob in the estate default gates or `docs/agents/review-gates.yaml`, so the armed state is correct. Blast radius is confined to code-fence rendering in the transcript; `globals.css` and the `hljs-*` class contract are untouched.
- Rationale is recorded where a routine `pnpm update` would encounter it: the `markdown.ts` header and the #121 convention line in `CLAUDE.md`.

**Non-blocking notes, not grounds for changes:**

- The follow-ups listed under "worth separate tickets" in the PR body are correctly out of scope, but none of them are ticketed — they exist only in this PR description. Two are worth issues rather than prose: **CI never runs `pnpm test`** (a standing breach of `~/code/platform/standards/testing.md` §3 — `deploy.yml` is the only workflow, so this new guard, like every other test, depends on someone running the suite by hand), and **`package.json`/`pnpm-lock.yaml` match no review gate**, which is exactly the gate-tuning signal `standards/review-gates.md` asks to be encoded.
- Suggestion only: the guard's `startsWith("lowlight@")` takes the first matching snapshot, so a hypothetical second `lowlight` wouldn't be checked on that edge — the `installed` set assertion covers the case anyway, so this is cosmetic.
- Declining the "runtime coverage assertion is dead weight" self-review finding was right, and verifying the premise before restoring it is the correct way to have reached that.

PR #157, opened 2026-08-18
A stalled outbound call permanently leaks a queue slot — one phantom wedges the whole box

### LINKED ISSUE #151: A stalled outbound call permanently leaks a queue slot — one phantom wedges the whole box

lennons301:
Closes #151

A stalled outbound call could hold a queue slot forever, and with `CAPACITY_SLOTS=1` one phantom occupant wedged the whole box. Both halves the ticket asks for are here, either of which would have prevented the 2026-08-18 wedge.

## 1. A reservation cannot outlive its task

`processingTasks` was doing two jobs with one lifetime, and that is the bug:

- a **re-entrancy lock** — don't dispatch a task another poll already picked up, don't deliver a follow-up turn into a container whose current turn is still finishing;
- a **slot reservation** — stand in for a container that is being provisioned, so back-to-back polls can't overfill the box.

The first must be held until the driving promise settles: `startTask` marks its container idle *before* it commits, pushes and opens the PR, so releasing sooner would run two agent execs in one container. The second is over the moment the container registers. Split into `inFlightTasks` and `slotReservations`, so the slot-relevant one never depends on a promise settling — and `occupiedSlots()` additionally ignores any reservation whose task has reached a terminal status, which makes the count self-heal within one poll.

`releaseSpentReservations()` runs at the top of each poll and drops a slot reservation once its container registers (or its task finishes without one), plus any lock left behind by a promise that never settled — a finished task can neither be picked up again (pickup reads `queued` only) nor delivered into.

Deliberate deviation from the AC's wording, worth a reviewer's eye: **the lock is not released at registration**, only the slot reservation is. AC 3 reads "a `processingTasks` reservation is released once its task registers in `activeTasks`", and its rationale ("the reservation exists only to cover pickup → `activeTasks.set`") reads the comment at `queue.ts:30-35` but not the set's other two uses at lines 76/134/170. Releasing the lock at registration reopens the window described above: `startTask` sets `state = "idle"` and then does its post-turn commit, push, PR create and idle notification, so a poll landing in that window would exec a second turn into the same container and the same `--resume` session. The slot-accounting intent of the AC is fully met — a registered pickup holds no slot — and one test pins it.

## 2. Bounded outbound I/O

- **GitHub.** `@octokit/request` v10 has no timeout option (it forwards only `fetch` and `signal` from `request`), so the bound is a `request.fetch` wrapper — the better seam anyway: a per-request deadline, so a paginated read isn't racing one budget across every page, and the signal covers reading the body, not just the headers. It both aborts the connection and releases the caller, so the bound holds even for a fetch that ignores signals. One factory (`createOctokit`) builds every client — App, installation, reviewer — and an eslint rule bans `new Octokit(` anywhere else. Timeouts are logged. 10s per request; Octokit's retry plugin will re-issue a timed-out request up to three times, so a black-holed endpoint costs about a minute, not the life of the process.
- **Discord.** Bounded at the two primitives every notification goes through (the channel lookup and each send attempt) plus the digest's history read, on the same primitive. A bound is needed because nothing above `@discordjs/rest`'s own 15s per-attempt abort is bounded: it retries that attempt, sleeps out a 429's `Retry-After`, and `sendWithRetry` retried again on top. **A timeout is now terminal for the send** — the attempt we walked away from may still be delivered underneath us, so retrying it is how one notification becomes two embeds. A dropped push notification is the better failure; the dashboard carries everything Discord pushes.
- One primitive, three policies: `raceWithTimeout` (`src/lib/timeout.ts`) returns a sentinel and each caller decides what a timeout means — the Docker admission probe still fails *open* (#115), a Discord call fails *closed*, a GitHub request additionally aborts. This replaced what was becoming a third hand-rolled copy of the same race; #128's probe now shares it, with its tests unchanged.

## 3. The draft-PR decision: implemented

The ticket asks this be implemented or explicitly declined — **implemented.** A branch level with its base cannot carry a PR (GitHub 422s "No commits between …"), so a grilling session, which never commits, re-attempted that doomed call every turn: one roll of the dice per turn, which is why the grilling shape was the worst exposed. The push script now reports how far the branch is ahead of `origin/HEAD` — free in the container's full clone, no extra round trip — and `runPostTurnCommitAndPush` consults it via `shouldOpenDraftPr`. A count that can't be read (`unknown`) attempts the PR exactly as before.

## Tests

- `queue-slots.test.ts` drives the real poll loop against an in-memory migrated DB, with the turn manager stubbed only where a hang has to be reproducible. Three cases, and I checked each against the pre-fix code rather than assuming:
  - a hung post-turn GitHub call (a **real** request through this repo's client, bound set past the test's horizon) outlives its task → the slot is free and the next task dispatches. Fails against the pre-fix accounting; pins the split.
  - a hung pickup whose task is cancelled → same. Fails with the split in place but the terminal-status skip removed; pins that skip.
  - a pickup's slot passes to the container it registers and is free once the pass parks, though the driving promise never settles; pins the registration clause.
- `request-timeout.test.ts` + `client.test.ts`: a stalled request rejects within the bound, aborts rather than abandons its socket, logs, composes a caller's own signal, and a client built by the factory fails a never-answering request.
- `notifications.test.ts`: a stalled channel lookup and a stalled send both settle, and a timed-out send is not retried.
- `timeout.test.ts`, `container-manager.test.ts` (script marker + `parseCommitsAhead`), `pull-requests.test.ts` (`shouldOpenDraftPr`).

On the AC "a test proves a hung post-turn GitHub call cannot hold a slot after the task completes": the first queue case now *is* that test — it hangs on a real GitHub request rather than a stand-in — and the request-timeout tests separately prove such a call would no longer hang in production at all.

## Notes for the reviewer

- **Gated paths → `human-signoff`, never auto-merge:** `src/lib/github/client.ts` (credentials), `src/lib/docker/container-manager.ts` (agent-sandbox). The label is on the PR for that path match alone — no review has happened yet; the ticket-reviewer with fresh eyes is still the thing that approves.
- **Deliberate collateral, disclosed:** 17 notify\* helpers now share one bounded channel lookup instead of each repeating the fetch-and-narrow dance — return values are identical; the only behaviour change is that a channel that resolves to a non-text channel now logs instead of failing silently. `isTerminalTaskStatus` moved from `chat/composer` to `src/lib/tasks/status.ts`: the orchestrator had come to depend on a view state machine for a task-domain predicate.
- **A hole the split exposed, now closed:** step 2 checks the task's status itself, so a stranded `activeTasks` entry (a status written without touching the map — a reaped review, a boot sweep) is never drained every poll, starting and stopping a parked container for a turn that breaks immediately. That was possible before this ticket too.
- Self-review ran `/code-review` (standards + spec) against `origin/main` with #151 as the spec; its findings are what the last three commits act on. The one finding I did not act on is the AC-3 wording deviation above, argued rather than silently ignored.
- Checks: `pnpm test` 951 pass, `pnpm build` clean, `pnpm lint` clean on every changed file (the repo's 16 pre-existing errors in `custom-server.js` / `preview-pane.tsx` / `output-parser.test.ts` are untouched).


lennons301:

| gate | glob | path |
|---|---|---|
| credentials | `src/lib/github/client.ts` | `src/lib/github/client.ts` |

lennons301-reviewer (approved):
Approving. Gated (`human-signoff`, auto-merge off) on the `credentials` (`src/lib/github/client.ts`) and `agent-sandbox` (`src/lib/docker/**`) globs in `docs/agents/review-gates.yaml` — my approval satisfies branch protection only; a human merges. I have not touched the label or auto-merge.

## Objective signals

No CI runs on PRs in this repo (only `deploy.yml` on main), so I ran them locally on the branch head:

- `pnpm test` — 48 files, **951 passed**
- `pnpm build` — clean
- `npx eslint` on all 22 changed files — clean (main's 16 pre-existing errors in `custom-server.js` / `preview-pane.tsx` / `output-parser.test.ts` are untouched, per the repo's lint baseline)

## Against the ticket

All seven acceptance criteria are met.

**Bounded outbound I/O.** `createOctokit` is now the single construction site and I verified it: `grep -rn "new Octokit("` across the repo returns exactly `client.ts:18`, and the `no-restricted-syntax` rule in `eslint.config.mjs` keeps it that way — enforcement in the tool rather than a source-text test is the right call. The `request.fetch` seam is the correct one: `@octokit/request` v10 genuinely has no timeout option, and because `@octokit/core` builds `graphql` from the same `request` instance, the four `octokit.graphql(...)` call sites in `pull-requests.ts` inherit the bound too — worth stating, since those were as unbounded as the REST ones. `AbortSignal.any` composition means the bound layers on a caller's signal rather than replacing it, and that is tested. Discord is bounded at `fetchTextChannel` and each `channel.send`, so all 17 `notify*` helpers and the digest are covered.

**A reservation cannot outlive its task.** The split is the right cut, and the pairing that makes it safe is the part I checked hardest: dropping the `inFlightTasks` lock for a terminal task would be dangerous *without* the new `taskIsFinished` guard in step 2, and with it neither re-entry path exists (pickup reads `status = queued` only; delivery now refuses a terminal task whatever its `activeTasks` entry says). A side effect the PR body doesn't claim but which I think is the most valuable part of the change: because slot accounting no longer waits on any promise, an unbounded *Docker* exec hang — `execFallbackCommitAndPush` is still unbounded, correctly out of scope — can no longer leak a slot either. The fix is not dependent on having found every unbounded call.

I checked the tests reproduce the incident rather than restate the fix. Against the pre-fix accounting, `queue-slots.test.ts` case 1 fails as claimed: `processingTasks` holds the delivery, `completeTask` removes the `activeTasks` entry, `!active.has(taskId)` counts the phantom, one slot is full and `turns.dispatched` stays empty. Case 2 pins the terminal-status skip specifically. Hanging case 1 on a real Octokit request through this repo's own client (bound pushed past the test horizon) rather than a stand-in is the right instinct.

**Draft-PR decision: implemented, and the reasoning holds.** Reporting commits-ahead from the push is free given `git clone "$GIT_URL"` is a full clone, so `origin/HEAD` resolves. Two details I checked and liked: the `|| echo unknown` path parses to `null`, not `0`, so an unreadable count attempts the PR exactly as before (no new silent-skip failure mode); and `parseCommitsAhead` taking the **last** marker means remote push chatter cannot pre-empt the echo that runs after it. A stale `origin/main` can only inflate the count, which errs toward attempting.

**AC-3 wording deviation.** I read the argument independently and agree with it. Releasing the re-entrancy lock at container registration would open exactly the window described — `startTask` sets `state = "idle"` before its commit/push/PR/notify tail — and a poll landing there would exec a second turn into the same `--resume` session. The AC's *slot-accounting* intent is fully met: a registered pickup holds no slot. Arguing it in the PR body rather than silently diverging is the right way to handle it.

**Gated-path changes are minimal.** `client.ts` gains a factory and a fetch wrapper — no change to token minting, caching, or the reviewer-PAT boundary. `container-manager.ts` adds one `echo` to the push script and a parser; no `Bind`, no privilege or network change, and the existing test still asserts the script leaks no token names.

## Notes (non-blocking, no change requested)

1. **Step 3 didn't get step 2's guard.** `queue.ts:265-276` (the ~30s dev-server scan) checks `inFlightTasks` and `isParked` but not `taskIsFinished`. Now that the lock is released for a terminal task, a stranded `activeTasks` entry for a completed interactive task can be port-scanned every 30 cycles where previously the held lock suppressed it. Harm is a caught `console.error` against a gone container, but it's the one place the new invariant isn't applied consistently.
2. **Case 3 is weaker than the body claims.** `occupiedSlots` already `continue`s on any reservation whose task is in `activeTasks`, so "hands the slot from a pickup to the container it registers" would still pass with the release-at-registration clause in `releaseSpentReservations` removed. The clause is real and worth keeping (set hygiene, and it's what AC 3 asks for literally), but it isn't pinned by that test.
3. **`hasDigestPostedSince`'s bound has no direct test** — the primitive and the pattern are tested, and I confirmed the digest's `RETRY_MS` loop is idempotent through the history check, so a timed-out-but-landed digest send is deduped rather than double-posted. Correct as-is; just untested at that call site.
4. **Suggestion, out of scope:** `src/lib/discord/client.ts:205,218` still inline `["completed","failed","cancelled"].includes(...)` — the fourth restatement of what this PR just gave a home in `src/lib/tasks/status.ts`. Those `message.react` awaits are also unbounded, but they sit in gateway handlers after the DB write, not in a slot-holding path, so nothing is lost by leaving them.
5. Both new conventions are recorded in `CLAUDE.md` in the house style, and the `isTerminalTaskStatus` move plus the `fetchTextChannel` de-duplication are disclosed collateral I'd have flagged as Duplicated Code / misplaced-responsibility if they hadn't been done.

PR #139, opened 2026-08-14
Global autonomy kill switch (backend + dashboard)

lennons301:
Closes #118

A global autonomy kill switch: one durable flag that stops all new autonomous pickup at the next sweep tick, with no restart, and says so on the dashboard so a held fleet is never mistaken for an idle one.

## What landed

**A durable `settings` row** (`src/db/schema.ts`, migration `0021_settings`) — one row, id `fleet`, holding `globalAutonomyPaused`. This is the runtime half of configuration: env config is fixed at boot, this is flipped by a human while the fleet runs. `src/lib/settings.ts` reads it fresh at every call site, never caches, which is what makes a flip take effect at the next tick rather than the next restart. Being a row, an engaged switch survives a restart — and boot logs that, so a deliberately held fleet doesn't read as a broken one.

**`decideNext` holds pickup and nothing else.** The gate sits at the daily cap's own gate, so the two paused fleets behave identically: verdicts still post, PRs are still gated or armed, repairs still run, a burnt ticket still routes to a human, a running turn is left to finish. Triage pickup is held by the same flag one level up — a triage pass is autonomous pickup that takes a container and spends money. Both webhook fast paths call `runAutonomySweep()`, so there is no pickup path that bypasses the flag.

**`PATCH /api/settings/autonomy`** toggles it (`GET` reads it). Explicit boolean only, mirroring the rule for arming a project: a stray truthy string must never decide whether the fleet picks up unattended work. The response also carries `envMaster` (the `AUTONOMY_ENABLED` boot master) so #119's settings screen can tell "held by the operator" from "autonomy was never armed at all" rather than mislabelling an env-disarmed fleet.

**Dashboard.** The read model gained `pickupPaused: { reason, body } | null`; the live dot and banner read it. The kill switch outranks the daily cap when both hold — midnight lifts the cap, only a human lifts the switch. `LiveDot` gained a `held` state so a deliberate hold reads amber and a breached ceiling reads red, dot and banner agreeing. `spend.capPaused` is untouched, so the gauge and the daily digest are exactly as before.

`AUTONOMY_ENABLED=false` still prevents sweeps from starting at all — unchanged.

## Verification

- `pnpm build` ✅ · `pnpm lint` clean on every changed file ✅ · 676 tests pass (`npx vitest run`) ✅
[...]

PR #158, opened 2026-08-18
Fleet-health watchdog is blind to phantom slot occupancy — it trusts the counter that lies

### LINKED ISSUE #152: Fleet-health watchdog is blind to phantom slot occupancy — it trusts the counter that lies
## Parent

#126 built the fleet-health watchdog. #151 is a wedge it could not see.

## What to build

The pickup-wedged signal defines a wedge as *a free slot with queued work*: `slotFree = snapshot.slots.occupied < snapshot.slots.total`, and `queuedWhileSlotFree` is gathered only when that holds (`sweep.ts:305`, `:338`). But `snapshot.slots.occupied` is `occupiedSlots()` (`sweep.ts:629`) — the in-memory counter.

So when occupancy is itself the lie, the watchdog is silent **by construction**. In the #151 incident a phantom reservation held the box's only slot for ~1.5h: the dashboard showed `used: 0` (that surface reads live tasks from the DB), the queue refused every pickup, and `needsYou` stayed empty because from the watchdog's point of view the box was legitimately saturated. Exactly the invisibility #126 existed to remove, on the one input #126 trusted.

The fix is a fourth observation rather than a fourth mechanism: compare the in-memory count against reality — live agent containers (the daemon already answers this for the memory-admission probe) plus parked containers — and treat a sustained divergence as a wedge, with the same debounce, card and one-time Discord ping the other three signals use. The card should name the divergence (`occupancy says N, M containers live`) because the operator action is a restart, not a wait.

Note the two surfaces disagreed during the incident: the dashboard's `slots.used` derives from DB rows via `isLiveTask`, while the gate uses `occupiedSlots()`. That divergence is the signal; the dashboard should not silently show a free slot the queue will not use.

## Acceptance criteria

- [ ] A sustained divergence between `occupiedSlots()` and observed reality (live + parked agent containers) produces a pickup-wedged needs-you card and a single Discord ping, deduped like the existing signals
- [ ] The card states the divergence and that recovery is an app restart
- [ ] The signal clears automatically once occupancy and reality agree
- [ ] A task sitting `queued` past the threshold is surfaced even when `occupied == total`, provided occupancy cannot be corroborated by real containers
- [ ] No false alarm from the legitimate shapes: a parked autonomous container, a task mid-provisioning (reserved, container not yet created), a Docker probe that times out
- [ ] Threshold configurable, defaulting alongside the existing `pickupWedgedMs`
- [ ] Unit tests over the pure evaluator cover: phantom occupancy, provisioning race, parked container, probe failure

## Blocked by

None. Independent of #151's fix — that one removes this cause, this one makes the next unknown cause visible.

lennons301:
Closes #152

## What this does

The pickup-wedged signal decides a slot is free from `occupiedSlots()` — the in-memory counter. So when occupancy is itself the lie, the watchdog is silent by construction: in the #151 incident a phantom reservation held the box's only slot for ~1.5h, the queue refused every pickup, and `needsYou` stayed empty because from the watchdog's point of view the box was legitimately saturated.

A fourth observation, not a fourth mechanism. The sweep asks the daemon what is actually running (`observeAgentContainers`) and hands both numbers to the pure evaluator, which treats a sustained disagreement as the same pickup-wedged card and one-time ping the other three signals use. What differs is the remedy — a phantom slot lives in orchestrator memory, so nothing but a restart clears it — and the remedy is decided in the evaluator so the card, the Discord ping and the server log cannot advise differently.

Key decisions:

- **One-directional (`occupied > live`).** More containers than counted slots is a parked pass mid-transition or a leaked container — the memory-admission probe's business, never a pickup wedge.
- **A null census is *unknown*, and unknown decides nothing in either direction.** It may not manufacture a divergence (a zero census reads as "restart the app", so a daemon that cannot answer must say nothing rather than be misread as saying none), and it may not erase one either — see the self-review section below, which is where this half came from.
- **Queued work is gathered every sweep**, not only behind a free slot: whether a slot *reads* free is the fact under suspicion, and a task starved by a phantom is precisely what must still be named (AC4).
- **The card carries its cause**, and the ping re-fires on an *upgrade* only (`dispatch` → `phantom-slot`). A standing dispatch wedge that turns out to be a phantom pings again, because its advice changed from "go and look" to "restart the app"; a downgrade never pings, because it means we observed less, not that the fleet recovered.

## Threshold: 20 minutes, not 3

`OCCUPANCY_DIVERGED_MINUTES`, defaulting beside `PICKUP_WEDGED_MINUTES` in `budgets.ts`. Deliberately much longer than the 3-minute pickup debounce, and set by the worst *honest* case: a task holds its slot from the moment it is reserved, but its container does not exist until `createWorkspaceContainer` returns — and that call runs `ensureImage` inside itself, so the first task after a `Dockerfile.agent` change waits out a full agent-image build (apt, `gh`, a global npm install) on a 2-vCPU box before anything corroborates it. The cost of being wrong is asymmetric: this card tells the operator to restart, which would kill exactly that legitimately-provisioning task. 20 minutes clears a cold build with room to spare and still turns a 1.5h invisible wedge into a card and a ping with over an hour left.

Two follow-ups worth having, neither in scope here: warming the image at boot would move that build out of the dispatch path and let this threshold drop; and `snapshot.slots.occupants` already distinguishes reservation-only occupancy, so a shorter bound is reachable for the shapes that are *not* mid-provisioning.

## Where I read the spec differently

The ticket says to compare the counter against "live agent containers ... **plus parked containers**". The evaluator compares against `live` only; the parked/stopped count is observed and shown on the card but does not corroborate a slot. Adding parked to the reality side would *mask* a phantom rather than prevent a false alarm: `occupiedSlots()` already excludes parked entries (they are `docker stop`ped since #93), so a parked container cancels on both sides and can never read as a divergence — while counting it into the ceiling would hide a phantom sitting behind one, which is a live shape on a 1-slot box. Note this is not hypothetical: with `live + stopped`, the actual #151 shape (phantom reservation plus one parked pass) computes `1 > 1` and stays silent. The stricter comparison satisfies every acceptance criterion including the parked-container false-alarm case, which has a test.

On the ticket's closing note — "the dashboard should not silently show a free slot the queue will not use" — the disagreement between `slots.used` (DB rows via `isLiveTask`) and `occupiedSlots()` is now surfaced as a red needs-you card naming both numbers, so it is no longer silent. I deliberately did **not** re-derive `slots.used` from the in-memory counter: the DB-derived figure is the honest one, and overwriting the dashboard with the counter would hide the very evidence this signal exists to show. Both readings were confirmed by the previous review pass.

## Changes since the review that requested changes

All three requested fixes, plus what a fresh self-review pass turned up on top of them.

### The three requested fixes

1. **`runBoundedProbe` now wraps `raceWithTimeout`** instead of re-implementing it. `src/lib/bounded-probe.ts` is deleted and the wrapper lives in `src/lib/timeout.ts` beside the race, so the tree holds one implementation of the subtle part again — the reviewer was right that the copy had already drifted (no `timer.unref?.()`). It has its own unit tests now, beside `raceWithTimeout`'s: the two non-answers, a synchronous throw, and the case worth pinning down — a probe whose *answer* is a refusal still answered, so `ok` stays the outcome of asking rather than the content of the reply. The module docstring's "three callers" claim is refreshed and now names the census.
2. **CLAUDE.md updated**, extending the existing "Bounded outbound I/O" and "Queue bookkeeping" bullets in place rather than appending (agent-context.md §3). This repo keeps its context in CLAUDE.md and has no AGENTS.md — a pre-existing divergence from the standard that I have not tried to fix inside this ticket.
3. **Both stray EOF blank lines dropped.** `container-manager.test.ts` is now byte-identical to `origin/main`, so it leaves the diff and the gate table entirely; the `tsc` baseline errors at line 396 are now demonstrably outside this branch's diff rather than merely claimed to be.

### What the self-review pass then found

Ran `/code-review` against `origin/main` with #152 as the spec. Its Standards and Spec axes independently found the same defect, and it was a real one:

- **A census that could not be taken erased the divergence clock.** `occupancyDivergedSinceMs` cleared whenever the census came back null, conflating the daemon *agreeing* with the counter and the daemon failing to answer. The census has a 5s bound and runs once per 30s sweep, so a 20-minute threshold needed forty consecutive answers — and a daemon degraded enough to time out every few minutes is the #125/#128 shape and the likeliest companion of a real phantom. It would have reset the clock on every hiccup and the card would never have fired. Now only positive agreement stops the clock, which is what AC3 asks for ("once occupancy and reality agree" — agreement is an observation). Firing still demands a *current* positive divergence, so a daemon that never answers still never raises this card; that fail-safe is unchanged.
- **The ping could downgrade its own advice.** `PickupWedgeCause`'s docstring already promised a re-ping on an *upgrade*, but the code re-pinged on any change of cause. With a census momentarily unavailable and a dispatch wedge standing, the card demoted to `dispatch` and pinged again telling the operator to "go and look" at a box that still needed restarting. The memory now keeps the highest cause an occurrence reached, and the ping fires on rank increase only. Three new tests cover the hiccupping daemon, agreement-clears-but-unknown-holds, and the no-downgrade-ping case.
- **One Docker-probe bound, not two matching literals.** The census asserted "same bound as `ADMISSION_PROBE_TIMEOUT_MS`" while being a second `5000` that nothing kept equal. `capacity.ts` already imported from `agent-containers.ts`, so the admission probe now derives from the shared constant with no new dependency edge and no other file touched.
- **Two CLAUDE.md claims I had just written were false**, which is the same class of defect the reviewer raised in finding 2 — so worth naming rather than quietly fixing. I had described the comparison as `slots.used` vs `occupiedSlots()` when the implemented comparison is the counter vs the daemon census; and the `unref` note described a draft not in the tree, which is exactly the changelog-in-a-context-file that §3 forbids. Both reworded.
- **One comment said "when both hold" where the code means "once it fires".**

Findings I did **not** act on, with reasons:

- *"`remedy: string` is prose in the reducer — derive it from the discriminant"*. Kept both, as accepted in the last review pass. `detail` already sets the precedent for prose in this type, and the alternative puts a cause→remedy map in three renderers. The fair sub-point — that `fleet-view.test.ts` hardcodes remedy prose, so a change to `PHANTOM_SLOT_REMEDY` is only caught by `health.test.ts`'s `toContain("restart")` — is real but very low value, and touching those assertions now would add churn to a diff a human is about to read.
- *"`AgentContainerCensus` belongs with the observer, not in `health.ts`"* — it sits with `OwedReviewObservation` and `QueuedTaskObservation`, likewise produced by the sweep and likewise declared in `health.ts`. The input contract belongs to the pure evaluator; the docker module imports it type-only, so there is no runtime edge. Accepted in the last pass.
- *Two daemon listings per sweep* (census plus admission probe) — left alone. Different call sites at different cadences, and changing what the admission probe asks the daemon has its own incident history (#93/#125) that this ticket has no reason to disturb.
- *The dashboard still shows a DB-derived `slots.used`* — deliberate, argued above and agreed in the last pass.
- The previous reviewer's non-blocking note that `gatherQueuedDispatchable()` now `.all()`s the `runs` table every sweep and filters in JS — agreed, and left as the obvious future place to push the filter into SQL. Trivial at current table size.

## Testing

- `pnpm test` — **984 pass / 50 files.** Unit tests over the pure evaluator cover all four shapes the ticket names (phantom occupancy, provisioning race, parked container, probe failure) plus over-count, clearing, starved-work-at-full-occupancy, the upgrade re-ping, and the three new cases above. `observeAgentContainers` has its own tests for the census split, the daemon error and the hang; `runBoundedProbe` now has its own too.
- `pnpm build` — compiles. `npx eslint` — clean on all changed files.
- `npx tsc --noEmit` reports 2 errors in `src/lib/docker/__tests__/container-manager.test.ts`; that file is now byte-identical to `origin/main`, so they are provably the existing baseline and not this diff.

## Note for the reviewer

This touches three gated paths — `src/lib/docker/**` (agent-sandbox), `src/lib/config.ts` (credentials) and `src/lib/orchestrator/autonomy/**` (autonomy-control) — so it is `human-signoff` by construction and must not auto-merge.


lennons301:

| gate | glob | path |
|---|---|---|
| agent-sandbox | `src/lib/docker/**` | `src/lib/docker/__tests__/agent-containers.test.ts` |
| agent-sandbox | `src/lib/docker/**` | `src/lib/docker/agent-containers.ts` |

lennons301-reviewer (dismissed):
Approving. Gated for human sign-off (`human-signoff` label present, auto-merge off) — this approval satisfies branch protection only; a human merges.

## Objective signals

No CI runs on PRs in this repo (`.github/workflows/` holds only `deploy.yml`), so I ran everything locally on `agent/issue-152`:

- `pnpm test` — 45 files, 937 tests pass
- `npx eslint` on all 15 changed source files — clean
- `pnpm build` — compiles
- `npx tsc --noEmit` — 2 errors, both in `src/lib/docker/__tests__/container-manager.test.ts:360`, on lines this PR does not touch (its only change to that file is a stray trailing blank line). Confirmed as the pre-existing baseline the PR body claims.

## Against the ticket

All eight acceptance criteria are met, and each of the four named shapes has a test over the pure evaluator:

- AC1/AC3 — sustained `occupied > live` raises the pickup-wedged card plus one Discord ping, and clears when the counter and daemon agree (`health.test.ts`, "fires — card + one-time ping…", "…clears once the count is real again")
- AC2 — `PHANTOM_SLOT_REMEDY` names the divergence and says restart; it is produced in the evaluator and consumed identically by the card (`fleet-view.ts:538`), the ping (`notifications.ts`) and the server log (`sweep.ts:284`), so the three cannot drift
- AC4 — `queuedDispatchable` is gathered unconditionally now, so a task starved behind a phantom is named even at `occupied == total`. I checked the behaviour is otherwise unchanged: `wedgedNow` still requires `slotFree`, so the old `queuedWhileSlotFree` gating was redundant, and the two extra SQLite reads per 30s sweep are trivial
- AC5 — provisioning race, parked container and probe failure each have a test, and I verified the underlying invariant rather than taking the tests' word for it: `occupiedSlots()` counts non-parked `activeTasks` entries plus `processingTasks`, `isParked` is `idle && kind !== "interactive"`, and `parkContainer` is the only stop path for a live entry (`cancelTask` deletes the entry as it stops). So a parked pass is absent from both sides and an interactive idle session is present on both — the comparison is genuinely balanced, not accidentally so
- AC6 — `OCCUPANCY_DIVERGED_MINUTES` defaults beside the other three in `budgets.ts`, threads through `config.ts`, and is documented in the runbook's env table

## Judgement calls I checked rather than accepted

**The 20-minute threshold.** The justification rests on `ensureImage()` running inside `createWorkspaceContainer` — verified at `container-manager.ts:228`, so the first task after a `Dockerfile.agent` change is legitimately uncorroborated for a full cold image build. Given the remedy is "restart the app", which would kill exactly that task, erring long is the right asymmetry, and 20 minutes still turns #151's ~1.5h into a card with over an hour to spare. The reasoning is recorded at the constant, not just in the PR body.

**The stated deviation from the ticket** (comparing against `live` only, not `live + parked`) is correct and I would have flagged the ticket's wording as the bug. `occupiedSlots()` already excludes parked entries, so parked cancels on both sides; adding it to the reality side could only mask a phantom sitting behind a parked container — a live shape on a 1-slot box.

**Not re-deriving `slots.used` from the counter** is also the right call: the DB-derived figure is the honest one, and overwriting it with the counter would hide the evidence. The ticket's "should not silently show a free slot" is satisfied by the divergence being loud rather than by making both surfaces agree on the wrong number.

**One-directional comparison** — over-count is left to the memory-admission probe. I also note the read skew works in the signal's favour: `slots.occupied` is sampled at gather time and the census after acting, so a transient over-count resets the timer rather than accumulating toward a false alarm.

**`runBoundedProbe` extraction** removes a verbatim duplicate of the timeout race and returns an outcome rather than a fallback, which is right given the two callers want opposite non-answers (fail open vs. say nothing). `AGENT_CONTAINER_NAME_PREFIX` now has one home that the census, the admission probe and the reaper all read. Both are real duplication removals, not churn.

Smell baseline: nothing to raise. The `AgentContainerCensus`-in-`health.ts` placement is consistent with `OwedReviewObservation`/`QueuedTaskObservation` and the import is type-only; the `cause` discriminant plus prose `remedy` is one map, not three.

## Non-blocking observations

1. `src/lib/docker/container-manager.ts` and its test file each gained a stray trailing blank line (leftover from moving the census out into its own module). Noise in a gated path; worth dropping if the branch is touched again.
2. A single failed or timed-out census resets `occupancyDivergedSinceMs`, so a flaky daemon probe restarts the 20-minute clock rather than merely pausing it. Fail-safe in the right direction and consistent with #128, and a daemon sick enough to do this repeatedly should trip the queue-heartbeat signal instead — noting it only so the interaction is on record.
3. Suggestion (mine, not tied to a standard or the ticket): the corroboration is count-vs-count, so a leaked *running* agent container that the reaper is exempting (#24) could inflate `live` and mask a genuine phantom. Container names embed the task id and `snapshot.slots.occupants` is already built from the active entries, so a set comparison would be strictly stronger. Out of scope here — the ticket asked for the count.

For the human merging: the three gated paths (`src/lib/docker/**`, `src/lib/config.ts`, `src/lib/orchestrator/autonomy/**`) are touched only in ways I would characterise as narrow — a name constant and a new read-only census under `docker/`, one env-var read under `config.ts`, and the watchdog evaluator plus its sweep wiring under `autonomy/`. Nothing changes what may be claimed, gated, armed or merged, and no container creation, mount, network or resource-limit behaviour changes.

lennons301:

| gate | glob | path |
|---|---|---|
| agent-sandbox | `src/lib/docker/**` | `src/lib/docker/__tests__/agent-containers.test.ts` |
| agent-sandbox | `src/lib/docker/**` | `src/lib/docker/agent-containers.ts` |

lennons301-reviewer (changes requested):
## Request changes — two findings tied to written standards, both small

The feature itself is well built and I verified it satisfies every acceptance criterion on #152. Objective signals are green locally (CI reports no checks on this branch, so I ran them):

- `pnpm test` — 975 pass / 50 files
- `pnpm build` — compiles
- `npx eslint` on all 16 changed files — clean
- `npx tsc --noEmit` — 2 errors, both in `src/lib/docker/__tests__/container-manager.test.ts` at line 396; I diffed that hunk against `origin/main` and it is byte-identical, so the baseline claim in the PR body checks out

I also verified the reasoning behind the deliberate spec divergence (comparing against `live` only, not `live + stopped`). `occupiedSlots()` in `src/lib/orchestrator/queue.ts` counts non-parked `activeTasks` entries plus unregistered reservations, so a parked pass really is absent from *both* sides of the comparison and cancels out — the stricter comparison is correct, and adding `stopped` to the reality side would indeed mask a phantom sitting behind a parked container on a 1-slot box. Good call, and the false-alarm test for it is honest. The one-directional check, the null-is-unknown discipline, and gathering queued work every sweep are all right, and the `cause` discriminant is the correct fix for the shared dedup flag.

What I am asking for:

### 1. `runBoundedProbe` is a second copy of a primitive CLAUDE.md declares singular

`CLAUDE.md` (line 42, "Bounded outbound I/O (issue #151)") states:

> Every outbound call the orchestrator awaits carries a wall-clock bound built on **one primitive**, `raceWithTimeout` (`src/lib/timeout.ts`), which returns a sentinel so each caller keeps its own policy — the admission probe fails *open* (#115) ...

`src/lib/bounded-probe.ts` re-implements that race from scratch — its own `Symbol("bounded-probe-timeout")`, its own `Promise.race`, its own verbatim copy of the "keep a losing promise's rejection from surfacing as an unhandled rejection" comment — and this PR then migrates `checkMemoryAdmission` off `raceWithTimeout` onto the copy. After this change the tree holds two independent implementations of the subtle part, which is precisely the drift the convention exists to prevent, and `src/lib/timeout.ts`'s own docstring ("Three callers so far — the Docker admission probe, a Discord REST call, an outbound GitHub request") is no longer true.

The drift is already observable: `raceWithTimeout` calls `timer.unref?.()` with the comment "Never a reason to hold the process open on its own". `runBoundedProbe` does not, so a probe in flight can now hold the event loop open for up to its bound. Small in effect, but it is exactly the class of divergence one-primitive rules are for.

The PR body's rationale — "It returns the outcome rather than a fallback, because the two callers want opposite things from a non-answer" — does not distinguish it from the existing primitive: `raceWithTimeout` returns a sentinel *for that same reason*, stated in its own docstring. The only thing new here is folding the try/catch into the return shape, which composes in a few lines:

```ts
try {
  const result = await raceWithTimeout(probe(), timeoutMs);
  return result === TIMED_OUT
    ? { ok: false, reason: "timeout" }
    : { ok: true, value: result };
} catch (error) {
  return { ok: false, reason: "error", error };
}
```

Please build `runBoundedProbe` on `raceWithTimeout` (or export it from `src/lib/timeout.ts` beside it) so one race remains. Note also that `raceWithTimeout` has a direct unit test (`src/lib/__tests__/timeout.test.ts`); the new primitive has none of its own and is exercised only indirectly through `agent-containers.test.ts`.

### 2. `CLAUDE.md` is not updated, and this PR makes part of it false

Estate standard `~/code/platform/standards/agent-context.md` §2 requires the context file to be kept current "as part of the normal development process", and this repo has no hook to do it — the last twelve feature merges on `main` each carried their own CLAUDE.md edit (#157, #154, #156, #153, #147, #144, #139, ...). This PR updates `docs/runbook.md`, which is good, but leaves CLAUDE.md untouched while:

- adding a fourth watchdog observation, the corroboration discipline, and a new tunable (`OCCUPANCY_DIVERGED_MINUTES`) that the existing #126/#151 paragraphs do not mention; and
- invalidating the "one primitive, `raceWithTimeout`" sentence quoted above.

Per §3 of that standard this should extend the existing "Bounded outbound I/O" / queue-bookkeeping paragraphs rather than being appended. This matters more than usual here: CLAUDE.md is the whole briefing the next agent gets, and right now it would tell that agent the admission probe uses `raceWithTimeout` when it no longer does.

### 3. Nit — two stray trailing blank lines in gated files

`src/lib/docker/container-manager.ts` and `src/lib/docker/__tests__/container-manager.test.ts` each gain a blank line at EOF; for the test file that is the *only* change. Both sit under the `agent-sandbox` gate, so unrelated whitespace edits there cost a human's attention for nothing. Please drop them.

### Non-blocking observations

- The re-ping-on-cause-change is right, but the transitions are asymmetric: `dispatch → phantom-slot` needs 20 minutes to re-accumulate, while `phantom-slot → dispatch` is instant once the census corroborates. A census that flickers under a standing dispatch wedge can therefore ping twice per 20-minute window. Bounded, so I would not change it — worth knowing.
- `gatherQueuedDispatchable()` now runs every 30s sweep instead of only behind a free slot, and it `.all()`s the whole `runs` table and filters in JS. Trivial today; if `runs` grows it is the obvious place to push the status filter into SQL.
- Your reading of the spec on `live` vs `live + stopped`, and the decision not to re-derive `slots.used` from the counter, are both documented in the PR body and I agree with both. Recording that here so the human merger does not have to re-litigate them.

Nothing here touches the correctness of the signal — fix 1, 2 and 3 and I expect to approve on the next pass.

lennons301:
All three requested changes are in, plus two defects a fresh self-review pass found on top of them. Head is `ba8ad73`.

**Your three:**

1. **`runBoundedProbe` now wraps `raceWithTimeout`.** `src/lib/bounded-probe.ts` is deleted; the wrapper lives in `src/lib/timeout.ts` beside the race, so one race remains. You were right that the copy had already drifted — it never called `timer.unref?.()`, and building on the primitive restores that for free. It has its own tests now (both non-answers, a synchronous throw, and a probe whose *answer* is a refusal — `ok` is the outcome of asking, not the content of the reply). The "three callers" docstring is refreshed and names the census.
2. **CLAUDE.md updated**, extending the "Bounded outbound I/O" and "Queue bookkeeping" bullets in place per agent-context.md §3 — the fourth observation, the corroboration rules, `OCCUPANCY_DIVERGED_MINUTES`, and the corrected `raceWithTimeout` sentence. (This repo has no AGENTS.md; that pre-existing divergence I left alone.)
3. **Both EOF blank lines dropped.** `container-manager.test.ts` is byte-identical to `origin/main` again, so it's out of the diff and out of the gate table — and the `tsc` baseline errors at :396 are now provably outside this branch rather than merely asserted to be.

**What the self-review then found.** Worth flagging because one of them is more serious than anything in the first pass, and it sits right next to your non-blocking note about the asymmetric cause transitions — you were closer to a real bug than the "bounded, wouldn't change it" framing suggested.

- **A census that couldn't be taken erased the divergence clock.** `occupancyDivergedSinceMs` cleared on any null census, conflating "the daemon agrees" with "the daemon didn't answer". With a 5s bound on a 30s sweep, a 20-minute threshold needed *forty consecutive answers* — so a daemon degraded enough to time out every few minutes (the #125/#128 shape, and the likeliest companion of a real phantom) would have reset the clock on every hiccup and the card would never have fired. Now only positive agreement stops the clock, which is what AC3 actually asks for. Firing still requires a *current* positive divergence, so the fail-safe you validated is unchanged: a daemon that never answers still never raises this card.
- **The ping could downgrade its own advice.** `PickupWedgeCause`'s docstring already promised a re-ping on an *upgrade*; the code re-pinged on any cause change. The concrete path is your flicker scenario: census unavailable for one sweep with a dispatch wedge standing → card demotes to `dispatch` and pings again saying "check the orchestrator" about a box that still needs restarting. The memory now keeps the highest cause an occurrence reached and the ping fires on rank increase only.
- **One Docker-probe bound instead of two matching `5000`s.** The census claimed "same bound as `ADMISSION_PROBE_TIMEOUT_MS`" while being a second literal nothing kept equal. `capacity.ts` already imported from `agent-containers.ts`, so this needed no new edge and touched no other file.
- **Two CLAUDE.md claims I'd just written were themselves false** — same class as your finding 2, so I'd rather name it than fix it quietly. I described the comparison as `slots.used` vs `occupiedSlots()` when it's the counter vs the daemon census; and the `unref` note described a draft not in the tree, which is the changelog-in-a-context-file §3 forbids.

Three new evaluator tests cover the hiccupping daemon, unknown-holds-but-agreement-clears, and the no-downgrade-ping case. `pnpm test` 984 pass / 50 files; `pnpm build` compiles; eslint clean on all changed files.

Two things I declined, in case you want to argue them: the `remedy`-from-`cause` derivation (kept as you accepted last pass — though the fair sub-point that `fleet-view.test.ts` hardcodes remedy prose is real, just not worth the churn now), and re-deriving the dashboard's `slots.used` from the counter. Full reasoning in the updated PR body.

lennons301:

| gate | glob | path |
|---|---|---|
| agent-sandbox | `src/lib/docker/**` | `src/lib/docker/__tests__/agent-containers.test.ts` |
| agent-sandbox | `src/lib/docker/**` | `src/lib/docker/agent-containers.ts` |

lennons301-reviewer (approved):
Approving. This PR is **gated** (`human-signoff` label, auto-merge off), so this approval satisfies branch protection only — a human still looks and merges.

## Objective signals (no CI on PRs in this repo, so run locally on `agent/issue-152`)

- `pnpm test` — **984 pass / 50 files**
- `npx eslint` on all 15 changed `.ts` files — clean
- `pnpm build` — compiles
- `npx tsc --noEmit` — 2 errors, both in `src/lib/docker/__tests__/container-manager.test.ts:396`. I checked with `git diff origin/main...HEAD -- <that file>`: it is byte-identical to `main`, so those are the known baseline and provably not this diff.

## The three previously requested changes

1. **One race in the tree.** `src/lib/bounded-probe.ts` is gone; `runBoundedProbe` now lives in `src/lib/timeout.ts` and is a genuine wrapper — `raceWithTimeout(probe(), timeoutMs)` inside a try/catch, nothing re-implemented, so the `unref` drift is structurally impossible now. `checkMemoryAdmission` reads the same wrapper and its fail-open behaviour is unchanged (both non-answers still return `{ ok: true }` after their own log line). The new unit tests are the right ones: the two non-answers, a synchronous throw, and the case that actually pins the contract — a probe answering `{ ok: false }` is still an *answer* and lands in `value`. The "swallows a timed-out probe's later rejection" test earns its place too.
2. **CLAUDE.md updated in place**, extending the existing "Bounded outbound I/O" and "Queue bookkeeping" bullets rather than appending — `agent-context.md` §3 as it applies to this repo's (pre-existing, not-your-ticket) CLAUDE.md-instead-of-AGENTS.md divergence. I re-read both bullets against the code: the `observeAgentContainers` / prefix-ownership / null-is-unknown / one-directional / 20-minute claims all match what the tree does, and the earlier false claims are gone.
3. **Stray EOF blank lines dropped.** `container-manager.ts` now has exactly two lines changed (an import and the name template), and the test file has left the diff entirely.

## Against the ticket

All eight acceptance criteria met; I checked the invariants rather than the tests' word for them.

- **AC1/AC3** — `occupied > live` sustained past `occupancyDivergedMs` raises the pickup-wedged card plus one ping; only *positive agreement* clears the clock. The self-review's fix here is the important one and it is right: with `null` erasing the timer, a 20-min threshold needed 40 consecutive answered 30s sweeps, which is exactly what a degraded daemon (the likeliest companion of a real phantom) never gives. Firing still demands a current positive divergence, so an unanswerable daemon still cannot manufacture this card.
- **AC2** — `PHANTOM_SLOT_REMEDY` names the two numbers and says restart; it is produced in the evaluator and consumed identically by the card (`fleet-view.ts:613`), the ping (`notifications.ts`) and the server log (`sweep.ts:284`), so the three cannot advise differently.
- **AC4** — `gatherQueuedDispatchable()` is now unconditional and `divergedOccupancyDetail` names the starved task, so work starved behind a phantom is surfaced at `occupied == total`. `wedgedNow` still requires `slotFree`, so the dispatch signal's behaviour is unchanged.
- **AC5** — I verified the balance of the comparison at the source, not just in the tests: `occupiedSlots()` (`queue.ts:84`) counts non-parked `activeTasks` entries plus unregistered reservations; `isParked` is `idle && kind !== "interactive"`; `parkContainer` `docker stop`s exactly those. So a parked pass is absent from *both* sides and an interactive idle session is present on both — genuinely balanced. Provisioning and probe-failure shapes likewise.
- **AC6** — `DEFAULT_OCCUPANCY_DIVERGED_MS` sits beside the other three in `budgets.ts`, threads through `config.ts`, and is in the runbook env table.
- **AC7/AC8** — covered above.

**The stated spec divergence** (corroborating against `live` only, not `live + stopped`) is correct and I would have called the ticket's wording the bug. Since `occupiedSlots()` already excludes parked entries, adding `stopped` to the reality side could only raise the ceiling and hide a phantom sitting behind a parked container — the exact #151 shape on a 1-slot box computes `1 > 1` and stays silent. The `stopped` count is still carried onto the card, which is the right compromise.

**The 20-minute threshold** rests on `ensureImage()` running inside `createWorkspaceContainer`, which I confirmed — a cold agent-image build is legitimately uncorroborated the whole time, and this card's advice is a restart that would kill that task. Erring long is the right asymmetry, and it still turns #151's ~1.5h into a card with an hour to spare.

**Gated-path check for the human merging.** `src/lib/docker/**` gains a read-only census module and one name constant; `container-manager.ts` changes only that literal into the constant (I grepped: no other `interlude-task-` literals remain). `src/lib/config.ts` adds one non-secret env read. `src/lib/orchestrator/autonomy/**` gains the threshold default and the census wiring. Nothing changes container creation, mounts, network attachment, resource limits, credential handling, or what may be claimed / gated / armed / merged.

Smell baseline: nothing to raise. `runBoundedProbe` removes the duplication rather than adding any; `ADMISSION_PROBE_TIMEOUT_MS = DOCKER_PROBE_TIMEOUT_MS` is a named alias with a stated reason, not a middle man; `cause` + `remedy` is one map in one place rather than three in the renderers.

## Non-blocking observations

1. `pickupWedgedAnnounced` is now held alive by `divergenceStanding`, which is true whenever `occupancyDivergedSinceMs` is set — including while it is being *held* by an unanswerable census. So on a box whose daemon has stopped answering after one observed divergence, a dispatch wedge that fires, clears, and later recurs is a new occurrence that will not re-ping (rank 1 is not > rank 1). Narrow, needs a persistently unanswerable daemon (which the queue-heartbeat signal should be catching anyway), and it errs quiet rather than noisy — noting it only so the interaction is on record.
2. Wording: the comment above `pickupWedgedAnnounced` says an unresolved divergence "keeps the card standing even on a sweep that could not observe it". What is kept is the *announce memory* — `pickupWedged` itself is null on such a sweep unless the dispatch wedge is separately firing, so the needs-you card can blink off for one refresh and come back. The behaviour is the deliberate "firing demands a current positive observation" rule and I would not change it; only that clause overstates what the code holds.
3. Carrying forward the earlier note: corroboration is count-vs-count, so a leaked *running* agent container could inflate `live` and mask a genuine phantom. Container names embed the task id and `snapshot.slots.occupants` is already built from the active entries, so a set comparison would be strictly stronger. The ticket asked for counts — a future ticket, not this one.
4. `gatherQueuedDispatchable()` now `.all()`s the whole `runs` table every 30s sweep and filters in JS. Trivial at current size; the obvious place to push the status filter into SQL when it isn't.
5. Suggestion (mine, not tied to a standard): the (b)+(d) block in `evaluateFleetHealth` is now the longest thing in the function by some way. If it grows again, lifting it to a `evaluatePickupWedge(input, prev, thresholds)` beside the two detail helpers would keep the top-level reducer reading as the four labelled signals it advertises.

PR #162, opened 2026-08-21
Interactive session closed in the UI leaks its slot — all task pickup wedges until the app is restarted

### LINKED ISSUE #159: Interactive session closed in the UI leaks its slot — all task pickup wedges until the app is restarted
## Symptom

After closing an interactive session from the UI ("Complete"), the slot is never released. The next
task sits in `queued` forever and **nothing dispatches again — interactive or autonomous — until the
app is restarted**. On prod (`CAPACITY_SLOTS=1`) that means one normal UI close wedges all pickup.

Reproduced four times on 2026-08-19 in prod, across two projects and both task kinds:

1. `01M0CVY2551TSV3SVC3GJW4DDY` (LPS, "start the dev server") — closed in the UI, slot stayed held
2. `01M0CXAS8TEJ9ZNFN83VFQ48CP` (LPS, "hi") — trivial session, closed in the UI, slot stayed held;
   next queued session never dispatched for ~20 min until `docker compose restart app`
3. `01M0CYWVYXFN3GBY9QXVBBY4RJ` (lemons, "hi") — closed in the UI at 12:16:50. Two **autonomous
   triage** tasks were queued behind it and sat undispatched for 100 s (verified at 12:18:32) with
   the queue polling every 2 s. `docker compose restart app` → the first triage dispatched within
   seconds.

So it is not specific to dev-server sessions, to LPS, or to interactive pickup: **any** interactive
session closed from the UI strands the slot, and the victims include autonomous passes.

Recovery is `docker compose restart app`. Queued tasks survive in the DB and dispatch on boot.

## The slot is held by a stale `activeTasks` entry

`occupiedSlots()` (`src/lib/orchestrator/queue.ts:29`) derives occupancy purely from two in-process
collections — turn-manager's private `activeTasks` map, plus `processingTasks`. Nothing reconciles
either against the Docker daemon.

The leak is localised to **`activeTasks`**, not `processingTasks`. Proof: the fleet dashboard's
`needsYou` "pickup-wedged" card names the occupant, and that label is built by iterating
`getActiveTasks()` (`src/lib/orchestrator/autonomy/sweep.ts:240-244`). During the wedge it read:

```
"cause":"pickup-wedged","severity":"red",
"body":"occupancy says 1 slot busy but the daemon reports 0 agent containers live —
        \"interactive: hi\" has not dispatched for 20m. The slot count is held in
        orchestrator memory with nothing behind it — restart the app to clear it."
```

At that same moment: the task row was `completed`, its `container_id` was `null`, and `docker ps`
showed **zero** agent containers. So the entry outlived the task, the container and the DB state.

## What makes it puzzling (the bit to diagnose)

`completeTask` (`turn-manager.ts:760`) *does* clear the map — `activeTasks.delete(taskId)` in its
`finally` (~line 879). And that `finally` demonstrably ran, because the very same block nulls
`containerId` and force-removes the container, **both of which were observed**.

So the entry is either re-added after the delete, or there is an ordering/duplicate-entry race.
Candidate paths worth checking first:

- `startTask` sets the entry at `turn-manager.ts:264` (`activeTasks.set(...)`). Is there a path where
  a turn still in flight re-sets it after `completeTask`'s `finally`? The complete route
  (`src/app/api/tasks/[id]/complete/route.ts`) is **fire-and-forget** (`completeTask(id).catch(...)`),
  so it runs concurrently with whatever the turn is doing.
- `processQueuedMessages` mutates `activeTasks.get(taskId)!.state` (~line 642) under a non-null
  assertion — that throws rather than re-adds, but the surrounding ordering is suspect.
- Whether two `completeTask` invocations can interleave (the route only guards on
  `task.status !== "running"`, read before the async work starts).

Ruled out during investigation, to save the next person the trip:

- **Not** a hung `processingTasks` promise. The occupant label proves `activeTasks` is the holder.
- **Not** a `runTurn` stream hang. `waitForExecStream` (`turn-manager.ts:1438`) has a 2s
  `exec.inspect()` poll that resolves when the exec stops *or* the container disappears.
- **Not** `scanPorts` (`port-scanner.ts`) — it has a 5s `setTimeout` guard and a try/catch.
- **Not** a Discord hang in `postIdleNotification`. LPS has `discord_channel_id: null`, so it returns
  at the early guard.
- **Not** an out-of-band `docker kill`. Sean closed these sessions normally through the UI.
- Red herring: `[autonomy] All 1 slot(s) busy: ` logs with an **empty** occupants list during normal
  provisioning. That is correct — between `processingTasks.add()` and `startTask`'s
  `activeTasks.set()` (`turn-manager.ts:264`) the task is counted by the second loop and has no
  occupant label yet. It is not evidence about the leak.

## Why it is unrecoverable without a restart

- `activeTasks` and `processingTasks` are **module-private**. `completeTask`, `cancelTask`, the
  container reaper and boot restart-recovery have no reach to reconcile them against the daemon.
- Completing or cancelling the ghost cannot help: `occupiedSlots()` counts `processingTasks` entries
  *not* in `activeTasks`, so removing an `activeTasks` entry can just shift the count to the other
  loop.
- The one daemon-truth check the queue owns is **short-circuited** (`queue.ts:79`):
  ```ts
  const admission = slotFree ? await checkMemoryAdmission() : { ok: false };
  ```
  `checkMemoryAdmission` is the thing that asks Docker what is actually running, and its doc comment
  claims it catches "any drift the slot bookkeeping missed" — but a phantom slot fails `slotFree`
  first, so it never runs. It only ever catches **under**-counting drift, never over-counting.

## Also: the dashboard disagreed with the gate

During the wedge, the fleet payload reported `"slots":{"total":1,"used":0,...,"occupant":"free"}`
while `occupiedSlots()` (the value that actually gates pickup, surfaced as `slots.occupied` in
`sweep.ts:409`) was `1`. So the dashboard read **free** while pickup was blocked, and only the
`needsYou` card told the truth. These two numbers should not be able to disagree.

Worse, the `pickup-wedged` card only fires after a dwell threshold (its text read "has not dispatched
for 20m"). For the first ~20 minutes the wedge is **completely silent**: the dashboard shows a free
slot, no card, and tasks simply never start. In reproduction 3 above the box was wedged and unflagged
for the whole 100 s window observed.

## Suggested direction

Three layers, cheapest first. Layers 1 and 2 are small and independently shippable — together they
turn a hard wedge into a self-correcting hiccup **even before the race in layer 3 is understood**.

### 1. A terminal task can never hold a slot (primary fix)

A true invariant, not a heuristic, and it needs no Docker call — one SQLite read over at most a
couple of ids per poll. This alone clears all four observed reproductions.

In `occupiedSlots()` (`queue.ts:29`), skip and prune any `activeTasks` / `processingTasks` entry whose
task row is `completed` / `failed` / `cancelled`:

```ts
// queue.ts — occupiedSlots()
// A task that has reached a terminal status cannot legitimately hold a slot.
// Nothing in-process should outlive its task row, but when it does (#159)
// pickup must self-heal rather than wedge until the app is restarted.
```

`activeTasks` is private to turn-manager, so this needs a small exported reconcile/prune helper there
(e.g. `pruneTerminalActiveTasks(): string[]`); `queue.ts` can prune `processingTasks` itself.

### 2. Reconcile against the daemon when pickup is blocked

Fix the short-circuit at `queue.ts:79` so the daemon probe can veto a **stale-busy** verdict, not only
a free one:

```ts
const slotFree = await capacityProvider.isSlotAvailable();
if (!slotFree) await reconcileSlots();   // drop entries with no live container
```

`checkMemoryAdmission` already runs exactly the query needed
(`listContainers({ filters: { name: ["interlude-task-"] } })`), and the `pickup-wedged` card already
computes precisely this comparison — it just *reports* it instead of acting on it.

Two care points:

- **Parked containers are `docker stop`ped**, so they are legitimately absent from the running set
  (`isParked` already excludes them from the count). Pruning must not evict them — they are needed to
  deliver a fix-up turn into the same attempt.
- **A task mid-provisioning has no container yet** (see the empty-occupants red herring above), so
  gate pruning on the task row being terminal *or* on a grace period. That means `processingTasks`
  should become a `Map<string, number>` carrying insertion time rather than a bare `Set`.

### 3. Fix the underlying re-add race

`completeTask`'s `finally` demonstrably runs, so either something re-adds the entry or two paths
interleave. Start with the fire-and-forget complete route running concurrently with the in-flight
turn (see "the bit to diagnose" above). Diagnose before writing a fix here — the candidates listed
earlier are guesses, not conclusions.

### Plus two observability fixes

These made the bug far harder to see than it should have been:

4. **Unify `slots.used` with `occupiedSlots()`** so the dashboard cannot read *free* while pickup is
   blocked. Right now the headline number and the gating number are computed differently and
   disagreed throughout both wedges.
5. **Cut the `pickup-wedged` dwell threshold** well below ~20 minutes, or drop the card entirely once
   layer 1 makes the condition self-healing. Until then, `slots.used` should not be trusted — the
   `needsYou` array is the only truthful signal, and a queued task that never starts is the real tell.

## Repro

1. Start an interactive session on any project (a trivial "hi" is enough).
2. When the agent goes idle, click Complete in the UI.
3. Start another interactive session → it stays `queued` indefinitely.
4. `curl -sN https://<domain>/api/fleet/stream | head -c 700` → `needsYou` shows `pickup-wedged`,
   while `docker ps` shows no agent containers.
5. `docker compose restart app` → the queued task dispatches immediately.

## Notes

- Distinct from #36, which is about surfacing **cgroup** OOM kills *inside* agent containers.
- Found while investigating a separate prod incident the same day (host OOM wedge: `dockerd` leaked
  to 1.19 GB — more than the whole 1024 MiB orchestrator reserve — plus `AGENT_MEMORY_MB=2048` on a
  3.7 GiB box and zero swap, so a Next.js dev server took the host down via `global_oom` while its
  own cgroup cap was never reached). That memory-overcommit problem is **not** covered here and is
  still unfiled.

lennons301:
Closes #159

## The phantom slot was two maps, not a race

The ticket's leading theory was a re-add race in `activeTasks`, and its stated proof was the `pickup-wedged` card naming `"interactive: hi"` as the occupant. Both turned out to be wrong, and knowing that is the fix:

- **The card names the *queued* task, not the occupant.** `starvedWorkPhrase` (`src/lib/fleet/health.ts:428`) reads `input.queuedDispatchable`, gathered by `gatherQueuedDispatchable` from tasks in `queued`. So `"interactive: hi" has not dispatched for 20m` was the *starved* session, and the evidence never identified the holder.
- **There is no re-add.** `activeTasks.set` has exactly one call site (`turn-manager.ts`, right after `createWorkspaceContainer` returns). No path re-adds an entry after `completeTask`'s `finally` deletes it.

What actually happened: **`turn-manager` is evaluated twice in one process, so there were two `activeTasks` maps.** Next compiles `instrumentation.ts` — where the queue poll loop and the autonomy sweep run — and the app-router route handlers into separate module graphs, each with its own registry. `POST /api/tasks/[id]/complete` mutated the app-router's copy; the queue counted the orchestrator's.

Measured, not inferred. I put a module-instance marker in `turn-manager`, built the production entrypoint (`output: standalone`, Next 16.1.6), booted it and asked the app-router how many distinct maps existed:

```
[DEBUG] turn-manager loaded, instance 01M0JAAJMR… (load #1, distinct activeTasks maps: 1)
[DEBUG] turn-manager loaded, instance 01M0JAAK65… (load #2, distinct activeTasks maps: 2)
$ curl -s /api/debug/… → {"route":{"own":"01M0JAAK65…","moduleLoads":2,"distinctActiveTasksMaps":2}}
```

The route's own instance is load #2; the queue holds load #1. After the fix the same probe reads `distinctActiveTasksMaps: 1`. (The probe was temporary and is not in this branch.)

This explains every observation the ticket found puzzling:

| Observation | Explanation |
|---|---|
| `finally` demonstrably ran (`containerId` nulled, container force-removed) | It did — all of it. Only `activeTasks.delete` was a no-op, landing in the other map |
| The entry outlived the task, the container *and* the DB state | Nothing in the orchestrator's map was ever touched |
| `docker ps` showed zero agent containers | The route reconnected via `task.containerId` (its map being empty) and removed the container correctly |
| Reproduced on **every** UI close, not intermittently | A split registry is deterministic; a race would not be |
| ✅-from-Discord always worked | `discord/client.ts` reaches `completeTask` by dynamic import *inside* the orchestrator's graph |
| Only an app restart cleared it | One process, two registries — nothing at runtime reconciles them |

## What's in here

**Layer 3 — the root cause.** `activeTasks` moves to `globalThis` via a new `processSingleton` (`src/lib/process-singleton.ts`), the only registry the two graphs share. The regression test loads `turn-manager` twice through `vi.resetModules()` — the only seam that can see a defect which is not in what either copy does, but that there were two.

**Layer 1 — a terminal task can never hold a slot.** `occupiedSlots` already refused to count a *reservation* whose task had finished (#151); it still counted an `activeTasks` entry, which is the half that held the box. Now the task row decides on both sides, and `releaseSpentReservations` drops what the count skipped. No Docker call: a `completed`/`failed`/`cancelled` task runs no agent process. `taskIsFinished` moves to `src/lib/tasks/stored-status.ts` so the queue and the turn manager share one predicate. The prune leaves containers alone — the stale-container reaper already owns any container whose task is neither live nor run-owned. A test pins that a parked pass is untouched.

**Layer 2 — the daemon can now veto a stale-busy verdict.** The queue owned a daemon-truth check and never ran it where it mattered: `checkMemoryAdmission` sat behind `slotFree`, so a phantom slot failed the slot test first and the probe never fired — it could only catch *under*-counting, which is the wrong direction for a wedge. A blocked pickup is now re-checked every ~32s, and a session whose container the daemon **positively** 404s is dropped *and recorded `failed`*, because releasing the slot alone would leave the task `running` with no container: counted by `slots.used`, uncounted by the queue — the very disagreement this ticket is about. Existence, not liveness (`observeContainerAbsent`): an entry in `setup` is created-but-not-started and a parked pass is `docker stop`ped, and `isContainerRunning` calls both gone. Unknown decides nothing. Run-owned passes are left to #95/#97/#106, which account for attempts properly.

**Teardown paths the fix made live.** Every `entry`-conditional branch in `completeTask` and `cancelTask` was unreachable from a route handler before, because the route's map was always empty. Sharing the map switched them all on, and three were wrong:

- the four `activeTasks.get(taskId)!.state = …` writes (which the ticket flagged as suspect) become a tolerant `setTaskState` — a missing entry turned an ordinary race into a TypeError inside `startTask`'s try, i.e. a `failed` task plus a Discord embed and an issue comment for a session just closed successfully;
- `startTask`'s catch returns early when its task is already terminal, so nothing can overwrite an outcome the owner asked for;
- `completeTask` verifies an in-memory container the same way it verifies one rebuilt from `containerId` — a held handle is not proof, and without the check completing a session whose container died out of band reports *failed*;
- `cancelTask` records `cancelled` *before* killing the container, or the dying turn writes `failed` over the cancellation and burns an attempt;
- `failImplementAttempt` takes its container handle before the status goes terminal — the awaited issue comment after it gave the new prune a window to drop the entry, leaking a ~2 GiB container until the 5-minute reaper.

## Where I did not do what the ticket asked, and why

**Item 4 — "unify `slots.used` with `occupiedSlots()`" — declined as worded, addressed at the other end.** CLAUDE.md records a deliberate #152 decision that `slots.used` stays DB-derived and is *not* re-derived from the counter, because those two surfaces disagreeing is what the #151 incident looked like from outside — overwriting either hides the evidence. I think that argument still holds, and the gap closes better from the other side: layer 1 makes the *gating* number read the same task rows the dashboard reads, so both now exclude terminal tasks and parked passes (each pinned by its own test — `fleet-view.test.ts:188` and the new slot tests), and layer 2 finalizes what it releases rather than leaving a row only one side counts. The specific divergence observed during the wedge is now impossible, and a residual one is real evidence again rather than an artefact. Happy to be overruled — say so and I'll make `slots.used` read the counter.

**Item 5 — "cut the `pickup-wedged` dwell threshold" — declined, and I think the ticket's own alternative is the right one.** The sharp observation here is correct and worth recording: on a saturated box the 3-minute `dispatch` card **cannot fire at all**, because `wedgedNow` requires `slotFree` — which is exactly why the first ~20 minutes were silent, leaving only the 20-minute phantom-slot card. But that structure is right: firing the dispatch card while saturated would alarm on every legitimately-busy one-slot box with a queue, and the 20 minutes is sized by the worst *honest* case (a cold `ensureImage` build inside `createWorkspaceContainer`), with a remedy — restart — that would kill exactly that task. The ticket offers the better fix itself: "or drop the card entirely once layer 1 makes the condition self-healing". Layers 1–3 make it self-heal in ≤2s (terminal) or ≤32s (container gone), well inside the debounce, so what remains for the card is the phantom the daemon cannot corroborate — where "restart the app" is still the right advice. I lowered no threshold and dropped no card.

**The reservation-grace half of layer 2 — built, then removed.** The ticket asks for it explicitly, including the `Map<string, number>` change, and I implemented it before concluding it is not safe here. Releasing a reservation on age alone cannot distinguish the failure from the honest case: a cold image build genuinely has no container to show, and `checkMemoryAdmission` sees zero *running* agent containers while a build is in progress — so it admits the next pickup, and when the slow provision lands there are two agent containers on a one-slot box with `AGENT_MEMORY_MB=2048`, with nothing re-checking memory. That is the overcommit shape behind the 2026-08-19 host OOM. `OCCUPANCY_DIVERGED_MINUTES` only *reports* at that bound; acting on it is a different risk. A test now pins that a provisioning reservation is held however long it takes, and `slotReservations` is back to a `Set`. A phantom reservation stays the watchdog's card, whose remedy is a human-gated restart. If you'd rather have the self-heal and accept the overcommit risk, it's a small change to restore.

## A separate defect found on the way — needs filing

**Webhook-triggered sweeps run on the app-router graph.** `POST /api/webhooks/github` calls `runAutonomySweep()` directly, so it gets that graph's own, freshly-empty `fleetHealthState`, `sweeping`, `inFlightClaims`, `saturationAnnounced`, `dailyCapAnnouncedDay` and every announce-dedup set — while `recordFleetHealth` **is** `globalThis`-backed. So a webhook sweep evaluates the #126 signals against debounce clocks that started that instant, computes all-clear, and **overwrites a standing needs-you card** in the shared store, dedup memory included. It also reads `isQueueRunning() === false` and `getQueueLastProgress() === null` (that graph never started the loop), reads an empty `slotReservations` so `claimableSlots` overcounts, and — because `sweeping` is per-graph — can run concurrently with the interval sweep and claim the same ticket, which is precisely what its single-flight comment promises it cannot.

Not fixed here, deliberately: eleven fields need the same treatment, concurrent sweeps need tests of their own, and there is a better shape worth considering first — have the route record "something changed" and let the orchestrator's loop pick it up next tick, which removes the problem by construction at the cost of ≤30s latency the spec already calls acceptable ("the webhook is only latency on top of this — the sweep is the backbone"). I couldn't open the issue from this session; the detail is in `CLAUDE.md` under the module-graph convention.

## Validation

- `pnpm test` — **996 passing** (52 files), up from 992 on `main`: +2 cross-graph, +5 slot accounting, +3 session teardown.
- `npx tsc --noEmit` — clean, other than two pre-existing errors in `src/lib/docker/__tests__/container-manager.test.ts` (confirmed identical with my changes stashed).
- `pnpm build` — succeeds.
- `pnpm lint` — 16 errors, all pre-existing and none in files I touched (`custom-server.js`, `preview-pane.tsx`, `output-parser.test.ts`), per the baseline noted on the ticket. Scoped my check to the changed files.
- The layer-2 and teardown tests were confirmed **red-capable** by disabling the code under test and watching them fail, not just by watching them pass.

Not verified on real Docker: the reconciliation and the teardown paths are covered by tests with the daemon stubbed, and the split-graph fix was verified against a real production build as shown above, but no agent container was actually started in this session.


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
## Approving — gated, so a human still merges

Reviewed with fresh context against issue #159. This is unusually good diagnostic work: the ticket's leading theory (a re-add race in `activeTasks`) and its stated proof (the `pickup-wedged` card naming the occupant) were both wrong, and the PR says so with a measurement rather than a hypothesis. I re-checked both claims independently and they hold:

- `pickupWedgeDetail` → `starvedWorkPhrase` (`src/lib/fleet/health.ts:402-407`) reads `input.queuedDispatchable`, so `"interactive: hi"` in that card was the *starved* task. The ticket's evidence never identified the holder.
- `activeTasks.set` has exactly one call site (`src/lib/orchestrator/turn-manager.ts:372`), so "there is no re-add" is right.
- `wedgedNow` requires `slotFree` (`health.ts:243-246`), so the item-5 declination's core observation — the 3-minute dispatch card cannot fire on a saturated box — is a fact about the code, not a preference.

### What I verified

- `pnpm test` — **996 passed / 52 files**, matching the PR body.
- `npx tsc --noEmit` — only the two pre-existing errors in `src/lib/docker/__tests__/container-manager.test.ts`. Confirmed pre-existing: that file was last touched by `b9a9fb1` (#157), which is on `main` and not in this diff.
- `npx eslint` on all eight changed files — clean. (Repo has a 16-error baseline on `main`; none in files this PR touches, as the ticket instructed.)
- `pnpm build` — exit 0.
- No CI is configured on this branch (`gh pr checks` reports none), so the above are local runs.

### The three layers

**Layer 3 (root cause).** `processSingleton` is the right shape and the doc comment earns its length. The regression test is the best thing in the diff: `expect(routeGraph).not.toBe(orchestratorGraph)` makes the seam self-verifying, so the identity assertion can't silently go vacuous — which matters for a defect that is not in what either copy does but that there are two.

Two second-order consequences of sharing the map that I checked and think are net positives, worth naming since neither is in the PR body: `POST /api/tasks/[id]/run` calls `startTask` **from the app-router graph**, so before this change a task started that way registered its entry where the queue could not see it — an *under*-count, i.e. silent overcommit. And `reapOrphanedReviewPasses` / `terminalizeFinalizedRunTasks` call `getActiveTasks().delete(...)`; from a webhook-triggered sweep those deletes previously landed in the empty copy. Both are now correct. The audit surface for "which branches did sharing switch on" is small and closed — only three route handlers reach the orchestrator at all (`run`, `cancel`, `complete` via `task-runner`), and the PR hardened the two that mutate.

**Layer 1.** The invariant is genuinely an invariant, the predicate is shared rather than restated (`stored-status.ts`), and moving `queue-slots.test.ts` off a stand-in map onto the real one means the prune under test is the prune production runs. The parked-pass test is the one that matters and it's there.

**Layer 2.** One-directional, positive-404 only, existence-not-liveness, unknown decides nothing, run-owned passes left to #95/#97/#106 — each of those is pinned by its own test, and `abandonSessionWithoutContainer` finalizing the row rather than only freeing the slot is the right call: releasing alone would have manufactured exactly the `slots.used` / `occupiedSlots()` disagreement the ticket complained about.

**The declinations.** All three (item 4, item 5, the reservation-grace half) are argued from recorded decisions or from the 2026-08-19 OOM shape, not from taste, and each leaves a test pinning the behaviour it chose. I'd back all three — in particular the reservation-grace one, where releasing on age alone plus a `checkMemoryAdmission` that sees zero *running* containers during a cold `ensureImage` build is a real overcommit path on a 1-slot / 2048 MB box. Since the ticket asked for all three explicitly, the merging human is the right person to confirm; nothing in the code depends on the answer.

### Findings — none blocking

1. **`abandonSessionWithoutContainer`'s guard doesn't match its doc.** The guard is `!task.runId`; the comment reads "Only a session no run owns, which is to say an interactive one." Triage passes also carry no run (CLAUDE.md: "triage owns no run"). So a triage pass whose container the daemon 404s is marked `failed` here, and because `startTask`'s catch now early-returns on a terminal task, the fail-closed `triageResult: unparseable` that catch exists to store is never written. It degrades to a bounded retry (`MAX_TRIAGE_PASSES_PER_ISSUE` in the candidate gather), not a loop — but it silently bypasses the "nothing applied, the owner told once, `needs-triage` kept" path. Either gate on `kind === "interactive"` or record why triage is fine here.

2. **The early return makes `alreadyReaped` dead code.** `turn-manager.ts:565` is `isReviewPass && storedTaskStatus(taskId) === "failed"`, but `failed` is terminal, so `taskIsFinished(taskId)` at line 533 has already returned. The comment above it still describes it as load-bearing against #95. Harmless as code; in this repo the comments are the record, so a comment describing an unreachable path is the kind of thing that misleads the next reader.

3. **One reader of `activeTasks` still doesn't apply the predicate.** `sweep.ts:468` gathers occupant labels skipping only `isParked`, so `slots.occupied` can read 0 while `slots.occupants` names a stranded terminal session for up to one poll — the same self-disagreement this PR set out to end, in the sweep's own snapshot. It only feeds `owedReviewReason`'s card body and the saturation log line, so it's cosmetic, but it's a one-line fix and the PR's whole thesis is that the fleet shouldn't disagree with itself about whether a task finished.

4. **`reconcileSlotsAgainstDaemon` doesn't skip `state === "completing"`.** When `completeTask` takes the new `observeContainerAbsent` branch and drops the handle, it then awaits `markPrReady` and `commentOnIssue` before writing `status: completed`. In that window a reconcile tick can write `failed` plus a "Session ended: …" line into a session that completes successfully a moment later. Final status ends up `completed`, so this is a spurious feed entry in an already-degraded scenario — but a `state === "completing"` skip closes it exactly, and the ordering elsewhere in this PR is otherwise careful to the point of being the thing under test.

5. **A reconcile-abandoned session is silent on Discord.** Every other failure path (`startTask`'s catch, `completeTask`'s catch) calls `notifyTaskFailed`. For an interactive session dispatched from Discord — the Phase 4 primary surface — the owner's only surface gets nothing. Strictly better than the wedge it replaces, so not a blocker, just unfinished.

6. **Comment overclaims.** "Deliberately not the port scan's 15, so the two do not land on the same tick for the life of the process" — 15 and 16 are coprime, so they coincide at every 240th poll (~8 min). The intent (different periods, so they don't collide *every* time) is right; the sentence isn't.

7. **`process-singleton.ts` enumerates "three older stores"; there are five.** `fleet/failing-checks.ts` and `discord/notifications.ts` also hand-roll `globalThis`. The rule the module states ("a hand-rolled slot when absence itself carries meaning") covers both, so nothing is wrong — but an exhaustive-looking list that is wrong on day one is worse than no list.

8. **The unfiled follow-up should be filed before this merges.** CLAUDE.md's new module-graph bullet says both "**Still split, and known (needs filing)**" and "Filed, not fixed here" — the second half is stale after `403b289` removed the issue citation. More importantly the defect itself is real and, I agree, out of scope here: `runAutonomySweep` from `POST /api/webhooks/github` runs on the app-router graph with its own empty `fleetHealthState` / `sweeping` / `inFlightClaims`, while `recordFleetHealth` *is* globalThis-backed — so a webhook sweep computes all-clear against a debounce clock that started that instant and overwrites a standing needs-you card, dedup memory included. That is a watchdog that can be silenced by an inbound webhook, and it will be lost if its only record is a sentence in a CLAUDE.md paragraph. Please open the issue (the route-records-a-flag / loop-picks-it-up shape the PR body suggests looks right) rather than merging on the note.

None of the eight changes what the ticket asked for or breaks an invariant, so I'm approving. This PR is **gated** (`human-signoff`, auto-merge off): my approval only satisfies branch protection — a human looks and merges. Worth their attention on the merge: the three declined items, and finding 8.

lennons301:
Finding 8 — follow-up filed as #163 (webhook sweeps on the app-router graph: a standing fleet-health card can be silenced by an inbound webhook, and two sweeps can claim one ticket). Filed before merge as the review asked, with the route-records-a-flag shape recorded as the suggested direction. The `CLAUDE.md` bullet in this PR should cite #163, but that push would dismiss the approval — leaving it for the owner to decide at merge time.

Issue #164, opened 2026-08-26
Quota-aware execution lanes: rate-limit pause/resume, tier degrade ladder, and API overflow guardrails


lennons301:
## Problem Statement

The company-paid Claude Enterprise account has expired. Interlude's fleet now
runs on a personal Claude Pro subscription, and the platform was built on the
assumption of effectively unlimited quota on a single provider.

Three things break as a result:

1. **Rate limits are invisible and destructive.** The Claude Code CLI already
   emits a `rate_limit_event` on its stream, carrying utilization, the limit
   type that tripped, and the reset time. Interlude explicitly discards it. When
   a wall is hit, the pass fails like any other failure — burning one of three
   attempts on a ticket, and eventually routing a perfectly good ticket to
   `ready-for-human` because the quota ran out, not because the work was hard.
   The owner's own interactive Claude Code sessions draw on the same window, so
   the fleet can silently consume the quota the owner is about to need.
2. **Everything defaults to the most expensive tier.** Opus is the account
   default, so every implement, review and triage pass reaches for it unless an
   environment variable says otherwise. On a Pro plan the weekly Opus allowance
   is small; on API billing it is the most expensive option available. There is
   no way to say "run the fleet on Sonnet, keep Opus for the passes that need
   it" without editing environment variables in Doppler and redeploying.
3. **The substrate is hard-wired to one vendor on one billing posture.** The
   harness (Claude Code), the auth path (a subscription OAuth token), and the
   model identifiers are entangled across container-manager, config and the
   autonomy reducer. Swapping the subscription for a Codex or Kimi plan, or
   overflowing to a metered API when the subscription wall is hit, currently
   means a code change in several places rather than a configuration change.

The owner's stated outcome: **pause and continue when you can**, plus the
ability to choose cheaper models — and, because this is a good moment to do it,
enough composability to flip the subscription, flip the harness, and overflow to
an API of choice, with guardrails against a surprise bill.

## Solution

Interlude becomes **quota-aware** and its execution substrate becomes
**pluggable**, through one unifying concept: the **execution lane**.

An execution lane is the answer to "how does a pass actually get run?" — a named
bundle of `{ harness adapter, auth environment variable names, base URL,
tier → model-id map, caps }`. Lanes are declared in a checked-in configuration
file so their shape is version-controlled and reviewable; the secrets they name
stay in Doppler and are referenced by environment-variable name, never inlined.
One lane is the **primary** (today: Claude Code against the Pro subscription);
[...]

PR #177, opened 2026-09-01
UI-editable settings, with per-pass-kind model tier defaults

### LINKED ISSUE #166: UI-editable settings, with per-pass-kind model tier defaults
## Parent

#164

## What to build

A settings section in the Interlude UI where the owner sets the default model tier for
each pass kind — implement, review, triage, interactive — and the change takes effect on
the next sweep without restarting the app.

Model choice becomes a **tier** (`heavy`, `standard`, `light`) rather than a vendor model
identifier, so the choice survives a change of provider. The existing `opus`, `sonnet`
and `haiku` names keep working as aliases, so tickets carrying a `model:` directive do
not break.

This ticket also lands the general mechanism that later tickets add fields to: a stored
override layer merged over the environment defaults by a pure resolver. An unset override
falls through to the environment default, so a fresh deployment behaves exactly as it does
today. Each field in the UI shows the effective value and whether it came from a UI
override or the environment, with a control to clear the override.

Note that the existing config accessor caches into a module-level value on first read, so
a UI override cannot ride on it — the resolved settings must be read fresh each sweep.

Safety ceilings stay out of the UI: the maximum per-attempt budget, the estate daily cap
and the attempt count are not editable here, and a UI override may never widen them.

## Acceptance criteria

- [ ] Each pass kind has a UI-settable default tier, and passes run on the tier selected
- [ ] `heavy`/`standard`/`light` are the vocabulary; `opus`/`sonnet`/`haiku` resolve as aliases
- [ ] A per-ticket `model:` directive still overrides the configured default
- [ ] An unset override falls through to the environment default; a fresh deployment behaves as it does today
- [ ] A change takes effect on the next sweep with no restart
- [ ] Each field shows its effective value and its provenance (UI override vs environment default), and can be cleared
- [ ] An out-of-range or disallowed value is rejected with a clear message, not silently clamped
- [ ] A UI override cannot exceed a hard ceiling
- [ ] The settings resolver is covered by pure unit tests for fall-through, override, rejection, ceiling and provenance

## Blocked by

lennons301:
Closes #166

Model choice becomes a **tier** — `heavy` / `standard` / `light` — and each pass
kind gets a UI-settable default for it, on top of the general override
mechanism later tickets in #164 add fields to.

## What's here

**The vocabulary** (`src/lib/model-tiers.ts`). A tier is the durable choice, so
it survives a change of provider; `opus` / `sonnet` / `haiku` keep working as
aliases, so a ticket already carrying a `model:` directive doesn't break. One
named tier→id map, which #164's execution lanes later replace per lane. Three
surfaces speak it: the settings UI, the environment defaults, and the ticket
directive (whose parser now normalises to a tier, so everything downstream deals
in tiers only).

**The override layer** (`src/lib/settings-resolver.ts`), a pure merge of the
stored overrides over the environment defaults, with four rules each pinned by a
test:

- an **unset** field falls through to *its own* environment default — a fresh
  deployment behaves exactly as it does today, and a cheap triage default never
  quietly decides what the reviewer runs on;
- a **set** field wins and reports its **provenance**, naming the variable that
  actually supplied the fall-back;
- only keys in `SETTINGS_FIELDS` are settable at all, which is how "a UI
  override may never widen a hard ceiling" is enforced — the ceilings are
  refused *by name* rather than absent by oversight;
- a disallowed value is **rejected with a message, never clamped**.

Overrides live on the existing single settings row as one JSON column, so a
later field is a registry entry rather than a migration, and a retired key is
dropped on read instead of left as a dead column.

**Freshness.** As the ticket notes, `getConfig()` memoises on first read, so an
override can't ride on it: `getSettingsOverrides()` is re-read at the point of
use — per pass and per follow-up turn — which is what makes a change take effect
at the next sweep with no restart.

**The screen.** Settings → Models: one row per pass kind, each showing the tier
in force, whether it came from the UI or the environment, which variable it
falls back to, and an `environment` option that clears the override.

Precedence, ordered once in `resolveAgentModel`: a ticket's `model:` directive
(work-carrying kinds only) → the UI override → the environment.

## Checks

`npx vitest run` — 1051 passed. Lint clean on every file this branch touches
(`pnpm lint` has a known pre-existing baseline elsewhere). `pnpm build` passes;
note that on a *fresh* `local.db` the build races itself running migrations from
several workers — a pre-existing local-only flake, unrelated to this change
(pre-migrating the file, or a second run, builds clean).

## Self-review

Ran `/code-review` against `origin/main` with #166 as the spec. Two real defects
came out of it, both in the fall-through path, both fixed in 6749aee:

- a review/triage row named its own variable while showing a value `AGENT_MODEL`
  had supplied — pointing the operator at a variable they'd find empty;
- a tier named in the environment reached the CLI as the word (`AGENT_MODEL=heavy`
  → `--model heavy`), which the harness doesn't accept — and this branch's own
  docs recommend that spelling. The environment now goes through the same tier
  map an override does; an id naming no tier is still passed through verbatim.

Also acted on: a duplicated `formatChanged` lifted into the shared atoms, the
ceiling list trimmed to spellings real constants use, the work-carrying-kind
test named once, an unread `group` field deleted, and a stale leaf-module
comment corrected.

Two review notes I did **not** act on, deliberately:

- *"`tierModelId` and `resolveModelTier` are Middle Men."* They are one-line
  delegates, but each is the name of a concept the callers should depend on
  rather than the table or the by-field form underneath — the indirection is the
  point when #164 makes the map per-lane.
- *"the ceiling rule is untestable further until a numeric field exists."*
  Agreed, and I chose not to invent one. #166 names exactly four fields; adding
  a bounded numeric knob purely to exercise the clamp path would be scope the
  ticket didn't ask for, and the ceilings it *does* name are enforced by
  construction and tested as such.

## Behaviour worth flagging

A run that has already recorded its model keeps it for its follow-up turns, so a
tier changed mid-run applies from the next claim rather than the next turn. That
is the pre-existing #80 pinning behaviour, not new here — interactive tasks,
which own no run row, pick up a change on their very next turn.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/model-tier-settings-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/model-tier-settings.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0022_settings_overrides.sql` |
| persistence | `drizzle/**` | `drizzle/0022_settings_overrides.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0022_snapshot.json` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/ticket.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/ticket.ts` |

lennons301-reviewer (approved):
Approving. Verified against issue #166 and the repo's own conventions; this PR is GATED (`human-signoff`, auto-merge off), so a human still merges.

## What I verified

**Objective signals.** GitHub reports *no checks* on this branch, so I ran everything locally in the worktree:

- `npx vitest run` — 1050 pass, 1 fail: `queue-slots.test.ts > holds a provisioning pickup's reservation however long it takes`, a 5 s vitest default-timeout blow-out. Not caused by this diff: the file is untouched, the poll loop it drives gained no work, it passes in isolation (0.87 s branch / 2.7 s main), and a full run on `origin/main` in the same conditions flakes the same way in a sibling file (`cross-graph-state.test.ts`) — the very flake this PR fixes with `TWO_GRAPHS_MS`. Load-sensitive, pre-existing.
- `npx eslint` on all 20 touched files — clean (the `main` baseline is elsewhere).
- `npx tsc --noEmit` — 2 errors, identical on `origin/main` (`container-manager.test.ts`), none in touched files.

**Acceptance criteria**, each traced to code and a test:

- Per-pass-kind settable tier, and passes run on it — `SETTINGS_FIELDS` covers implement/review/triage/interactive, `MODEL_TIER_FIELD_BY_KIND` maps `repair` onto the implement field, and `settings-overrides.test.ts` closes the loop by asserting `resolveAgentModel(...getSettingsOverrides())` returns what the PATCH response advertised.
- `heavy`/`standard`/`light` with `opus`/`sonnet`/`haiku` as aliases — `normalizeModelTier`, case- and whitespace-insensitive.
- Ticket `model:` still outranks the configured default, and still cannot touch review/triage — `isWorkPassKind` is one named predicate, tested both ways.
- Fall-through is per-field, never cross-field (`never spreads one kind's override onto another`), and a raw env model id (`claude-opus-4-8`) still passes through verbatim — so the live Doppler config keeps working unchanged.
- No restart needed — `getSettingsOverrides()` is re-read at both `turn-manager` call sites rather than riding `getConfig()`'s memo, and `settings.test.ts` pins the row round-trip.
- Provenance and clearing — `describeSettings` is the one thing both API and screen read, and `model-tier-settings-render.test.tsx` asserts the reading rather than the classes.
- Rejected, never clamped; ceilings refused by name — `parseSettingsPatch` is all-or-nothing (`refuses the whole patch when one key of several is a ceiling`) and `FIXED_CEILINGS` is asserted disjoint from the settable allowlist.

**Things I went looking for and did not find a problem with:**

- `runs.model` stores a model id, and is fed back as `ticketModel` on a repair/follow-up pass. Old rows only ever hold `opus`/`sonnet`/`haiku`, which round-trip through the tier map to themselves, so no in-flight run changes behaviour on deploy. An env-pinned raw id in that slot now falls through to the configured default instead of pinning — a deliberate, tested edge (`ignores a recorded raw model id in the directive slot`), and the more correct answer.
- Migration hygiene: journal `when` for 0022 is past 0021's, per the CLAUDE.md gotcha; the column is nullable, so existing rows read as `{}`.
- `sanitizeOverrides` on the read path means a retired key or narrowed vocabulary can never reach `--model`.
- The route is unauthenticated, consistent with its sibling `PATCH /api/settings/autonomy` — not a regression introduced here.

## Non-blocking

1. **`<legend>` is not a direct child of its `<fieldset>`** (`src/components/model-tier-settings.tsx:149-153`) — it sits inside the flex `<div>`. Browsers compute a fieldset's accessible name only from a legend that *is* a direct child, so as written each of the four radio groups is announced without its pass-kind name. The repo's one other fieldset (`new-task-form.tsx:139-140`) gets this right with a direct-child `sr-only` legend. One-line fix (`aria-label={field.label}` with `role="radiogroup"`, or hoist the legend), but it is a suggestion, not a documented-standard breach — a human may reasonably merge and follow up.
2. **The neighbour flake.** This PR gives `cross-graph-state.test.ts` its own timeout for exactly the reason `queue-slots.test.ts > holds a provisioning pickup's reservation…` fell over twice on my box: a 40-minute fake-timer advance racing a 5 s real-time default. Worth the same treatment in a follow-up.
3. `describeSettings` calls `resolveModelTierField` for every registry entry, so the "general mechanism" is general in its allowlist and validation but not yet in its resolution — the first non-tier field #164 adds will have to open it up. I think that is the right call (no speculative generality) and flag it only so the next ticket isn't surprised.

Issue #172, opened 2026-08-26
Execution lanes: checked-in lane config, pure resolver, and harness adapter seam


lennons301:
## Parent

#164

## What to build

The execution substrate becomes **pluggable**: which harness runs a pass, against which
provider, on which model, becomes configuration rather than code.

An **execution lane** is a named bundle of `{ harness adapter, auth environment variable
names, base URL, tier → model-identifier map, caps, billing kind }`, where billing kind is
`subscription` or `metered`. Lanes are declared in a checked-in configuration file so
their shape is version-controlled and reviewable. **Secrets appear only as environment
variable names** — never inlined — and the resolver reads the named variables at execution
time. A lane whose named variables are absent is reported as unavailable rather than
failing at exec time.

Because several providers expose Anthropic-compatible endpoints, flipping the subscription
is a base-URL plus token plus model-identifier change under the *same* harness adapter. So
exactly **one adapter ships** — Claude Code — behind an interface reviewed against what an
OpenCode or Codex adapter would need. Building a second adapter is out of scope.

Lane resolution is a pure function of `(lane config, pass kind, resolved settings)`
producing a concrete auth value, base URL, model identifier and caps, testable with no
provider involved. Which lane is primary is settable in the settings UI.

Every existing security constraint carries forward unchanged: agent containers get no host
bind mount, auth is exec-scoped and never in the persistent container environment, the
reviewer token never enters a container, and the GitHub CLI token stays absent for every
autonomous pass kind. **No lane secret may be stored in or served from the database** — a
project API route has previously leaked a stored token in cleartext.

The run ledger records which lane a run used, so spend stays interpretable.

## Acceptance criteria

- [ ] Lanes are declared in a checked-in configuration file, with secrets referenced by environment variable name only
- [ ] A lane resolves its tier map to a concrete model identifier for a given pass kind
- [ ] A lane whose named environment variables are absent is reported as unavailable rather than failing at exec
- [ ] The harness command and exec environment are built from a resolved lane, including its base URL
- [ ] One Claude Code adapter ships and all existing passes run through it unchanged
- [ ] No long-lived credential appears in the persistent container environment
- [ ] Autonomous pass kinds still carry no GitHub CLI token
- [ ] No lane secret is stored in the database or returned by any API route
- [ ] The primary lane is settable in the settings UI
- [ ] The run ledger records the lane used
- [ ] Lane resolution is covered by pure unit tests

## Blocked by

- #166

lennons301:
[...]

Issue #171, opened 2026-08-26
Quota admission gate: stop claiming new tickets when the window is nearly spent


lennons301:
## Parent

#164

## What to build

The fleet **stops claiming new tickets when quota is nearly gone**, so it does not start
work it cannot finish.

Above a configured utilization threshold — or when the account-wide status is already
rejected — no new tickets are claimed. Work already in flight is allowed to finish, and
paused runs are still allowed to resume; only *new* pickup stops. This sits alongside the
existing pause reasons for slots and the daily cap, and never overrides the global kill
switch or a per-project autonomy toggle.

**A single threshold, with no headroom reserved for the owner.** The fleet and the owner's
own Claude Code sessions draw on one pool, and the fleet takes what it takes. This was
chosen explicitly over reserving a slice of the window for interactive use.

The threshold is settable in the settings UI. The gate is announced once per transition,
following the existing announce-once pattern used for slot saturation and the daily cap,
and Discord is notified **only when the whole fleet is stalled on quota** — not once per
paused run.

## Acceptance criteria

- [ ] Above the configured utilization threshold, no new tickets are claimed
- [ ] An account-wide rejected status also stops new pickup
- [ ] Work already in flight is unaffected by the gate
- [ ] Paused runs still resume while the gate is closed
- [ ] The threshold is settable in the settings UI and falls through to its default when unset
- [ ] The gate is announced once per transition, not once per sweep
- [ ] Discord is notified only when the entire fleet is stalled on quota
- [ ] The kill switch and per-project toggle still take precedence
- [ ] The gate is expressed in the autonomy reducer and covered by table tests

## Blocked by

- #166
- #167

lennons301:
🤖 Attempt 1/3 starting.

Issue #167, opened 2026-08-26
Capture rate-limit telemetry and show fleet quota state on the dashboard


lennons301:
## Parent

#164

## What to build

The rate-limit telemetry the Claude Code CLI already emits stops being thrown away.

Today the output parser explicitly ignores `rate_limit_event`. After this ticket the
event reaches the turn result, the most recent observation across the fleet is persisted
as a single quota-state record, and a **quota tile** appears on the fleet dashboard
showing current utilization, which limit type is closest to tripping, when it resets, and
how long ago it was observed.

The event's known shape, recovered from the CLI's own schema:

```
status:          "allowed" | "allowed_warning" | "rejected"
resetsAt:        unix seconds (optional)
rateLimitType:   "five_hour" | "seven_day" | "seven_day_opus"
               | "seven_day_sonnet" | "seven_day_overage_included"
               | "overage"   (optional)
utilization:     number (optional/nullable)
overageStatus:   "allowed" | "allowed_warning" | "rejected" (optional)
overageResetsAt: unix seconds (optional)
```

Nothing acts on this yet — no pausing, no gating. This ticket is purely making the fleet
able to *see* its own quota. If the spike found the event does not reach stdout, apply
the fallback it recorded and parse the signal from the terminal result event instead,
accepting a coarser signal without a structured reset time.

## Acceptance criteria

- [ ] A `rate_limit_event` on the stream reaches the turn result with its fields intact
- [ ] The latest observation is persisted as fleet quota state, with an observed-at time
- [ ] The fleet dashboard shows a quota tile: utilization, closest limit type, reset time, observed-at
- [ ] The tile renders sensibly when no observation has been made yet
- [ ] A stream containing no rate-limit event parses exactly as it does today
- [ ] An unrecognised enum member from a future CLI version does not throw
- [ ] Parser behaviour is covered by fixture-driven tests using the real captured stream

## Blocked by

- #165

lennons301:
🤖 Attempt 1/3 starting.

PR #178, opened 2026-09-01
Spike: make rate-limit behaviour observable without waiting for a real rate limit

### LINKED ISSUE #165: Spike: make rate-limit behaviour observable without waiting for a real rate limit
## Parent

#164

## What to build

Answer, **deterministically and without waiting for a real rate limit**, the questions the
pause design in #164 rests on.

The naive version of this ticket — "run until you hit a wall, then write up what
happened" — does not work. The agent doing the spike is itself an LLM drawing on the same
personal subscription: at the moment the wall arrives it loses the ability to report, and
until then it is competing for the quota it is trying to study. **The orchestrator is the
observer, not the agent.** Interlude is an ordinary Node process reading the container's
stdout; it is not rate-limited and does not need to be conscious at the moment of the wall.

Two things make this testable. First, `rate_limit_event` is derived from response
**headers** — the `anthropic-ratelimit-unified-*` family carries utilization, reset,
status, overage-in-use and overage-disabled-reason, and every field of the event's schema
maps onto one. A stub endpoint that sets those headers therefore reproduces any limit
state on demand, including a tier-scoped rejection, without spending real quota. The
harness already takes a base URL, and #172 needs base-URL redirection regardless, so the
stub is a prefactor of work we are doing anyway. Second, the transcript question needs no
rate limit at all.

### Three things to deliver

**1. Transcript round-trip (no rate limit needed).** Run a pass, interrupt it mid-way,
copy the session transcript out of the container, provision a fresh container, restore the
transcript, and resume. Report whether the resumed pass genuinely continues the same
conversation on the same branch. The lossless-pause design in #169 rests entirely on this;
if it fails, the fallback is resuming as a fresh pass on the same branch with prior context
declared lost.

**2. A rate-limit stub.** A local endpoint the harness can be pointed at that returns the
`anthropic-ratelimit-unified-*` headers with configurable values. Use it to answer:

- **Does a headless pass wait, exit, or something else, when the limit is rejected?** If it
  waits, agent containers may already be holding memory for hours unnoticed, and the pause
  design must *detect and tear down* rather than react to an exit.
- **Does `rate_limit_event` reach stdout under `--output-format stream-json`, or only the
  SDK bridge?** If it does not appear, report what the terminal `result` event carries
  instead.
- **Is there a negotiated slow mode?** The header family includes `slow-offer`,
  `slow-max-wait` and `slow-budget-utilization`, which suggests the server can offer a
  degraded lane with a stated maximum wait rather than a flat rejection. If that is what it
  is, it is a **third behaviour** the pause design does not currently account for — say so
  plainly, because #168 and #170 would both need to change.

Where the stub cannot faithfully reproduce subscription behaviour — the headers are
Anthropic's, and the subscription path may differ from the API path — say which findings
are stub-derived and therefore provisional.

**3. A passive recorder.** Durable logging of unrecognised stream events and pass exit
conditions, verbatim and timestamped. This is ordinary code with no LLM in the loop, so
when a real wall does arrive the evidence is captured whether or not anyone is watching —
and the stub's provisional findings can be confirmed against reality later. This is the
part that must survive the ticket: it ships.

## Acceptance criteria

- [ ] The transcript round-trip is performed and its result reported, with the fallback named if it fails
- [ ] A configurable rate-limit stub exists that reproduces the `anthropic-ratelimit-unified-*` headers
- [ ] The wait-vs-exit question is answered against the stub, with the observed behaviour described
- [ ] Whether `rate_limit_event` reaches stdout is answered, with the fallback signal identified if not
- [ ] The slow-mode headers are investigated and the finding states whether a third behaviour exists
- [ ] Each finding states whether it is stub-derived (provisional) or observed against real quota
- [ ] Passive recording of unrecognised stream events and exit conditions is shipped and durable
- [ ] A captured stream containing a `rate_limit_event` is committed as a parser fixture
- [ ] Findings are recorded as a comment on this issue, not as a separate document
- [ ] No agent is required to self-report at the moment it loses the ability to report

## Blocked by

lennons301:
Closes #165

A spike, delivered as the ticket framed it: **the orchestrator is the observer, not the agent.** No finding here depends on an LLM being able to report at the moment it loses the ability to report.

Every question the ticket asks is answered below. Full findings, with the stub-derived/real-quota split stated per finding, are recorded as a comment on #165 as the ticket requires — no separate document.

## Headline findings

| Question | Answer | Evidence |
|---|---|---|
| Does `rate_limit_event` reach stdout? | **Yes** — second event in the stream, before any assistant output | **Real quota** |
| Does a rejected headless pass wait or exit? | **Exits, in ~2s.** It does not wait, does not honour `retry-after` | Stub-derived |
| Is there a negotiated slow mode? | **Yes, it exists** (the CLI calls it *low-priority mode*) but it is **not a third behaviour for us** — server-gated off and entered by a human action | Static + stub-derived |
| Does a restored transcript resume the same conversation? | **Yes**, including from a turn killed mid-tool-call | **Real quota** |

Two findings the milestone should absorb before #167/#168 are built:

1. **A quota wall reports itself as a *successful* turn.** `subtype: "success"`, with the wall visible only in `is_error: true` / `terminal_reason: "api_error"` / `api_error_status: 429`. `subtype` is the only field the orchestrator reads today, so #164's premise that a wall "fails like any other failure" is wrong — it looks like a pass that finished having done nothing. This is why `TurnResult` now carries the terminal event whole.
2. **The unified-window machinery is subscription-only.** Pointed at the same stub with an API key, the CLI emits **no `rate_limit_event` at all**. A metered overflow lane (#172/#173) therefore yields zero quota telemetry.

A third, smaller: the real-quota event carried **no `utilization` field**, and two fields beyond the shape #167 documents (`isUsingOverage`, `overageInUse`). #171's utilization-threshold gate needs a defined answer for "no utilization observed" rather than treating absence as zero.

## What ships

- **`src/lib/orchestrator/stream-recorder.ts`** — the part that must survive the ticket. Durable, verbatim, timestamped JSONL beside the SQLite database (so it inherits `/data` with no migration and no schema commitment before #167 knows what it wants), holding every unrecognised stream event and every pass's exit condition. Bounded three ways (payload cap; a per-(task, event-type) **per-turn** cap that emits one `suppressed` marker rather than going quiet; rotation keeping the older generation, because the oldest evidence is usually the interesting evidence) and it swallows its own errors — it sits in the stream-parse path of every turn, and a log that can break the pass it observes is worse than no log.
- **`scripts/rate-limit-stub.mjs`** — reproduces any limit state on demand via the `anthropic-ratelimit-unified-*` headers, with live scenario switching and a request log. A prefactor of #172's base-URL redirection.
- **Two captured fixtures** + tests asserting what the streams contain (acting on it is #167's job).
- Wiring: "ignored" no longer means "lost".

## Notes for the reviewer

- **The stub's enum values are not guesses.** `slow-offer` is an experiment arm (`treatment` | `control`); a plausible `"true"` reproduces nothing. My first pass at the slow-mode question used `"true"` and produced a **false negative** — the corrected run is the one reported.
- **No agent self-reported anything.** Findings come from stdout captures, the stub's request log, and static analysis of the shipped binary.
- **The in-CLI simulator is not usable.** The binary carries its own rate-limit simulator with this exact scenario vocabulary, but the released build compiles its enablement predicate to a constant `false`. Hence a stub of our own — which is what the ticket asked for anyway.
- **Cost:** ~$0.10 of real quota, on `haiku`, for the transcript round-trip and the one real-quota capture.

## Self-review

Ran `/code-review` against `origin/main` with #165 as the spec before pushing. It found six defects and **I agreed with and fixed all six** (commit `853f415`) — the most serious being that `passExit` sat on the normal return path rather than in a `finally`, so the OOM / lost-stream / torn-down-mid-turn exits the record exists for were exactly the ones that skipped it. I then tightened one thing the review did not raise: the exit-code probe now has its own 1s bound rather than the shared 5s one, since it fills in a log field on every turn's completion path.

Two timing-sensitive tests (`cross-graph-state`, `queue-slots`) intermittently hit vitest's 5s timeout on this machine. I verified this is **pre-existing and load-induced, not a regression**: neither test executes any changed code (both stub `startTask`), and `origin/main` reproduces the same failures under the same load average (~12 on 8 cores) while this branch runs 1037/1037 green when the box is quiet.

## Validation

- `pnpm test` — **1037 passed** (54 files) on a quiet box
- `eslint` — clean on every changed file
- `tsc --noEmit` — no new errors (the two `container-manager.test.ts` errors are the known baseline)


lennons301-reviewer (approved):
Approving. Reviewed against #165 as the spec, with no memory of how this was written.

**Objective signals** (no CI is configured on this branch, so run locally in the worktree):
- `pnpm test` — 1037 passed / 54 files, exit 0, on a quiet box. No flakes seen in `cross-graph-state` or `queue-slots` on this run.
- `eslint` on every changed file (including `scripts/rate-limit-stub.mjs`) — clean.
- `tsc --noEmit` — only the two known baseline errors in `src/lib/docker/__tests__/container-manager.test.ts`, which this PR does not touch.
- Gate check: no changed path matches an estate default or `docs/agents/review-gates.yaml` glob (nothing under `src/lib/orchestrator/autonomy/**`, `src/lib/docker/**`, `src/lib/config.ts`, `src/db/**`, `drizzle/**`, components/pages/CSS, workflows). Correctly armed.

**Against the acceptance criteria** — all ten are met, and the findings are on the issue as a comment rather than a doc artefact, as required. The transcript round-trip is reported with its fallback explicitly named as unnecessary; the stub reproduces the `anthropic-ratelimit-unified-*` family; wait-vs-exit, stdout-reachability and slow mode are each answered; every finding is tagged real-quota or stub-derived; two fixtures are committed verbatim; and nothing in the method asks an agent to self-report at the wall.

**Code I checked rather than took on trust:**
- `passExit` is genuinely in a `finally`, so the exits the record exists for (stream error, OOM, container torn down mid-turn) reach the log. The `await` inside that `finally` cannot mask the original error: `runBoundedProbe` folds both a timeout and a rejection into its outcome and never rejects, and `passExit` swallows its own errors.
- The cap reset (`counts.delete`) sits outside `guard`, and the test at `stream-recorder.test.ts:314` pins the reason — a sink that threw must still release the next turn's budget.
- `shouldRecordEventType`'s "known" set matches what the parser actually handles: `assistant`, `user`, `result` and `error` (`output-parser.ts:243`) all have branches, and `system` is the deliberate drop. So the allowlist is not claiming more understanding than the parser has.
- `resolveObservationsPath` follows `DATABASE_URL` the same way `src/db/index.ts` reads it (a bare path, not a URL), so "beside the database" is true on the VPS and in local dev, and the `.gitignore` entries cover the local-dev case.
- Rotation is bounded and its ENOENT-on-first-write path is handled; both fixtures are free of credentials (session ids and `/tmp` paths only).
- Recognised events cost only a set lookup and never consume cap budget, so the hot-path addition is a handful of sync writes per turn at most.

**Three non-blocking notes**, none tied to a criterion — take or leave:

1. `getStreamRecorder`'s comment says "the only caller is the stream-parse path in the orchestrator's own module graph". That is not quite true: `POST /api/tasks/[id]/run` → `runTask` → `startTask` → `runTurn` would resolve the recorder on the app-router graph (the #159 trap). I convinced myself the consequences really are benign — `appendFileSync` holds no shared fd, and a concurrent rotation degrades to the already-handled ENOENT — so this is a comment-accuracy point, not a `processSingleton` request. Worth a word if that route ever gets a UI caller.
2. Moving `handler.flush()` into the `finally` also means a trailing partial line now lands as a raw system message on the *error* path, where it previously did not — and a throwing DB insert there would replace the stream error. Both look like improvements-on-balance rather than problems, but neither is stated.
3. The turn-manager wiring itself (the `finally`, the 1s exit-code bound) has no test, which is where the self-review found the most serious defect. `runTurn` is private and Docker-coupled with no existing harness, so this matches the repo's current coverage of that function rather than regressing it.

The shipped half is the right shape for what it is: fail-safe by construction, bounded three ways, no schema commitment made before #167 knows what it wants, and the provisional stub findings arranged to be confirmed or refuted by the log without anyone watching.

Issue #173, opened 2026-08-26
Interactive overflow to a metered lane, with confirm-then-cap guardrails


lennons301:
## Parent

#164

## What to build

When the subscription is walled, **interactive work overflows onto a paid lane** instead
of stopping — because a human is sitting there waiting — while autonomous work still
pauses, so unattended work never spends real money without the owner knowing.

The money machinery this relies on — the real-money daily cap, the confirm-once-then-cap
gate, the metered-vs-subscription spend split, and the rule that guards key off a resolved
lane's **billing kind** rather than off having overflowed — is built in **#174**. This
ticket is its first consumer, not its owner. What lands here is the overflow itself:

**The trigger.** A walled primary lane routes interactive work onto an available metered
lane rather than failing or queueing it. Autonomous passes do not take this path; they
pause (#168/#169).

**The at-the-keyboard confirmation.** #174 establishes the once-per-local-day gate at fleet
level. Interactive overflow surfaces it where the human already is — in the UI, at the
moment of the crossing, with enough context to decide — and tells the user plainly that
they are capped rather than silently continuing once the cap is reached.

**Overage as a paid lane.** An active Anthropic overage means the account is already
spending real money rather than quota. **Decision: treat it as already being on a paid
lane** — subject to the same guard and counted against the same real-money cap. Without
this, an account with overage billing enabled would never see a `rejected` at all: the wall
would silently become a bill, which is exactly the surprise this milestone exists to
prevent. This rule reads `overageStatus` off the rate-limit event, which is why it lives
here with the telemetry chain rather than in #174. *(Stated rather than confirmed in #164 —
check it before building.)*

Interactive tasks own no run row and are today exempt from the daily autonomous cap *by
construction*. #174 closes that hole for metered lanes generally; this ticket must verify it
holds on the overflow path specifically, since that is where an interactive task first
touches real money.

## Acceptance criteria

- [ ] Autonomous passes never overflow to a metered lane; they pause
- [ ] Interactive work overflows to a metered lane when the primary lane is walled
- [ ] Overflow is refused when no metered lane is available or permitted, and the user is told why
- [ ] The first metered use in a local day is confirmed by the human in the UI, using #174's fleet-level gate
- [ ] Further overflow that day is automatic until the cap is reached
- [ ] At the cap, interactive work is told it is capped rather than overflowing silently
- [ ] Metered interactive spend is counted against the real-money cap from #174
- [ ] Subscription-lane interactive work remains exempt from the real-money cap
[...]

Issue #168, opened 2026-08-26
Pause an autonomous run on an account-wide rate limit instead of failing it


lennons301:
## Parent

#164

## What to build

An autonomous run that hits an account-wide rate limit **pauses instead of failing**.

Today a quota wall looks like any other failure: it burns one of the ticket's three
attempts, and enough of them route a perfectly good ticket to `ready-for-human` because
the quota ran out rather than because the work was hard.

After this ticket, a run that is rejected on the account-wide five-hour window enters a
new `rate_limited` state carrying a `resumeAfter` timestamp taken from the event's reset
time. Its container is **torn down** — parked containers hold memory without holding
slots, which is what wedged the host on 2026-08-04 — and the run consumes **neither an
attempt nor an interruption**, so the attempt budget keeps measuring how hard the work
was and the interruption bound keeps measuring orchestrator restarts.

On the fleet dashboard the run shows as a paused card labelled in place with when it will
resume. It deliberately does **not** enter "needs you", which keeps that section meaning
"a human decision is required".

The run must survive an orchestrator restart: boot recovery leaves `rate_limited` runs
alone, the way it already leaves gated and blocked runs to their human, because they are
waiting on a clock rather than on a lost turn.

Resuming is the next ticket — after this one a paused run stays paused.

## Acceptance criteria

- [ ] A run rejected on the account-wide window enters `rate_limited` with a `resumeAfter` from the event's reset time
- [ ] The run's container is torn down on pause
- [ ] A quota pause leaves the attempt counter unchanged
- [ ] A quota pause leaves the interruption counter unchanged
- [ ] The run appears on the dashboard as paused, showing when it will resume
- [ ] A paused run does not appear in "needs you"
- [ ] Boot recovery leaves a `rate_limited` run in place rather than re-claiming or failing it
- [ ] The global kill switch and the per-project autonomy toggle still stop the run
- [ ] The pause decision is expressed in the autonomy reducer and covered by table tests

## Blocked by

- #165
- #167

lennons301:
🤖 Attempt 1/3 starting.

PR #179, opened 2026-09-01
Capture rate-limit telemetry and show fleet quota state on the dashboard

### LINKED ISSUE #167: Capture rate-limit telemetry and show fleet quota state on the dashboard

lennons301:
Closes #167

Rate-limit telemetry stops being thrown away. The `rate_limit_event` the Claude Code CLI already puts on its stream now reaches the turn result, the latest observation across the fleet is persisted as one quota-state row, and a quota tile on the dashboard says where the account stands.

No fallback path was needed: #165 settled that the event **does** reach stdout under `--output-format stream-json --verbose`, as the second event of the stream.

## What's here

- **`src/lib/quota/rate-limit-event.ts`** — the pure reader, plus the vocabulary (`quotaSeverity`, `describeRateLimitType`) the tile paints from. Total by construction: it runs on the stream-parse path of every turn, so an unreadable field becomes null rather than a throw that would fail a pass over a log line.
- **`quota_state`** — one durable row, latest wins, written by the parser **at the moment of observation** rather than when the turn settles (an interactive turn can run for an hour, and the fleet's freshest fact must not arrive last).
- **`FleetView.quota` + the tile** — under the spend gauge, because it answers the neighbouring question: spend is what the fleet has cost, quota is whether it may keep going at all.

Nothing acts on any of it. Pausing (#168) and admission (#171) are the tickets that decide.

## Three things the wire does that the ticket's stated shape doesn't

Each pinned by a test against #165's captured fixtures:

- **`utilization` and `resetsAt` are frequently absent** — not null, not zero. Absence is kept distinct from zero throughout, because a gate reading a missing utilization as 0% would describe a walled account as an idle one, and #171's threshold inherits exactly that reading. The tile omits the bar rather than drawing it at zero, which is the *usual* case on this account.
- **The fields nest under `rate_limit_info`.** Both shapes parse.
- **`isUsingOverage` / `overageInUse` ride along**, undocumented. Unused by the tile; stored, because #173 decides "already spending real money" from them and a record written without them can't be revisited.

Enums are held **verbatim as strings, never parsed against a union**: a member a later CLI adds must reach the screen, since the value the fleet couldn't interpret is precisely the one an operator needs to see. An unknown status paints in a tone that claims nothing.

Only one limit window appears because the event carries one, and that *is* the answer to "which is closest to tripping" — the server picks the representative claim, the window nearest its ceiling.

## Design notes worth a reviewer's eye

- **A table, not a column on `settings`.** The two have opposite lifecycles: `settings` stamps `updatedAt` when a human presses something, and folding in an observation written on every API attempt would make the settings screen's "last changed" report the fleet's traffic instead of the operator's last press.
- **A row, not an in-memory store** like the sweep's observations. A quota window is 5h/7d, longer than the gap between deploys — and the writer (stream parser, orchestrator graph) and reader (dashboard route handler, app-router graph) share nothing but the database (#159).
- **The row is stored in the wire's own encoding** and read back through `parseRateLimitEvent`, so there is exactly one reader of an observation in the codebase.
- **The recorder goes on logging `rate_limit_event`** after this ticket consumes it, against the letter of CLAUDE.md's "until #167 consumes it". The state row is latest-wins and parsed; the log is the only verbatim history, and the only place a field this build does not model survives. Its comment now says that instead of pointing at this ticket.
- **A reading whose stated reset has passed goes quiet** — words kept, colour and bar dropped. A red `rejected` over a wall that lifted hours ago is the tile crying wolf.

## Self-review

`/code-review` against `origin/main` with #167 as the spec. Three findings acted on in 4e8d063: the store's duplicate defensive readers (now one reader), the stale-rejection tone, and two parallel tone maps over the same severity type (now one).

Two findings noted and deliberately not changed:

- **`TurnResult.rateLimit` has no production consumer yet.** The ticket asks for it in as many words ("A `rate_limit_event` on the stream reaches the turn result with its fields intact"), and #168 reads it.
- **`utilization` is assumed to be 0–100.** #165 observed `utilization: 91` from a `*-utilization: 91` header, so this is the right reading of *that* field; the fractional `slow-budget-utilization` is a different header that doesn't map to it. If a future CLI ever sent a fraction here it would render as 0% — flagging rather than guessing at a clamp.

## Checks

`npx vitest run` — 60 files, 1125 tests, all passing. `npx eslint` clean on every changed file (the repo's lint baseline on `main` is untouched). `next build` succeeds.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/quota-tile-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/quota-tile.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0023_quota_state.sql` |
| persistence | `drizzle/**` | `drizzle/0023_quota_state.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0023_snapshot.json` |

lennons301-reviewer (approved):
Approving. Reviewed against #167's acceptance criteria, CLAUDE.md's conventions, and the #165 findings this ticket is built on. This PR is gated (`human-signoff`, auto-merge off), so the approval satisfies branch protection only — a human merges.

## Objective signals

No CI is configured on this branch, so I ran it locally in the worktree:

- `npx vitest run` — 60 files, 1125 tests, all passing (after `rm local.db*`, per CLAUDE.md's drizzle gotcha; no flakiness in `cross-graph-state` / `queue-slots` on this run).
- `npx eslint` over all 22 changed files — clean; the `main` lint baseline is untouched.
- `npx tsc --noEmit` — only the two pre-existing errors in `src/lib/docker/__tests__/container-manager.test.ts`, confirmed untouched by this PR (`git diff origin/main..HEAD` does not list it).
- Branch contains `origin/main` (ca55ff8), so nothing here is reviewed against a stale base.

## Acceptance criteria

- **Event reaches the turn result with fields intact** — `TurnResult.rateLimit`, pinned against both captured fixtures including the wire-only `isUsingOverage`/`overageInUse` pair.
- **Latest observation persisted with an observed-at** — single `quota_state` row, upsert, `observedAt` from the orchestrator's clock at parse time (the event carries none). Latest-wins is pinned both within a turn (`keeps the last event of a turn, not the first`) and across writes (`keeps the latest observation, in one row`).
- **Tile shows utilization / closest limit / reset / observed-at** — all four, plus the null cases.
- **Renders sensibly with nothing observed** — and specifically asserts it does not read as `0%`, which is the failure that mattered.
- **A stream with no rate-limit event parses exactly as today** — the strongest test in the set: every other `TurnResult` field *and* the whole message feed compared between the stripped and unstripped replays, with the stripped one run first so "nothing persisted" is a fact about that replay.
- **An unrecognised enum does not throw** — enums held as `string`, `quotaSeverity`/`describeRateLimitType` both total, covered at parser, read-model and tile layers.
- **Fixture-driven against the real captured stream** — yes, through the default sink so the wiring is exercised rather than a stand-in.

## Correctness checks I made on the new read/write path

- `recorder.streamEvent` is still called *before* the type switch in `handleEvent`, so adding the `rate_limit_event` arm with an early `return` does not cost the #165 JSONL log its verbatim copy. That was the easiest thing to break here and it isn't broken.
- Storing in the wire's encoding and reading back through `parseRateLimitEvent` genuinely leaves one reader: a retired-shape row falls out as `null` via the same `status`-is-required check the stream uses (`isRecord(rate_limit_info) ? … : event` degrades correctly for a non-record column), and a JSON column that won't decode throws inside the `try`. Round-trip keeps `null` distinct from `0` for both `utilization` and `resetsAt` — the distinction #171 inherits — and the seconds-precision loss on `resetsAt` is the wire's own granularity, not a truncation.
- Write failures swallowed with a log, on the stream-parse path of every turn — consistent with the recorder's rationale, and pinned by a test.
- `quota-store` holds no module-level state, so #159's `globalThis` rule doesn't apply even though writer and reader sit in different module graphs; the DB is the only shared thing, as the comment claims.
- `Gauge`'s tone widening is purely additive (`red → amber → cool → green` default), and `tone !== "quiet"` narrows `keyof typeof TONES` exactly onto `Gauge`'s union — no cast, so the "quiet draws no bar" rule is type-enforced rather than remembered. `--fl-amber` exists in both themes and `fleet-tokens.test.ts` still passes.
- `formatCountdown` returning `null` (rather than `"0m"`) for a past instant is what makes the quiet-when-spent inference a single expression: `resetsAt !== null && resetsIn === null`. Judged against the dashboard's ticking `now`, consistent with `RunningList`/`RecentLedger`.
- Migration hygiene: `_journal.json` `when` is exactly +1 day past 0022, `0023_snapshot.json`'s `prevId` chains from 0022's `id`, and the DDL matches the schema (json text + `timestamp_ms`).
- Scope: `settings`-vs-table and row-vs-memory are both argued from lifecycles rather than taste, and nothing acts on the observation — the ticket's explicit boundary is respected. The one inference beyond the ticket (quiet once the stated reset has passed) is small, justified, and tested.

## Non-blocking suggestions

1. **A stored row this build cannot decode is silent.** `getQuotaObservation` logs only when the read *throws*; a row that exists but yields `null` from `parseRateLimitEvent` returns quietly, and the tile then says "not observed yet" / "no pass has reported a limit window" — the same words as "never seen one". A `console.warn` on that one branch (`row && parsed === null`) would keep the repo's own "write down what it does not understand" discipline for the single decoding regression this path can have. The no-crash half, which is the half that matters, is already pinned.
2. **`unknown` severity suppresses the gauge even when a utilization *was* reported**, because `tone === "quiet"` gates both the colour and the bar. A future status with a real `91` would print `91%` with no bar beside it. Deliberate per the comment, and the number still reaches the screen — just noting that one flag is carrying two decisions ("a status I can't vouch for" and "a figure I can").
3. **Cosmetic:** in `rate-limit-fixture.test.ts`, the stripped-`ALLOWED` replay is fed under taskId `"task-rejected"`, which reads as the rejected fixture on a skim. The ordering rationale in the comment is right; the id just fights it.

None of these are tied to a stated requirement or standard, so none block. Nice work on the absent-vs-zero discipline in particular — that is the field the two downstream tickets will be wrong about if this one is.

Issue #175, opened 2026-08-26
OpenRouter lane: prove the third-party-provider path end to end on a cheap open-weights model


lennons301:
## Parent

#164

## What to build

#164 justifies the single-adapter design on the grounds that Kimi K2, GLM and MiniMax
expose Anthropic-compatible endpoints, so flipping providers is a base-URL plus token plus
model-id change. That reasoning is sound but **unproven** — no lane other than the Claude
subscription is exercised anywhere in the milestone, and #172 explicitly ships one adapter
without a second provider to test it against. An interface designed against a hypothetical
provider is an interface that breaks on contact with a real one.

This ticket ships and verifies a **second, real lane** against an aggregator, which is the
configuration with the best cost optionality: one credential and one base URL giving access
to the whole open-weights field, so changing model is a tier-map edit rather than a new
credential.

**Finding (verify before building, 2026-08-26):** OpenRouter exposes a native Anthropic
Messages endpoint — their "Anthropic Skin" — at `https://openrouter.ai/api`, reached by
setting `ANTHROPIC_BASE_URL` to it, `ANTHROPIC_AUTH_TOKEN` to an OpenRouter key, and
`ANTHROPIC_API_KEY` to empty. Extended thinking blocks and native tool use are documented
as passing through, which is what the agentic passes depend on. **No proxy and no second
harness adapter should be needed.** If verification shows otherwise, stop and report rather
than building a translation shim — that is a scope change, not an implementation detail.

Three things are genuinely uncertain and are the substance of the work, not incidentals:

1. **Cost reporting.** Budget enforcement reads the cost the CLI reports on a turn. Under a
   third-party base URL that figure may be absent, zero, or computed at Anthropic's prices
   rather than the lane's. Every spend guard in the milestone — per-attempt budget, daily
   cap, the real-money cap in #174 — is downstream of this. If reported cost is not
   trustworthy on a third-party lane, the lane must derive spend from its own configured
   per-token prices and token counts instead.
2. **Quota telemetry.** `rate_limit_event` is an Anthropic subscription construct. A metered
   third-party lane will not emit it, so the quota state, the degrade ladder and the
   admission gate must be understood as **per-lane** and must not read a stale Anthropic
   observation as authoritative for a lane that cannot produce one. A lane with no quota
   signal is gated by spend, not by utilization.
3. **Prompt caching.** Cache semantics and cache-read pricing differ per provider and
   materially change the real cost of a long agentic pass. Record what the lane actually
[...]

Issue #81, opened 2026-08-04
Reasoning effort per pass kind — investigate the CLI mechanism and wire it like AGENT_MODEL


lennons301:
Owner-raised alongside #74/#79 (2026-08-04): model choice is now configurable, but **effort** (reasoning/thinking depth) is not — and it's the other half of the cost/quality dial. A triage pass at high effort wastes money; an implement pass on a hard ticket at low effort wastes attempts.

**First step is investigation** (the mechanism is uncertain): what does the current headless CLI actually expose — an effort flag, `MAX_THINKING_TOKENS`-style env, or settings.json fields — and what are its interactions with `--model`? Then wire the winner exactly like #74: `AGENT_EFFORT` (+ per-kind overrides) from Doppler, per-exec, recorded on the run row, and optionally a clamped `effort:` ticket directive to pair with the model directive.

Out of scope until measured: any default other than the CLI's own.

lennons301:
Armed `ready-for-agent` under Sean's standing authorization (2026-08-04) — joins the pilot-fix chain on the laptop loop. Human decision, agent applied the label.

lennons301:
🤖 Attempt 1/3 starting.

PR #180, opened 2026-09-02
Execution lanes: checked-in lane config, pure resolver, and harness adapter seam

### LINKED ISSUE #172: Execution lanes: checked-in lane config, pure resolver, and harness adapter seam

lennons301:
Closes #172

Which harness runs a pass, against which provider, on which model becomes
**configuration** rather than code.

## What's here

- **`lanes.yaml`** at the repo root declares each lane as `{adapter, auth env
  var names, base URL, tier→model map, caps, billing kind}`. Secrets appear
  only as variable *names*: `auth` maps the variable the *harness* reads to
  the orchestrator variable holding the secret (Claude Code reads
  `ANTHROPIC_AUTH_TOKEN`; the OpenRouter key is provisioned as
  `OPENROUTER_API_KEY` — hence a mapping, not a list), and the parser enforces
  `^[A-Z][A-Z0-9_]*$` on both sides, so an inlined credential is a parse error
  rather than a secret in git. The production image `COPY`s the file, since
  Next's standalone output carries only what the bundle imports.
- **`src/lib/lanes/resolve.ts`** is pure — `(catalog, pass kind, config,
  overrides, env)` in, `{auth values, base URL, model id, caps}` out. It
  **reports rather than falls back**: an unavailable lane refuses *before* a
  container is provisioned, naming the missing variables, replacing a live
  agent dying inside the harness with "Not logged in". Only the *unset*
  default walks the file's `primary` order (subscription, then API key —
  exactly what `config.ts` did before); an explicit choice from the settings
  screen or `AGENT_LANE` is honoured even when broken, because routing around
  an operator's choice is how a fleet spends money nobody authorised.
- **`src/lib/harness/`** is the adapter seam — `buildExecEnv`, `buildCommand`,
  `createOutputHandler`, chosen because those three were already pure and
  already tested. One adapter ships (`claude-code`); nothing in the interface
  names a vendor. `container-manager.ts` is harness-agnostic below it.
- The **primary lane is settable on the settings screen**, and `runs.lane`
  records which lane a run used.

## Security constraints, each pinned by a test

- No model-provider credential in the **persistent** container environment —
  `ANTHROPIC_API_KEY` used to be, and is not any more; the `anthropic-api`
  lane supplies it per exec.
- `ghToken` stays null for every autonomous pass kind.
- No lane **secret** is stored in the DB or served by any API route: the
  settings payload carries variable names only. (The row does store the chosen
  lane's *id*, and `runs.lane` the id a run used — ids, not credentials.)
- Agent containers still get no host bind mount; the reviewer PAT still never
  enters a container.

## Default-behaviour equivalence

`claude-subscription`'s tier map (`opus`/`sonnet`/`haiku`) is the pre-lane
`TIER_MODEL_IDS`; an unset `AGENT_MODEL` still passes no `--model`; the
`primary` order reproduces "subscription token, else API key". One real
change, and it is the intended fix: with *both* credentials set, a container
previously saw persistent `ANTHROPIC_API_KEY` **and** exec-scoped
`CLAUDE_CODE_OAUTH_TOKEN`. Now only the chosen lane's variables are present,
so two credentials can never race to authenticate one turn. An install that
was actually authenticating via the API key while holding a stale OAuth token
would flip to the subscription — set `AGENT_LANE=anthropic-api` there.

## Self-review

Ran `/code-review` against `origin/main` with #172 as the spec. Both axes are
acted on in `3cd0f9d`; the two correctness findings (both raised
independently by each axis) were:

- a stored `primaryLane` naming a lane a deploy had renamed was sanitized away
  on read, making the dangling-choice report unreachable and letting the next
  unrelated PATCH erase the choice permanently;
- `requireLane` in `processQueuedMessages` threw into the poll loop's bare
  `.catch(console.error)` — the task stayed `running`, the message stayed
  queued, and the loop retried every 2s forever with nothing on the feed.

Three findings I did **not** act on, with reasons:

1. *"`runs.lane` is written only for the implement pass, so a review pass on a
   flipped lane is misattributed."* True, and deliberate: `runs.model` and
   `runs.effort` are implement-only for the same reason (#74, #81) — a run has
   one row and the implement pass is the substrate its spend should be read
   against. One column cannot hold two lanes; per-pass attribution wants a
   column on `tasks`, which is a bigger change than this ticket asked for. The
   schema comment states the scope explicitly.
2. *"`caps.daily_budget_usd` is parsed and displayed but never enforced."*
   Correct — the ticket asks for caps as part of the lane bundle; spending
   against them is #175. I fixed the `lanes.yaml` comment that overstated
   this, so the file no longer claims an enforcement that does not exist.
3. *"`runs.model` now holds a tier rather than a resolved id — a schema
   semantics change #172 did not request."* It is required by the ticket, not
   incidental: the column is read back as the run's `model:` directive on
   every later pass, and a lane-specific id (`anthropic/claude-sonnet-4.5`)
   names no tier, so recording it would silently drop the directive the moment
   the fleet left an alias-mapped lane. Old `opus` rows and new `heavy` rows
   both normalise through the aliases #166 added for exactly this.

Flagged as scope beyond the ticket, kept: the settings screen's two panels
were split into presentational panels behind one `SettingsOverrides` state.
That is not decoration — fetched independently, the model rows named
identifiers from the pre-lane map while the lane panel named the lane in
force, which is the provenance failure #166 exists to prevent.

## Validation

`npx vitest run` — 1205 passed, 64 files. `npx eslint` clean on every changed
file (repo lint has a known baseline on `main`). `npx tsc --noEmit` clean.
`npx next build` succeeds. `npx drizzle-kit check` + `generate` report no
drift; `main`'s `0023_quota_state` was merged in and the lane migration
renumbered to `0024_run_lane`.

Touches five review-gate categories (Dockerfile, docker, config, schema/drizzle,
components) — `human-signoff`, never auto-merge.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/model-tier-settings-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/execution-lane-settings.tsx` |
| visual-ui | `**/components/**` | `src/components/model-tier-settings.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0024_run_lane.sql` |
| agent-sandbox | `Dockerfile` | `Dockerfile` |
| persistence | `drizzle/**` | `drizzle/0024_run_lane.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0024_snapshot.json` |

lennons301-reviewer (approved):
**Approving.** This PR is gated (`human-signoff`, auto-merge off), so this satisfies branch protection only — a human still merges.

## Objective signals (no CI on the branch, so run locally)

- `npx vitest run` — **1205 passed, 64 files**, exit 0 (after `rm local.db*`, per the repo's migration note).
- `npx eslint` on all 27 changed `.ts`/`.tsx` files — clean, exit 0. (The known `main` lint baseline is untouched.)
- `npx tsc --noEmit` — clean.
- Branch is `MERGEABLE`, `main` is an ancestor, no drift.

## Acceptance criteria, each checked against the diff

| Criterion | Where |
| --- | --- |
| Lanes declared in a checked-in file, secrets by env var name only | `lanes.yaml`; `parseAuth` enforces `^[A-Z][A-Z0-9_]*$` on **both** sides, so an inlined `sk-…` is a parse error. `lane-config.test.ts` additionally asserts the *shipped* file contains no `sk-`-shaped value |
| Tier map → concrete model id per pass kind | `resolveLane` → `lane.models[tier]`, with `resolveAgentModelChoice` stopping at the tier |
| Absent variables reported, not failed at exec | `laneMissingEnv` → `{ok:false, reason}`; `requireLane` is called *before* `createWorkspaceContainer` |
| Command + exec env built from a resolved lane, base URL included | `buildTurnEnv` / `buildClaudeTurnCommand`, `execAgentTurn` takes `{command, env}` and nothing else |
| One adapter, existing passes unchanged | `claudeCodeAdapter`; `lane-config.test.ts` pins `claude-subscription` to `{heavy: opus, standard: sonnet, light: haiku}` — the pre-lane `TIER_MODEL_IDS` verbatim, and an unset `AGENT_MODEL` still emits no `--model` |
| No long-lived credential in the persistent container env | `ANTHROPIC_API_KEY` removed from `createWorkspaceContainer`; pinned by `container-manager.test.ts:259` |
| Autonomous kinds carry no `gh` token | `ghToken: opts.isGenerationSession ? gitAuthToken : null`; pinned by `claude-code.test.ts:156` |
| No lane secret in the DB or on an API route | `LaneView` carries names only; pinned twice — `resolve.test.ts` ("carries variable names only") and the route test ("never serves a lane secret") |
| Primary lane settable in the UI | `ExecutionLanePanel` + `primaryLane` in `SETTINGS_FIELDS`, with the runtime vocabulary passed in as `SettingsContext` so an undeclared lane is refused *by name* |
| Ledger records the lane | `runs.lane`, migration `0024_run_lane`, journal `when` bumped +1 day per the repo's drizzle note |
| Pure unit tests for resolution | `resolve.test.ts` (391 lines) — no filesystem, no provider, `env` as a parameter |

Also confirmed the issue's implementation note is honoured and *pinned*: `openrouter.auth` is exactly `ANTHROPIC_AUTH_TOKEN: OPENROUTER_API_KEY`, asserted in `lane-config.test.ts`.

## Things I checked because they are how this breaks in production, not in tests

- **The file reaches the image.** `COPY . .` in the build stage (`.dockerignore` does not exclude `lanes.yaml`), `COPY --from=build /app/lanes.yaml ./lanes.yaml` in run, `WORKDIR /app`, and `getLaneCatalog` reads `path.join(process.cwd(), …)` — cwd is `/app` for `node custom-server.js`. Consistent.
- **`yaml` is already a direct dependency** (`^2.9.0`), so no new package and no reliance on a transitive resolve.
- **Back-compat of the `runs.model` semantics change.** Pre-lane rows hold `opus`/`sonnet`/`haiku`, which are exactly the `MODEL_TIER_ALIASES`, so old rows normalise. `runs.model` is rendered on no surface (only `decide.ts` references it, in a comment) and is read back solely as `run?.model` → `normalizeModelTier`. The self-review's finding #3 reasoning holds; the mixed vocabulary is invisible.
- **The two failure paths.** `requireLane`'s throw is now inside the `try`, so an unavailable lane fails the task with the reason on the feed rather than leaving it `queued` for a 2s re-pickup; `laneForFollowUp` reports instead of throwing, because the poll loop's `.catch(console.error)` is not a failure path. Both are the right shape, and the dedup comparison against `JSON.stringify({ text })` does match `insertSystemMessage`'s stored form (line 1916).
- **The `getLaneCatalog` module-level cache vs. the #159 two-module-graph rule.** It is read by both the orchestrator graph and the settings route handler, so there will be two copies — but it is immutable data derived from a checked-in file, nobody mutates it cross-graph, and the cost is one extra `readFileSync` per graph. Correctly *not* a `processSingleton` case; the mutable half (the lane *choice*) lives on the settings row and is re-read at the point of use, which is the rule #166 set.
- **The `sanitizeOverrides` asymmetry** (read path without the catalog, write path with it) is the right way round, and the reasoning in `settings.ts` is worth the lines it takes.

The smell baseline finds this net-negative: `TierOption`/`fallback` were deduplicated into `ChipRadio`/`fallbackNote` in `fleet-bits`, and `container-manager.ts` shed the vendor knowledge it had accreted.

## Non-blocking notes (suggestions, not conditions)

1. **`AGENT_LANE` is undocumented for operators.** `docs/runbook.md`'s "Key environment variables" table lists `AGENT_MODEL*` but not `AGENT_LANE`, and `.env.example` doesn't mention it or `OPENROUTER_API_KEY`. `AGENT_LANE` decides which account gets billed, which makes it a stronger candidate for that table than some entries already in it. The settings panel does name the variable and its missing credentials on screen, so this is a gap rather than a hole.
2. **Circular import for `FALL_THROUGH`.** `settings-overrides.tsx` imports both panels, and both panels import `FALL_THROUGH` back from it. It resolves fine (the const is only read at render time, and the render tests pass), but `lane-id.ts` in this same PR makes the argument against it better than I can — "a leaf module with no imports, because two layers need the shape and neither should drag the other in". `FALL_THROUGH` beside `ChipRadio` in `fleet-bits` would close it.
3. **Blast radius of a fleet-wide lane misconfiguration.** An unavailable lane routes an implement pass into `interruptImplementPass`, so every armed ticket burns its 5 interruptions and lands on `ready-for-human` before anyone fixes the variable. Bounded and visible, and I think it beats the alternatives — just worth the operator knowing that the recovery is re-labelling, not just setting the variable.
4. **Doc drift in `CLAUDE.md`:** "the *adapter* owns `baseUrlEnvVar`" names a member the `HarnessAdapter` interface does not have — the adapter module exports `CLAUDE_CODE_BASE_URL_ENV` and settles the mapping inside `buildExecEnv`. The claim is right, the name in it isn't.

None of the four changes what the code does. The security constraints the ticket called out are each pinned by a test rather than by prose, the default-behaviour equivalence argument is testable and tested, and the one real behaviour change (both credentials set → only the chosen lane's variables present) is stated in the PR body with the remediation for the install it would affect.

Issue #169, opened 2026-08-26
Resume a rate-limited run losslessly once its window resets


lennons301:
## Parent

#164

## What to build

A paused run **resumes by itself** once the limit resets, continuing the same agent
conversation rather than starting over.

The existing 30-second reconciliation sweep is the resume driver — no new scheduler. When
a run's `resumeAfter` has elapsed it becomes eligible again, subject to the same slot
admission as any other work, and **resuming existing work outranks claiming new work**.

Resume is lossless: the session transcript is copied out of the container before it is
torn down on pause, persisted outside the container so it survives an orchestrator
restart, and restored into the fresh container on resume so the pass continues the same
conversation on the same branch. If the spike found that resuming against a restored
transcript does not work, apply its fallback — restart the pass on the same branch and
note in the run that prior context was lost.

Resume times are jittered so that a fleet-wide pause does not stampede every run at the
same instant.

A per-attempt bound on how many times a run may pause and resume stops a pathological
ticket looping forever across quota windows; exceeding it routes the ticket to
`ready-for-human` the way exhaustion does. The bound is settable in the UI.

## Acceptance criteria

- [ ] A run whose `resumeAfter` has elapsed is resumed by the sweep
- [ ] Resume provisions a fresh container and restores the session transcript
- [ ] The resumed pass continues the same conversation on the same branch
- [ ] The stored transcript survives an orchestrator restart
- [ ] Resuming existing work is preferred over claiming new tickets when slots are scarce
- [ ] Resume times are jittered so a fleet-wide pause does not stampede
- [ ] A per-attempt pause bound routes to `ready-for-human` when exceeded
- [ ] The pause bound is settable in the settings UI and falls through to its default when unset
- [ ] Resume and bound decisions are expressed in the autonomy reducer and covered by table tests

## Blocked by

- #168

lennons301:
🤖 Attempt 1/3 starting.

lennons301:
🤖 Attempt 2/3 starting.

PR #181, opened 2026-09-02
Pause an autonomous run on an account-wide rate limit instead of failing it

### LINKED ISSUE #168: Pause an autonomous run on an account-wide rate limit instead of failing it

lennons301:
Closes #168

A quota rejection did not look like a quota rejection. The CLI reports it as
`subtype: "success"` with `is_error: true` / `terminal_reason: "api_error"` /
`api_error_status: 429` (#165's finding), and `subtype` was the only field the
orchestrator read — so a walled implement pass looked like a pass that finished
having done nothing, fell into #106's empty-pass path, and spent one of the
ticket's three attempts. Three of those routed a perfectly good ticket to
`ready-for-human` because the account ran out of window.

## What it does

A pass the account's quota **refuses** now parks its run: `runs.status =
rate_limited` carrying `resumeAfter`, taken verbatim from the limit event's own
reset time. The container is torn down, and neither the attempt counter nor the
interruption bound moves — both keep measuring what they say they measure.

**Two signals, and it takes both**, because each alone lies:

- the **exit condition** says the pass ended on an API error but reports
  `subtype: "success"` at a wall and cannot name which error;
- the **rate-limit event** says the account was rejected, but it is emitted per
  API attempt and describes the *account*, not this pass — a pass can observe a
  rejection, be retried past it, and still finish its work. Pausing that run
  would park a finished attempt on a five-hour clock.

The decision is the reducer's (`pauseRunOnRateLimit`), emitted **ahead of both
other readings of a finished pass**: a refused turn never reached the model, so
its "final message" is the CLI's own session-limit line (which the
blocked-marker detector would read as a question nobody asked) and its empty
diff is the wall's, not the work's.

The container is **removed, not stopped** as a blocked run's is — a parked
container holds ~2 GiB while holding no slot (the 2026-08-04 wedge), and a
five-hour window is far too long to hold one for.

The run-status vocabulary moved to one leaf module (`run-status.ts`) so both
questions it answers can be tested without loading Docker. Both lists are
**inclusion** lists, so `rate_limited` is *active* (the ticket is still this
run's; claiming a second run over it would spend the very attempt the pause
protects) and *not reclaimable at boot* (it waits on a clock, not a lost turn,
exactly as `gated`/`blocked` wait on a human).

On the surfaces it stays under **Running**, labelled paused with its countdown,
and never reaches needs-you — nobody has to *do* anything about a quota window.

## Tests

Four levels, each pinning what only it can see:

- `detectQuotaRejection` — the boundary: a warning, a rejection observed by a
  pass that finished anyway, a 429 with no reset time, a container death with no
  result event, a metered lane. None of them pause a run.
- The **captured CLI streams** (#165's fixtures) — that both signals survive the
  parser off real bytes, and that an ordinary turn on a healthy account (which
  also carries a `rate_limit_event`) is never read as a wall.
- `decideNext` table tests — precedence over the blocked and empty-pass
  readings, the parked-run guard, and that neither the kill switch nor a
  project's autonomy toggle is loosened by a pause.
- The pass-completion seam over a real migrated ledger — where the two
  accounting ACs actually live: `attempt` and `interruptionCount` unmoved, the
  container removed rather than stopped, the run left non-terminal.

Plus the surfaces: the card reads paused with its countdown, never enters
needs-you, holds no slot; the digest says the same thing from the same field;
boot recovery's reclaim query passes over a paused run.

`pnpm test` 1168 passing, `pnpm build` clean, eslint clean on changed files
(`tsc` has the two pre-existing `container-manager.test.ts` baseline errors from
`main`).

## Self-review (`/code-review` vs `origin/main`, issue #168 as spec)

Acted on: a stale `PassDecision` doc and call-site comment; `RateLimitedTurn`
renamed `TurnQuotaSignals` (the old name asserted the conclusion its own
function computes); the executor's third restatement of `{runId, resumeAfter,
limitType}` replaced by the extracted `Action`, as the sweep's executors take
theirs; a comment claiming implement is "the only pass the wall can interrupt"
(it is the only pass this ticket *pauses*); and CLAUDE.md + the operator runbook.

Where I disagreed, stated rather than silently ignored:

- **Review passes deliberately do not pause.** The harm this ticket names is the
  attempt budget, which only an implement-shaped pass spends; a walled review
  already fails closed to a human without a strike, and #171 is the ticket that
  stops a pass starting under a wall at all. Handling review here would also
  need the run to remember which status to return to, which is #169's design to
  make, not this ticket's to guess.
- **A rejection with no reset time takes the ordinary path** (and spends the
  attempt). A run paused on an unknown window is stranded where no later ticket
  can find it, and inventing a window length the fleet does not know would be
  wrong in the direction that costs a real attempt. Narrower than it sounds:
  #167 records `resetsAt` as *present on a rejection* and absent from some
  `allowed_warning` events, and the captured rejection fixture carries it.
- **The pass's task goes `failed`, not `cancelled`.** Purely cosmetic (both are
  terminal to every reader), and it matches the closest precedent —
  `interruptImplementPass`, which is also "not the work's fault" and also gives
  its run a non-failure status. `cancelled` in this repo reads as a deliberate
  human act.
- **Scope**: the issue comment mirrors every other lifecycle transition (claim,
  fail, interrupt all comment) and is the only place an operator learns why the
  run stopped; the digest line reads the *same* `paused` field the dashboard
  does, because a digest calling a paused run "running" is exactly the
  surfaces-disagreeing failure this repo builds against; `evaluatePassOutcome`
  is exported as the seam the two accounting ACs are asserted at.

## Known, and deliberate

Until #169 lands **a paused run stays paused** — nothing resumes it, and the
sweep will not claim a fresh run over its ticket either, because a paused run
still holds it. That is the ticket's own decision ("after this one a paused run
stays paused"); the alternative is spending attempts on quota. The runbook now
says so, and says how to move one by hand.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/running-list-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/running-list.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0025_run_rate_limited.sql` |
| persistence | `drizzle/**` | `drizzle/0025_run_rate_limited.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0025_snapshot.json` |

lennons301-reviewer (approved):
Approving. Reviewed with fresh eyes against issue #168 and the repo standards; verified objective signals locally since no CI reports on this branch.

## Objective signals

- `pnpm test` — 68 files, **1248 passing** (fresh `rm local.db*` first, per CLAUDE.md's migration note)
- `npx tsc --noEmit` — **clean** (the two `container-manager.test.ts` baseline errors the PR body mentions don't reproduce here; either way, nothing new)
- `npx eslint` on every changed source/test file — **clean**
- Migration hygiene: journal entry `0025` is exactly one day past `0024`'s `when`, sequence unbroken, `resume_after integer` is the right shape for `timestamp_ms`; `createTestDb` migrates and the `run-status` test seeds a `rate_limited` row through it

## Acceptance criteria, each traced

| AC | Where |
|---|---|
| `rate_limited` + `resumeAfter` from the event's reset | `pauseRunOnRateLimit` (turn-manager) writes both; `quota-pause.test.ts` asserts against a real migrated ledger |
| container torn down | `removeContainer`, asserted as `docker.calls == ["removeContainer"]` — i.e. *not* `stopContainer`, which is the mistake worth pinning |
| attempt unchanged | seeded `attempt: 2`, asserted still 2 |
| interruption unchanged | seeded `interruptionCount: 1`, asserted still 1 |
| dashboard paused, with resume time | `buildFleetView` → `RunningCard.paused`; pinned in `fleet-view.test.ts` and rendered in the new `running-list-render.test.tsx` |
| not in needs-you | `needsYou` only ever filters `blocked`/`gated`/`exhausted` — I checked the whole builder, `rate_limited` cannot reach it — plus an explicit test |
| boot recovery leaves it | `RECLAIMABLE_RUN_STATUSES` excludes it, and **both** boot passes read that one list: `markInterruptedRuns` and `finalizeDanglingRuns`. That second one is the trap I went looking for — a paused run has all tasks terminal and (on an initial-turn wall) no PR on the run row, which is #106's dangling shape exactly; it is saved by the status filter, not by luck |
| kill switch / project toggle | reducer tests: a pass that already ran still pauses, while `claims` stay empty and `pausePickup` still fires |
| decision in the reducer, table-tested | `pauseRunOnRateLimit` action in `decideNext`, 11 table tests |

## What I checked beyond the ACs

- **Ordering in `startTask`.** The pause sits after the `!producedResult` interruption check and after `attemptExhaustion` — both correct: a wall *does* emit a terminal result (so it isn't misread as a container death), and it costs ~\$0 (so it doesn't trip budget exhaustion). Within `decideNext` the wall is read first, ahead of the blocked-marker and empty-pass readings, which is the substance of the fix and is pinned by two tests using a `BLOCKED:`-leading final message and `producedPr: false`.
- **`detectQuotaRejection` really needs both signals**, and reads them defensively: a null `terminalResult` explicitly is *not* a refusal (leaving container deaths to #97's bound), `resetsAt === null` falls through to the ordinary path, and the status/window enums are compared as strings so an unknown window still pauses. The "either `api_error_status` or `terminal_reason`" disjunction is the right defensiveness for a field #165 found to be the only discriminator.
- **The fixture tests are the ones that would have caught the original defect**: the captured *allowed* stream also carries a `rate_limit_event`, and it must not pause. Both directions asserted off real bytes.
- **Slot accounting.** The task goes `failed` (terminal per `taskIsFinished`), so `occupiedSlots` and `slots.used` both drop it, and the container handle is read from `activeTasks` *before* the terminal write — the #159 lesson, and the comment says why. `slots.used == 0` beside a paused card is pinned.
- **Repair passes.** `processQueuedMessages` gates on `implement | repair`, so a walled repair pass parks too. That's coherent (a repair pass has already moved its run to `implementing` and its PR is on the ledger) even though the PR body frames the scope as implement-shaped — worth knowing, not wrong.
- **Drain loop.** `break` on `"paused"` is right; the container is gone, so continuing would exec into nothing. Queued messages stay undelivered, same as the `failImplementAttempt` break above it.
- **Interactive work is untouched**: `evaluatePassOutcome` returns `proceed` with no `runId`.

## Non-blocking suggestions (for #169, not for this PR)

1. `pauseRunOnRateLimit` doesn't copy the task's PR fields onto the run row, where `failImplementAttempt` deliberately does ("keep the PR on the ledger"). A wall on the initial turn happens *after* `runPostTurnCommitAndPush`, so a draft PR can exist on the task while the run row reads PR-less. No visible consequence today (the paused card and digest line show no PR, and `finalizeDanglingRuns` skips the status), and #72's PR adoption covers a manual re-claim — but a resume in #169 will want it.
2. `finalizeDanglingRuns`' docstring still enumerates only `gated`/`blocked` as the statuses it leaves alone. The behaviour is correct via the shared list, whose own doc covers `rate_limited`; only the prose is now one status short.
3. The AC "the global kill switch and the per-project autonomy toggle still stop the run" is read here as "neither control is loosened by a pause", which matches the repo's documented semantics (an engaged switch holds *pickup*, never an in-flight turn). Reasonable — but when #169 makes a resume a pickup-shaped action, that action is the one that has to sit behind the same gate.

## What the human merging should know

This is the ticket's own decision, stated in the issue and now in the runbook, but it is the one operational consequence: **until #169 lands, a walled run parks forever.** It holds its ticket (so no fresh run is claimed over it), it never enters needs-you (by design), and past its reset the card reads "quota window has reset" indefinitely. The trade is deliberate and the right way round — before this, three walls routed a good ticket to a human — and `docs/runbook.md` documents the manual nudge. Correctly gated: the diff touches `src/db/schema.ts` + `drizzle/**` (persistence), `src/lib/orchestrator/autonomy/**` (autonomy-control) and `src/components/**` (visual UI).

Standards check: no `new Octokit(`, no new unbounded outbound I/O, the issue comment is fire-and-forget with the same rationale the interruption path gives, the status vocabulary is a single leaf module with inclusion lists (so a later status is safe by default), and CLAUDE.md + the runbook are updated in the house style. The one smell I'd name is Duplicated Code — `pauseRunOnRateLimit` is the third function with the read-task/take-container/insert-message/terminal-task/update-run/comment/remove shape — but that shape predates this PR (`interruptImplementPass` already mirrors `failImplementAttempt`), the three differ in exactly the parts that matter, and following the local precedent is the right call over extracting a four-way-parameterised helper here.

PR #182, opened 2026-09-02
Quota admission gate: stop claiming new tickets when the window is nearly spent

### LINKED ISSUE #171: Quota admission gate: stop claiming new tickets when the window is nearly spent

lennons301:
Closes #171

The fleet stops claiming new tickets when quota is nearly gone, so it does not start work it cannot finish.

## What it does

`evaluateQuotaGate` (`src/lib/quota/quota-gate.ts`) is one pure gate that two callers share: `decideNext` refuses pickup with it, `buildFleetView` names the hold with it. A gate computed twice is a gate that eventually disagrees with itself, and the dashboard saying "running" while the reducer claims nothing is the exact confusion #148 exists to remove.

It holds **new pickup only** — a claim, and the triage pass that is also pickup. Everything already in flight is decided above it and untouched: refusing to *finish* work would waste the quota already spent on it, and a parked run resuming is that same work continuing.

Precedence, stated once in each surface and identical in both: the kill switch outranks it (a human's hold beats an observation, and naming quota under an engaged switch would point the owner at a wall instead of at the control they hold); it outranks the slot check (a full box empties by itself in minutes, a spent window does not); on the dashboard the daily cap outranks it (of the two self-lifting ceilings, the cap's is the one a human chose). Per-project holds are decided further up and keep saying their own piece.

## Three rules that make it safe to gate on telemetry the fleet doesn't control

- **No observation is not a closed gate.** An API-key lane emits no `rate_limit_event` at all (#165's finding 6). Silence the fleet can never break must not read as a wall.
- **An absent utilization is not 0%.** #167's rule, inherited verbatim. Only `rejected` gates on status — a member a later CLI adds reads as itself and decides nothing, so the fleet is never stopped by a word this build has not heard.
- **A spent observation stops gating.** Load-bearing rather than tidiness: only a pass making an API call produces a fresh observation, so a gate held by a stale rejection would suppress the very traffic that would lift it. An observation expires at its stated reset, or at `QUOTA_OBSERVATION_STALE_MS` when it stated none.

## The Discord ping

Fires from *past* the eligibility filter, which is what makes it mean "the fleet is stalled on quota with N tickets armed" rather than "a run paused". Its once-per-transition flag is cleared from the **gate**, not from whether a pause was emitted — the reducer only pauses when there is work to hold, so clearing on "no pause" would re-arm the ping every time the backlog emptied under a standing wall and fire it again on the next armed ticket.

## The setting

`Settings → Quota`, falling through to `QUOTA_PICKUP_THRESHOLD_PERCENT` and then to a built-in 90. A single threshold with **no headroom reserved for the owner**, per the ticket: the fleet and the owner's own sessions draw on one pool.

It follows the **lane field's** shape rather than the tiers': it shares `SETTINGS_FIELDS` — the allowlist, and so the mechanism that decides what is settable at all — but has its own resolver, its own view model and its own panel, because asking a percentage "what tier is in force?" is not a meaningful question. It resolves through three layers where a tier has two: a tier may resolve to "pass no `--model` and let the harness decide", but a gate needs a number, so an unset override *and* an unset variable both land on the built-in default. The environment is read through the same `normalize` an override is — a typo is refused there too, never clamped — but is still *shown* verbatim, because a refused value collapsed to "unset" would read back on screen as a variable nobody had set.

The panel joins the other two in one client state even though it is independent of them: one PATCH returns the whole resolved settings state, so a separately-fetched panel would be left showing a row that is no longer the row.

**Judgement call worth flagging:** the control is a fixed set of chips (50/70/80/85/90/95/100) rather than a free number input. The spread offered is finer than the decision it feeds — a fleet gated at 87 rather than 85 behaves the same — a value outside it is refused *with the list* rather than clamped, and the row keeps the one-press `environment` way back in the voice the rest of the room speaks. If it proves too coarse, a number input is a contained change.

## Self-review

`/code-review` against `origin/main` with #171 as the spec raised three low-severity findings. Two are fixed in the last commit:

- **Duplicate Discord embed across module graphs.** A webhook-triggered sweep runs on the app-router graph (#159), where a module-level flag is that graph's own and freshly false. The flag moved to `processSingleton`. The two older flags beside it (`saturationAnnounced`, `dailyCapAnnouncedDay`) are left alone deliberately — they are part of the same known split, and moving one at a time without the sweep's other cross-graph state would imply a fix that is not there.
- **A refused env value read back as "unset".** `QUOTA_PICKUP_THRESHOLD_PERCENT` is now held verbatim in config and validated in the resolver, as `agentModel` already is, so the screen shows what was typed beside the default actually in force.

The third I disagree with, and have answered in the code rather than changed: the reset-less staleness bound clocks from `observedAt`, so an event seen four hours into a five-hour window holds pickup roughly four hours past the real reset. That is real, and the comment understated it — it now states the worst case plainly. But tightening the bound is the more expensive mistake, not the cheaper one: reopening early does not cost "a probe", it costs a *ticket* one of its three attempts, because a rejected pass exits in ~2s (#165's finding 5) and is recorded as a failed attempt like any other. A fleet probing every half hour through a long wall would burn three real tickets to a human before the window reopened. The bound also only ever binds `allowed_warning` events — a rejection carries `resetsAt` and is released exactly at it — and those are the case where the fleet is stopping short on purpose and loses nothing by stopping a while longer.

## Merged `origin/main`

#180 (execution lanes) landed mid-branch and restructured the settings layer underneath it — for the better. `SettingSpec` was already generalised past model tiers, the lane field already had its own resolver / view model / panel beside the shared registry, and `fleet-bits` already carried the `ChipRadio` and `fallbackNote` atoms this branch had extracted by hand. The merge drops all of that duplication: the discriminated `detail` union on `SettingFieldView` is gone, the hand-rolled panel plumbing is gone, and the quota field is re-applied in the lane field's shape. Net effect on `main` is now additive on the settings layer rather than a reshaping of it.

## Checks

`pnpm test` — 66 files, 1259 tests, all passing. `pnpm build` compiles. `npx eslint` clean on every changed file (the repo's lint baseline on `main` is untouched).

Acceptance criteria are table-tested: the threshold boundary and the rejection override in `quota-gate.test.ts`, and through the reducer in `decide.test.ts` — in-flight work untouched, triage held with claims, the once-per-transition announcement, silence when nothing is eligible, and both precedence edges. The banner and digest lines are pinned in `fleet-view.test.ts` and `digest.test.ts`; the panel's reading in `quota-gate-settings-render.test.tsx`.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/quota-gate-settings-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/quota-gate-settings.tsx` |

lennons301-reviewer (approved):
Reviewed with fresh eyes against issue #171, the repo's CLAUDE.md conventions and `docs/agents/review-gates.yaml`. **Approving** — the PR is gated (`human-signoff`, auto-merge off), so this satisfies branch protection and a human merges after looking.

## Objective signals (no CI configured on this branch, so run locally)

- `pnpm test --run` — 66 files, 1259 tests, all pass.
- `npx tsc --noEmit` — clean.
- `npx eslint` on all 23 changed files — clean (the `main` lint baseline is untouched; no baseline file was edited).
- `pnpm build` — compiles, all routes emit.

## Acceptance criteria, each traced to code and a test

| Criterion | Where |
| --- | --- |
| Above the threshold, no new tickets claimed | `evaluateQuotaGate` + `decideNext`'s gate at the pickup point; boundary table in `quota-gate.test.ts` and `decide.test.ts` (`89.9/90/94` against `50/90/95`) |
| An account-wide `rejected` also stops pickup | Status check ahead of utilization; tested at threshold 100 with 3% utilization |
| In-flight work unaffected | Gate sits below every in-flight decision; `keeps driving in-flight work` asserts the exact action sequence `postVerdict, startReview, armAutoMerge, exhaust, pausePickup` |
| Parked runs still resume | Resume/follow-up delivery is decided ~350 lines above the gate; covered by the same test's `startReview` on a parked run |
| Settable in the UI, falls through when unset | `resolveQuotaThreshold` + `QuotaGatePanel`; three-layer fall-through pinned in `settings-resolver.test.ts`, the panel's reading in `quota-gate-settings-render.test.tsx` |
| Announced once per transition | `quotaGateAnnouncement` flag; `announces… once` / `does not re-announce` |
| Discord only when the fleet is stalled | Notify emitted from past the eligibility filter; `says nothing at all when there is nothing eligible` |
| Kill switch and per-project toggle take precedence | Gate placed after the switch; `names the kill switch ahead of the gate`, `leaves a project's own holds saying their own piece` |
| Expressed in the reducer, table-tested | `decide.test.ts` block, ~240 lines of tables |

## What I checked beyond the ticket

- **One gate, two callers.** `decideNext` and `buildFleetView` both call the same pure `evaluateQuotaGate` with the same threshold resolved from the same row + config, so the banner cannot say "claiming" while the reducer refuses. That is the #148 property, and it holds. `PickupPause["reason"]` gained `quota-gate` and both keyed maps in `fleet-bits.tsx` plus the digest switch were extended — the union is exhaustive, confirmed by the typecheck.
- **The three safety rules actually hold in code**, not just in comments: `observation === null` returns an open gate before anything else; `utilization` is compared only when non-null, so absent never reads as 0; `quotaObservationIsSpent` reopens at the stated reset. Only the literal `"rejected"` gates on status — an unrecognised status decides nothing, matching #167's verbatim-enum rule.
- **Module-graph discipline (#159).** The announce-once flag is on `processSingleton`, correctly — a webhook-triggered sweep runs on the app-router graph, where a module-level flag would be freshly false and would double-post the embed. Usage is the mutable-container form the helper requires. Leaving `saturationAnnounced`/`dailyCapAnnouncedDay` alone is the right call and the reason is written down.
- **No false watchdog interaction.** `pickupPausedWithFreeSlot` matches only `reason === "no-slots"`, so a quota pause cannot trip the #126 pickup-wedged card. Checked, because that was the obvious way this could have gone wrong.
- **Clearing from the gate, not from the pause.** The sweep re-evaluates the gate on the same snapshot inputs (identical result by construction) rather than inferring from actions — necessary, because the pause is only emitted when eligible work exists. The reasoning is in the comment and is correct.
- **Docs.** `.env.example`, the runbook (pause-pickup section + env table) and CLAUDE.md all updated in the house style; the runbook's `paused` dot wording was corrected for the second reason now sharing it.
- **Gated paths confirmed present**: `src/lib/orchestrator/autonomy/**` (autonomy-control), `src/lib/config.ts` (credentials), `src/components/**` (estate visual-UI default). The gate label is correct.

## Non-blocking notes (no change requested)

1. **The reset-less staleness bound is the one judgement a human should confirm.** An `allowed_warning` at or above threshold that carried no `resetsAt` — and #167 records that `resetsAt` is *frequently absent* — holds pickup for 5h from observation, potentially ~4h past the real reset, on an otherwise idle fleet (idle meaning no pass making API calls, hence no fresh observation to reopen it). The PR argues waiting is the cheaper mistake because a probe costs a ticket one of its three attempts (#165's finding 5), which I find persuasive, and the constant is documented in place with the worst case stated plainly. Flagging it as the availability trade-off worth a nod on merge, not as a defect.
2. `QuotaGate.observedAt` is carried on the returned struct but no consumer reads it (the tile reads the observation directly). One unused field — trivially either used or dropped later.
3. `QuotaGate`'s doc says the fields are "nulls when there was none"; they are also nulls when the observation was *spent*. One word in a repo that otherwise keeps comments exact.
4. `settings-overrides.tsx` now has two `import type … from "@/lib/settings-resolver"` lines that could be one.
5. The "window" phrasing (`the quota window` vs `the ${describeRateLimitType(...)}`) is built twice — once in `fleet-view.ts`, once in `notifications.ts`. Consistent with this repo's pattern of each surface owning its own voice, so I'd leave it; noting it as the only duplication in the diff.
6. When both the override and `QUOTA_PICKUP_THRESHOLD_PERCENT` are unset the panel reads `environment` / `from QUOTA_PICKUP_THRESHOLD_PERCENT unset` while holding at the built-in 90, and a *refused* env value shows as `= 93` beside "holds pickup at 90%" without saying it was refused. Both are honest and deliberate (and tested), but a reader has to infer the third layer from the numbers.

Issue #176, opened 2026-08-26
Cross-lane cost routing: pick the cheapest capable lane, and fail over instead of pausing


lennons301:
## Parent

#164

## What to build

After #172, #174 and #175, the fleet can run on any of several lanes — but **nothing ever
chooses between them on cost, and nothing ever moves work between them.** Two separate
mechanisms exist and neither crosses a lane boundary:

- The **degrade ladder** (#170) steps `heavy → standard → light` *within one lane*. It is a
  tier mechanism, not a provider mechanism.
- **Overflow** (#173) crosses lanes, but only for interactive work, only when walled, and
  only onto a designated overflow lane.

So a walled autonomous run **pauses for hours** (#168/#169) even when a lane costing a
fraction of the primary is sitting idle and available, and an implement pass runs on
whatever the primary lane is regardless of whether a cheaper lane could do that pass kind
perfectly well. On a personally-funded fleet where cost is the primary driver, that is the
largest remaining lever in this milestone.

This ticket adds lane selection to the reducer. Two halves, which can land in either order
but share one selection function:

**Cost-first selection.** Lane choice becomes part of resolving a pass, not a fixed
`primary` setting: given a pass kind, its configured tier, and the set of available lanes,
pick the cheapest lane that can serve it. Pass kinds get a **minimum lane** — an implement
pass on a hard ticket may be pinned to a capable lane while triage and review are free to
run on the cheapest thing available. Selection stays a pure function so it is table-testable
with no provider involved, and the ledger records the lane so the cost case is auditable
after the fact.

**Wall-triggered failover.** When the current lane is walled and the bottom of its tier
ladder is reached, the reducer prefers **moving to another available lane over pausing**,
subject to every guard already in place: an autonomous run may only move onto a metered lane
within the real-money cap and confirm gate from #174, and a lane with no quota telemetry is
gated by spend rather than utilization (#175). Pausing (#168/#169) becomes the fallback when
no lane is both available and permitted — not the first response. Lane moves reuse the
lossless transcript restore from #169, so a failover costs no attempt and no interruption,
and is bounded by the same quota-pause counter so a run cannot thrash across lanes forever.

**The escape hatch stays.** Pinning the fleet to one lane must remain expressible in
settings — cost routing is a default, not something the owner has to fight. Likewise the
kill switch and per-project autonomy toggle still stop everything; routing never overrides a
human stop.

Explicitly **not** in scope: quality-aware routing (choosing a lane by how well it did on
[...]

PR #183, opened 2026-09-02
Billing-kind money guardrails: real-money cap, confirm-once gate, and a metered primary lane

### LINKED ISSUE #174: Billing-kind money guardrails: real-money cap, confirm-once gate, and a metered primary lane
## Parent

#164

## What to build

#164 specified the money guards — confirm-once-then-cap, the daily real-money cap, the
metered-vs-subscription split, "autonomous never spends cash unknowingly" — as part of
interactive overflow. That attaches them to a **transition** ("we overflowed") when the
thing that actually costs money is a **lane's billing kind**. Two consequences, both worth
fixing before either is built:

1. **The cheapest configuration of this platform slips every guard.** It is not
   "subscription primary, metered overflow"; it is metered-**primary** — drop the
   subscription and run the whole fleet per-token on an open-weights lane. Read literally,
   autonomous passes there spend real money on every turn with no confirmation, outside the
   real-money cap, and absent from the dashboard's spend split, because nothing is
   *overflowing*.
2. **It puts cost work behind rate-limit work.** Overflow depends on pause/resume and the
   admission gate; metered-primary depends on neither. Gating the fleet's largest cost lever
   behind the whole quota chain is the wrong order now that the fleet is personally funded.

So the guards land **here**, keyed off the resolved lane's billing kind, and interactive
overflow (#173) becomes their first consumer rather than their owner. A metered lane is a
metered lane whether it is primary, an overflow target, or reached by failover (#176).

**Autonomous work on a metered primary is explicitly allowed** — that is the entire point of
the cost flip. It is bounded by the real-money cap rather than exempted from it. The existing
$500/day autonomous cap measures quota-funded work; a metered primary needs a separately
configurable cash ceiling that is meaningfully smaller. The confirm-once-per-day gate applies
at **fleet level** here, not per interactive session — nobody is sitting at the keyboard when
an autonomous pass crosses into billing.

The daily-cap pause reason already exists in the reducer; this is a second cap feeding the
same mechanism, not a new pause path.

**Out of scope, stays in #173:** treating an active Anthropic *overage* as a paid lane. That
rule reads `overageStatus` off the rate-limit event and so genuinely depends on the telemetry
chain; it is a billing-kind determination made from quota data, not a guard. #173 also keeps
the interactive-overflow trigger itself and the at-the-keyboard confirmation UX.

## Acceptance criteria

- [ ] Money guards key off a resolved lane's billing kind, not off whether a pass overflowed
- [ ] A metered lane selected as primary is subject to the real-money daily cap
- [ ] Autonomous passes are permitted on a metered primary lane — not treated as overflow and blocked
- [ ] The first metered spend of a local day requires one confirmation at fleet level
- [ ] Further metered spend that day proceeds automatically until the cap
- [ ] Reaching the real-money cap pauses new claims via the existing daily-cap pause reason, announced once per transition
- [ ] Subscription-lane work remains exempt from the real-money cap — the cap measures money, not quota
- [ ] The dashboard shows metered spend separately from subscription usage, correct when the primary lane is metered
- [ ] The real-money cap is settable in the settings UI and falls through to its env default when unset
- [ ] Switching the primary lane between billing kinds takes effect on the next sweep without a restart
- [ ] The run ledger records the lane and billing kind each run used
- [ ] Reducer behaviour for a metered primary is covered by table tests with no Docker or network

## Blocked by

- #172

lennons301:
Closes #174

The money guards key off a resolved lane's **billing kind**, never off whether a pass overflowed. That is the ticket's point: the cheapest configuration of this platform is not "subscription primary, metered overflow" but metered-**primary**, and a guard hung off the overflow transition would let exactly that configuration spend real money on every turn — unconfirmed, uncapped, and absent from the dashboard's split. A metered lane is now guarded whether it is primary, an overflow target (#173) or reached by failover (#176), and **autonomous work on a metered primary is allowed**: bounded by a cash cap rather than exempted from one.

## What landed

**One pure core.** `src/lib/lanes/money.ts` answers two questions — what is the cap in force, and what do the guards make of today — and both consumers (the autonomy reducer and the fleet read model) evaluate the *same* function over the same facts, so the sweep and the dashboard cannot disagree about whether the fleet is held. One impure reader beside it (`money-state.ts`) gathers "which lane, at what cap, spent how much" for all three callers.

**The cap** is the lower of the operator's dial (`METERED_DAILY_CAP_USD`, $20, settable on the settings screen up to a hard $100 code ceiling refused by name) and the lane's own declared `caps.daily_budget_usd` — neither can widen the other, and which one binds is reported. Reaching it pauses new claims through the reducer's **existing** `daily-cap` reason with a detail naming which cap: two caps, one mechanism.

**The confirm-once gate** is fleet-level and per local day, with its own pause reason — the cap lifts itself at midnight, this one waits for a press, and telling an operator the wrong one sends them to a control that cannot help. On a capped day the cap wins, because confirming would start nothing. Both announce once per local day in Discord, so a metered fleet held for a confirmation pings rather than sitting silent.

**Real money is a per-day ledger** (`metered_spend`), written as each turn's cost lands, for a pass whose recorded lane bills per token. A task's `totalCostUsd` is a running total carrying no day, so attributing it by creation under-counts a long session (spending past the cap) and by last-update double-counts it (holding the fleet over money it never spent today); the *increment* is the thing that is unambiguous. `updateTask` is the single funnel every task-cost write goes through, so the booking lives there — and each turn's dollars go to the lane that actually spent them.

**The ledger records who paid**: `runs.laneBilling` beside the existing `runs.lane`, plus `tasks.lane`/`tasks.laneBilling` (a task is the unit money is spent by — triage owns no run, interactive never has one). Billing is stored, not looked up, because `lanes.yaml` changes under a deployment and a retired lane would otherwise rewrite the billing history of every past run.

**Surfaces**: a second gauge on the dashboard's pulse strip, a second Spend line in the digest (printed only when there was cash), two `PickupPause` reasons, two needs-you cards each carrying the remedy that applies, and a *Real money* panel on the settings screen where confirming takes the same deliberate strip that arming a project does. The two spend figures overlap by construction and are never added.

Switching the primary lane between billing kinds takes effect at the next sweep: the guards read the lane and the settings row fresh every tick.

## Judgement calls a reviewer should weigh

1. **The guards hold new autonomous pickup, not interactive dispatch.** A self-review read "the first metered spend of a local day requires one confirmation" as also covering an interactive task dispatched from the UI or Discord on a metered lane. I've deliberately left that ungated: an interactive dispatch is attended (a human is pressing the button), the ticket puts "the at-the-keyboard confirmation UX" explicitly in #173, and the only thing I could ship here without inventing that UX is a hard task failure — worse than the confirm dialog #173 will build. What interactive work *does* do is **count** against the cap, so an attended session's cash still stops autonomous pickup. lanes.yaml, `.env.example` and the runbook now all say exactly this rather than over-claiming.

2. **Interactive metered spend counts toward the cap at all.** The existing $500 cap exempts interactive work by construction; this one deliberately does not, because it measures money rather than quota and a chat session on a metered lane charges the same card. The consequence — a heavy interactive day can pause autonomous pickup — reads correct for a cash ceiling, but it is a choice.

3. **Scope beyond the checklist**, all small: the `$100` code ceiling on the cap (consistent with #166's rule that a UI press may not widen a safety ceiling, but it means a deployment wanting more must restart); the lane's declared cap binding the dial down (implied by the driver note, not stated in the criteria); `tasks.lane`/`tasks.laneBilling` and the `metered_spend` table, which the criteria didn't ask for but the real-money sum genuinely needs; and the Discord embeds and digest lines.

The driver note's `lanes.yaml` fix is done on both metered lanes: enforcement is #174's, and the comment says how.

## Checks

`vitest run` — 1265 passing, including 15 reducer table tests for the guards (no Docker, no network), the pure `money.ts` table, the ledger's day-splitting, the fleet-view split and the settings endpoint against a from-migrations DB. `tsc --noEmit` clean, `eslint` clean on every changed file, `pnpm build` succeeds. Two migrations (0025, 0026), journal `when` bumped a day past the prior entry each time.

## Merged with main

#181 (pause on a rate limit) and #182 (the quota admission gate) landed while this was in review; both touch the same seams and are merged in here (merge only, no rebase). Everything is additive, and the resolutions worth a reviewer's eye:

- **Pickup gate order** in `decideNext` is now: daily cap → real-money cap → kill switch → spend confirmation → quota gate → slots. The two money holds deliberately straddle the switch. The cap goes above it, where the daily cap already sat, because its announcement is news whether or not a human has stopped the fleet. The confirmation goes below it, where #171 put the quota gate, because asking for a press under a hold a human engaged points them at a control that would change nothing.
- **`PickupPause`** now carries five reasons, ranked on the dashboard by what a reader must act on: boot master, switch, daily cap, cash cap, unconfirmed day, quota gate. #171's stated precedence (below the switch, below the cap, above slots) is unchanged.
- **One settings read** per sweep tick and per view build serves the kill switch, the quota threshold and the money guards — the raw merge had left two reads of the same row side by side, which can describe two instants.
- **Migrations renumbered**: main took `0025`, so this ticket's two collapse into a single `0026` generated against main's snapshot, journal `when` a day past it.

`vitest run` — 1362 passing after the merge; `tsc --noEmit`, `eslint` on changed files and `pnpm build` all clean.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/metered-spend-settings.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0026_metered_money_guards.sql` |
| persistence | `drizzle/**` | `drizzle/0026_metered_money_guards.sql` |

lennons301-reviewer (approved):
Approving. Gated PR (`human-signoff`, auto-merge off) — this approval satisfies branch protection; a human merges.

## What I verified

No CI checks are reported on this branch, so I ran the objective signals locally in the worktree:

- `pnpm vitest run` — **72 files, 1362 tests, all passing** (matches the PR body's post-merge figure)
- `npx tsc --noEmit` — clean
- `npx eslint` on every changed `.ts`/`.tsx` — clean (no reliance on the known `main` lint baseline)

Against the ticket's acceptance criteria, each traced to code and a test:

| Criterion | Where |
|---|---|
| Guards key off billing kind, not overflow | `evaluateMeteredSpend` takes `billing`, never a transition; `primaryLaneOf` derives it from `describeLanes`, and `resolveLane` picks the same lane for every pass kind, so the guard's prediction and the pass's reality can't diverge |
| Metered primary subject to the cap | `readMoneyGuards` → `decideNext` cap block; `decide.test.ts` "pauses through the existing daily-cap reason…" |
| Autonomous passes permitted on a metered primary | "permits autonomous claims on a metered primary lane once the day is confirmed" |
| Confirm-once at fleet level, then automatic | `settings.meteredSpendConfirmedAt` + `sameLocalDay`; "treats yesterday's confirmation as no confirmation", "lets further metered spend proceed automatically" |
| Cap pauses via the existing reason, announced once | `pausePickup { reason: "daily-cap", detail }`; announcement keyed on the local day, and — importantly — placed **after** `if (eligible.length === 0) return actions`, so the ping means "held with work armed", the same discipline #171's does |
| Subscription work exempt | `metered === false` ⇒ `hold === null`; "exempts subscription work from the real-money cap entirely" |
| Dashboard shows the split | second gauge in `pulse-strip.tsx`, `FleetView.spend.metered`, never summed with `todayUsd` |
| Settable, falls through to env | `SETTINGS_FIELDS.meteredDailyCapUsd` + `resolveMeteredCapSetting`; route tests cover store / clear / refuse-past-ceiling / refuse-by-name |
| Lane switch effective next sweep, no restart | `readMoneyGuards` re-reads the settings row each tick and each view build; `getConfig()` memoisation is only the env floor |
| Ledger records lane + billing | `runs.laneBilling`, `tasks.lane`/`tasks.laneBilling`, written at pass start and re-written per follow-up turn |
| Table tests, no Docker/network | 15 reducer cases + 14 pure `money.ts` cases, all in-process |

Things I checked rather than took on trust:

- **The ledger is genuinely increment-based.** Both cost write sites in `turn-manager.ts` (l.508 first turn, l.1034 running total) pass a *total*, and `recordMeteredSpend` books `new − old`, so no path double-books; a decrease books nothing, which under-counts rather than refunds. Day-splitting, idempotency and past-day readability are each pinned in `spend.test.ts`.
- **No secret crosses the new route** — it serves numbers, a lane id, a label and a billing kind, and there's an explicit test for it. Consistent with the #172 rule the project route once broke.
- **Migration hygiene** — `0026` journal `when` is exactly +1 day past `0025`, and the endpoint test runs against a from-migrations DB.
- **The driver note's `lanes.yaml` correction is done** on both metered lanes; no `#175` cross-reference survives.
- **Precedence** matches what CLAUDE.md now states, and the money blocks sit past the in-flight handling, so a hold never stops work already running ("keeps driving in-flight work while a money guard holds pickup").

## Findings — none blocking

1. **The digest's confirmation line is evaluated at the wrong instant for what it claims.** `digest-schedule.ts` builds the view at `viewAt = window.end − 1ms` (the covered day's last instant), so `evaluateMeteredSpend` compares `confirmedAt` against *yesterday*. The `metered-unconfirmed` line says "Held right now", and the comment above it says the confirmation "is taken off the live settings row, so it says how the fleet stands as the digest is written" — neither is true in the two mixed cases: confirmed-yesterday-not-yet-today reads quiet while pickup is actually held, and confirmed-early-this-morning reads held when it isn't. This is the same class as the documented kill-switch tense caveat ("a switch lifted before the morning send leaves the covered day reading quiet"), and the added test only exercises confirmation *on* the covered day, so it isn't caught. Cheapest fix is to correct the comment/copy; the truthful one is to judge the confirmation against real `now`.

2. **The funnel that feeds the ledger is untested.** Everything downstream is tested purely, but the two facts the cap's accuracy actually rests on — that `tasks.laneBilling` is written before any cost write, and that `updateTask` books the delta only when `before?.billing === "metered"` — have no test. `turn-manager.ts` has no unit tests in this repo, so this is the baseline rather than a regression, but a slip here is money-shaped in both directions (silent over-spend, or a phantom hold). Worth a follow-up if any of that becomes extractable.

3. **Scope of "the first metered spend of a local day requires one confirmation".** Implemented as "the first *autonomous pickup* of the day requires it": an attended interactive dispatch can spend the day's first cash unconfirmed, and it counts toward the cap. I think this is the right reading — the ticket itself grounds the gate in "nobody is sitting at the keyboard", and puts the at-the-keyboard UX in #173 — and it is now stated plainly in CLAUDE.md, `lanes.yaml`, `.env.example` and the runbook rather than over-claimed. Flagging it as the one substantive judgement call for the human signer to agree with, since it is the difference between "never spends cash unknowingly" and "never spends cash *unattended* unknowingly".

4. **Two small smells, both suggestions.** The display predicate `metered.active || metered.todayUsd > 0` is written twice (`pulse-strip.tsx`, `digest.ts`) and would sit naturally as a field on `MeteredSpendView`. And local-day arithmetic now has three helpers across two modules — `sameLocalDay` is expressible as `localDayKey(a) === localDayKey(b)`, which would leave one notion of a local day rather than two spellings of it.

5. **Panel independence.** The Real money panel fetches its own endpoint, so switching the primary lane in the lane panel leaves it naming the previous lane until reload — a deliberate trade (the route docstring argues it, and raising the cap while held does need the hold recomputed in one answer), but a small divergence from #171's stated reason for folding the quota panel into the shared settings state.

Also noted for the human, not a defect: the cash cap sits **above** the kill switch in `decideNext` and **below** it in `buildFleetView`'s `pickupPaused` ranking. That mirrors the existing daily-cap precedent exactly and is documented in CLAUDE.md, so the two orders differ by design, not by accident.

The design reasoning is unusually well recorded at the sites that need it, the pure/impure split is real (one `evaluateMeteredSpend` for the reducer and the tile, one `readMoneyGuards` for all three impure readers), and the fail-safe directions are the right way round throughout: an unresolvable lane holds nothing, an over-spent day reads as zero remaining rather than a debt, a mistyped env cap keeps the default rather than becoming 0 or NaN, and a lane's declared cap can only bind the dial down.

Issue #170, opened 2026-08-26
Tier degrade ladder: step down instead of pausing when a tier allowance is exhausted


lennons301:
## Parent

#164

## What to build

Losing the top tier's weekly allowance **steps the fleet down a tier and carries on**
rather than stopping it.

This is the point where the two halves of #164 turn out to be one mechanism. Because the
rate-limit event distinguishes a tier-scoped weekly allowance from the account-wide
five-hour window, exhausting the heavy tier does not mean no work can run — it means
heavy work cannot run. So a tier-scoped rejection steps the run down the ladder
`heavy → standard → light` and retries, and only an account-wide rejection, or a rejection
at the bottom of the ladder, actually pauses the run.

The run records the tier it stepped down to, so the ledger stays interpretable: spend can
be read against the tier that actually ran, not the tier that was requested.

## Acceptance criteria

- [ ] A tier-scoped weekly rejection steps the run down one tier and retries rather than pausing
- [ ] An account-wide rejection pauses the run rather than degrading it
- [ ] A rejection at the bottom of the ladder pauses the run
- [ ] The tier that actually ran is recorded on the run and visible in the ledger
- [ ] The dashboard shows when a run is running below its configured tier
- [ ] The ladder decision is expressed in the autonomy reducer and covered by table tests spanning each limit type

## Blocked by

- #166
- #168

lennons301:
🤖 Attempt 1/3 starting.

lennons301:
🔧 Repair pass starting — merging main into PR #185 to resolve conflicts.

lennons301:
🔧 Repair pass starting — merging main into PR #185 to resolve conflicts.

PR #184, opened 2026-09-02
OpenRouter lane: prove the third-party-provider path end to end on a cheap open-weights model

### LINKED ISSUE #175: OpenRouter lane: prove the third-party-provider path end to end on a cheap open-weights model

lennons301:
Closes #175

Ships and verifies a second, real execution lane against OpenRouter, and turns the two assumptions that did **not** survive contact with it into mechanisms. Full measurements are recorded on the issue; the short version:

**The Anthropic-skin finding is confirmed.** With `ANTHROPIC_BASE_URL=https://openrouter.ai/api` and `ANTHROPIC_AUTH_TOKEN=<OpenRouter key>`, the real Claude Code harness ran a pass with native tool use, extended thinking, `--resume` and prompt caching all intact, and did the work it was asked to do. **No proxy, no second adapter, and no provider-specific branching in the adapter seam** — `buildExecEnv` and `buildCommand` are unchanged.

**Cost reporting is fiction off an Anthropic-direct endpoint.** The CLI reported `total_cost_usd: 0.194985` for a turn on a *free* model: `38387 × $5/Mtok + 122 × $25/Mtok` exactly, Anthropic list rates applied to a model the endpoint never billed at them. It says so itself — `costBasis: "unknown"` where the control run against Anthropic reads `"list"` and reproduces its own figure to the cent. So a lane declares `prices` per tier and `lane-cost.ts` charges every budget from them; an Anthropic-direct lane keeps the CLI's number, which is correct there and only there.

**No quota telemetry at all**, confirming #165's finding 6 against a second provider: no `anthropic-ratelimit-*` header, no `rate_limit_event` on a full harness turn, while the subscription lane emitted one from the same harness in the same minute. `quota_state` is keyed by lane id.

**Prompt caching works; cache writes are invisible.** Two identical passes: 21261 input / 0 cached, then 210 input / 21051 cached — 8.8× cheaper. `cache_creation_input_tokens` is never a count (`null` raw, `0` through the harness), so cache writes arrive as ordinary input tokens — which is why an unpriced cache column is charged at the input rate rather than at zero.

## Two defects this found beyond the ticket's three uncertainties

**`--max-budget-usd` was still denominated in the CLI's currency.** The CLI enforces it against its own cost figure, so a $20 attempt budget on the GLM lane would have stopped a turn at roughly $0.30 of real spend — mid-work and *invisibly*, because a budget stop is not `error_max_turns`, so the orchestrator would have parked a truncated pass as finished. Exactly the "cheap lane that fails every ticket" the ticket warns about. A lane with prices is now given no ceiling the harness would misapply; `--max-turns` and the fleet's own (now-correct) accounting still bound it. The subscription lane keeps the flag and is untouched.

**Per-lane quota stopped being forward-looking mid-PR.** #182 landed the admission gate on `main` while this branch was in progress. Read fleet-wide, a subscription rejection would have closed that gate over a metered lane that cannot be rate-limited at all, holding every pickup on a fleet that was — on that lane — entirely free to work. The sweep now hands the gate the primary lane's observation via `currentPrimaryLane`, one helper it shares with the dashboard so the two can never name different lanes. The gate's own "no observation is not a closed gate" rule does the rest, so this is the whole fix rather than a new branch in it.

## Notes for the reviewer

**Two acceptance criteria are unmet, and both need you rather than more code:**

- [ ] *A full implement pass runs to completion on the lane in a real container, producing a branch and a PR*
- [ ] *A review pass and a triage pass each run on the lane*

Two independent blockers. The **OpenRouter account has never purchased credits** — every paid model, GLM 5.3 Flash included, returns `HTTP 402 billing_error` — so all of the above was measured on a free model. And the worktree environment has **no Docker daemon** (WSL2, Docker Desktop integration off), so no container could be provisioned regardless. Consequently the three GLM slugs are verified only against OpenRouter's model listing, which is also where their prices come from; `lanes.yaml`'s note above the lane says that rather than implying more. To finish: add credits, run `node scripts/openrouter-lane-check.mjs --model z-ai/glm-5.3-flash`, switch **Settings → Execution lane → `openrouter-glm`** and arm a small ticket. Every other criterion is met and covered by tests.

**The "degrade ladder" in criterion 7 is #170, still open.** It inherits per-lane keying by construction — there is one keyed source of an observation and its readers pass a lane — so there was nothing further to build for it here.

**Scope I added deliberately**, none of it asked for by name:

- Prices on the pre-existing `openrouter` lane as well as the new one. It is equally a third-party lane, so its spend would otherwise have been the same fiction.
- The `openrouter-glm` lane is a *second* lane rather than a retune of the first. Its only differences from its neighbour are three model slugs and three rows of prices — same credential, same endpoint, no code — which is the ticket's own thesis stated as a fact about the file.
- A per-turn feed line when the lane's charge differs from the harness's claim. The feed's existing "Turn complete (cost: …)" line carries the harness figure, and leaving it as the only thing an operator sees would make a working lane look like one that is bankrupting them.
- `scripts/openrouter-lane-check.mjs`, mirroring #165's `rate-limit-stub.mjs`. #164's single-adapter design rests on an assumption about live providers, and the first check of it found two parts false — so the check is a script, not a paragraph in this PR. Harness-free on purpose, so a surprising result points at the endpoint or the harness rather than at both.

**Self-review acted on** (`/code-review` against `origin/main`, both axes): the budget-flag defect above; a `usage`-object fallback in `readTurnUsage` that would have undercharged a priced lane ~90× while looking correct (this branch's own fixture proves `usage` is the last API iteration, not the turn, so `modelUsage` is now the only source); a feed line that asserted a vendor's pricing behaviour from orchestration that names no vendor; a duplicated billing union; an over-claiming comment in `lanes.yaml`; and an unread `charge` field on `runTurn`'s return type.

**Not fixed, deliberately:** `pnpm lint` still reports the known baseline — 9 `require()` errors in `output-parser.test.ts` on lines this PR did not touch (present on `main`), and one `set-state-in-effect` in `preview-pane.tsx`, untouched here. Every file this PR changes is otherwise lint-clean. Test suite: 1353 passing.

**One thing worth its own ticket, not invented here:** attempt cost accumulates in one column (`runs.totalCostUsd`), so a lane switched *mid-run* would mix the harness's currency with lane prices. Harmless today — a switch is a deliberate press between attempts — but #174's real-money cap and #176's cross-lane routing both make mid-run switching routine. Flagged rather than guessed at with a per-lane cost ledger.

## Review round 2 — both blocking findings addressed

**1. The lane's prices did not apply in the default configuration.** Correct, and worse than a gap: `tier` is null in the *unset* state (no `AGENT_MODEL`, no stored tier), which is what one press of the lane chip leaves behind — so on the shipped `lanes.yaml` the fleet charged $0.194985 for a ~$0.0029 turn *and* re-added `--max-budget-usd 20` against that fiction. Two halves, two answers, as the review suggested they might need:

- **The flag** is keyed on the lane *definition* (`ResolvedLane.declaresPrices`), never on the resolved tier. "Does the CLI price this provider?" is a fact about the endpoint, true before any tier resolves, so the pinned-model case keeps the guard too.
- **The tier** now falls back to the lane's own `standard` on a priced lane (`laneFallbackTier`). "Let the harness decide" cannot mean anything coherent there: the identifier it would pick belongs to Anthropic's catalogue, so the endpoint either refuses it or quietly serves a Claude model at a price this lane's table does not hold — and the fleet charges the fiction for it. On an Anthropic-direct lane, unset still means exactly what it always did.

Pinned by four tests, including through the real settings route: switching to `openrouter-glm` and nothing else now resolves a slug, a price, and no ceiling.

**2. `drizzle/meta/0026_snapshot.json` was missing.** Written, and verified the way the review found it: `drizzle-kit generate` now reports "No schema changes, nothing to migrate" with no rename prompt. CLAUDE.md's migration section says why the snapshot is not optional for a hand-written migration — `migrate()` never reads one, so nothing local fails, but the next author's `generate` re-asks and either answer is a boot failure.

**Stray blank line** in `rows.ts` removed.

### Three more from a second `/code-review` pass over the result

- A priced lane with a pinned model reported `basis: "harness"` — the basis documented as "the lane declares no prices" — so the CLI's fiction was labelled a list price and `noteLaneCharge` stayed silent. It now reports `harness-unpriced`, which already means "prices that could not be applied".
- The tier rows on the settings screen would have said "no `--model` — the account default" over a priced lane. `laneFallbackTier` is threaded into `describeModelTierSettings`, so the row names what the pass runs. One function, both readers — the rule is not written twice.
- **The dropped `--max-budget-usd` removes the only in-turn ceiling a review or triage pass had on a priced lane.** `attemptExhaustion`, which the comment cited, runs for implement/repair only. Corrected in the code and stated here rather than papered over: what remains is `--max-turns` and the daily cap (which counts every autonomous pass kind, now at real prices). Still the right trade — enforced against the CLI's figure, a $5 review budget cuts the pass off at roughly $0.08 of real spend, mid-work and invisibly, which is a reviewer that silently reads half a PR. Converting the ceiling into the CLI's currency would mean modelling its pricing, and a wrong ratio is that same invisible truncation back again.

**Where I disagree with the automated spec pass**, for the record: it reads AC7's "a lane with no quota signal is gated by spend" as unmet because `caps.daily_budget_usd` is declared but not enforced. That enforcement is #174's whole ticket (`lane-config.ts` says so at the field), and #172 shipped the declaration deliberately so #174 is a policy change rather than a config-shape change. It also called the priced-lane tier fallback scope creep; it is the fix for this PR's blocking review finding, and it does re-point the pre-existing `openrouter` lane — which is the intended consequence, since that lane's unset state had the same defect.

Re-verified after the changes: `tsc --noEmit` clean, 1359 tests passing, every changed file lint-clean against the known baseline.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/quota-tile-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/execution-lane-settings.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/quota-tile.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0026_quota_state_per_lane.sql` |
| persistence | `drizzle/**` | `drizzle/0026_quota_state_per_lane.sql` |

lennons301-reviewer (changes requested):
Request changes. The measurement work here is excellent and the shape of both fixes is right — but the priced-lane mechanism does not engage in the default configuration, which is the one state a fresh deployment is in, and the migration is missing its drizzle snapshot.

## What I verified

- Typecheck clean (`npx tsc --noEmit`); full suite green: 73 files / 1353 tests.
- Lint on every changed file: only the known baseline (9 `require()` errors in `output-parser.test.ts`, on lines this PR did not touch — confirmed identical on `origin/main`). Matches the PR body.
- **Migration 0026 applied for real**: built a DB from journal entries 0000–0025, inserted the pre-existing `fleet` row, then applied `0026_quota_state_per_lane.sql`. The row survives, re-keyed `claude-subscription`, and the primary key is preserved (`"lane" text PRIMARY KEY NOT NULL`). Both `src/db/index.ts` and `src/test/create-test-db.ts` go through `migrate()`, so the rename is exercised by every test file. The re-attribution comment in the SQL is a fair statement of fact.
- Secrets: `lanes.yaml` carries variable *names* only (and its own test asserts that), `FleetLaneRow` deliberately takes three fields and no `auth`, and `LaneView` exposes `authEnvVars` plus prices. No lane credential in the DB or on a route.
- The cost substitution really is once: every consumer — `tasks.totalCostUsd` (turn-manager:500), `attemptExhaustion` (:544, :1086), `runs.totalCostUsd` (:1070) — reads `turnResult.costUsd`, and `spend.ts` sums those two columns. Nothing reads `total_cost_usd` behind `runTurn`'s back.

## 1. Blocking — the lane's prices do not apply in the default state, so both new mechanisms are off

`src/lib/lanes/resolve.ts:278`

```ts
prices: tier !== null ? (lane.prices?.[tier] ?? null) : null,
```

`tier` is null whenever no tier resolves, and that is not only the pinned-model escape hatch the comment above it describes — it is also the **unset** case. `AGENT_MODEL` unset (`config.ts:176` → `null`) with no UI override gives `resolveModelTierField` `envTier === null`, hence `{ tier: null, pinnedModel: null }`. The PR's own test comment in `settings-overrides.test.ts` names this state ("with neither an override nor AGENT_MODEL, no tier resolves … which is a real state") and then sets `modelTierImplement: "heavy"` to step around it.

Measured against the shipped `lanes.yaml` with `primaryLane: "openrouter-glm"`, `AGENT_MODEL` unset, kind `implement`:

```
tier: null  model: null  prices: null
has --max-budget-usd: true
charge: { usd: 0.194985, basis: 'harness', reportedUsd: 0.194985 }
```

So on the lane that declares prices for all three tiers:

- the fleet charges the CLI's fiction — $0.194985 for a turn whose real cost at the lane's own prices is ~$0.0029 — against the attempt budget and the daily cap. That is acceptance criterion 6 ("where the CLI figure is untrustworthy, spend is derived from configured lane prices and token counts") unmet in the default configuration.
- `buildClaudeTurnCommand` (`claude-code.ts:114`) re-adds `--max-budget-usd 20`, which is precisely the second defect this PR reports finding: the turn stops at roughly $0.30 of real spend, mid-work, and is parked as finished because a budget stop is not `error_max_turns`.

Setting `modelTierImplement` does not rescue it: I re-ran with that override and `kind: "review"` and got the same three lines, because each tier field falls through to its own environment default by design (CLAUDE.md's rule, deliberately).

Neither existing test can see this. `lane-config.test.ts`'s "prices every lane that is not Anthropic-direct" asserts on the lane *definition*; the settings round-trip test pins a tier.

The two halves may deserve different answers, and I am not prescribing one. Which tier's prices to charge genuinely is not knowable when the harness picks its own model, so falling back to the harness figure there is arguable and errs in the direction this PR states is safe. But the *flag* decision needs no tier: "this lane's provider is not the one the CLI prices against" is a fact about the lane definition (`lane.prices !== null`, or `baseUrl !== null`) available before tier resolution, and keying the flag on the resolved tier reintroduces exactly the invisible truncation the PR set out to remove. Whichever way it goes, please pin the unset-tier case with a test — it is the default, not an edge.

## 2. Blocking — `drizzle/meta/0026_snapshot.json` is missing

Migrations 0019–0025 each ship a snapshot; 0026 ships only the `.sql` and a journal entry, so the newest snapshot still records `quota_state.id`. `migrate()` does not read snapshots, which is why everything passes here — but `drizzle-kit generate` does. Run against a copy of `drizzle/` plus `schema.ts`:

```
Is lane column in quota_state table created or renamed from another column?
❯ + lane      create column
  ~ id › lane rename column
```

The next migration author is therefore re-asked about a rename that already shipped, and either answer puts a second `RENAME COLUMN id TO lane` (or an `ADD COLUMN lane`) into 0027. `migrate()` runs at module load in `src/db/index.ts`, so on any DB already at 0026 that is a boot failure rather than a caught error. Given this PR edits the "Adding a generated migration" section of CLAUDE.md, the snapshot is the one artefact that section's neighbours would expect to find beside the SQL.

## Not blockers

- **The two unmet acceptance criteria** (a full implement pass, and a review + triage pass, in a real container) are blocked by an unfunded OpenRouter account and the absence of a Docker daemon in the worktree — not by anything in this diff. Reporting them rather than faking them is what the ticket explicitly asks for, and `lanes.yaml`'s note above the lane is appropriately honest that the three slugs are listing-verified only. That is a call for the human signing off, not for me.
- Stray double blank line at `src/lib/fleet/rows.ts:170`.
- No smell-baseline findings. `parsePrices` follows the shape `parseModels`/`parseCaps` already set, `lane-cost.ts` is the single home for the charge, and extracting `currentPrimaryLane` so the sweep and the dashboard cannot name different lanes is the right call rather than a second copy of the same three reads.

The evidence discipline in this change — four `result` events checked in verbatim, `costOverstatement` derived from them, `readTurnUsage` refusing the `usage` fallback with the fixture that proves why — is its strongest part, and I would like to see it land once the resolution gap above is closed.

lennons301:
Driver note for attempt 2: PR #183 (#174) also mints migration index **0026** (`0026_metered_money_guards` + snapshot) and will likely merge first. When integrating `origin/main`, renumber this branch's quota-state migration to **0027** (SQL + snapshot + journal entry, `when` bumped ≥1 day past 0026's) rather than resolving the journal conflict in place.

lennons301:
Driver note, superseding my renumber-to-0027 comment: PR #185 (#170) *also* minted a migration 0026, so the right rule is order-agnostic — when integrating `origin/main`, renumber this branch's migration to **one past whatever index main's journal then ends at** (SQL filename + snapshot + journal entry, `when` ≥1 day past the prior entry).

lennons301:
Both blocking findings are addressed on the branch — details in the updated PR body under **Review round 2**.

- The priced-lane mechanism now engages in the **unset** state, which was the real hole: the flag decision is keyed on the lane *definition* (`declaresPrices`), so it needs no tier at all, and the tier itself falls back to the lane's own `standard` on a priced lane rather than to "let the harness pick a model from another provider's catalogue". Pinned by four tests, including one through the settings route with no tier set.
- `drizzle/meta/0026_snapshot.json` is written; `drizzle-kit generate` now reports no changes and no rename prompt. CLAUDE.md's migration section says why the snapshot is not optional.
- Stray blank line in `rows.ts` gone.

A second `/code-review` pass over the result found three more, all fixed or stated: a priced lane with a pinned model was reporting `basis: "harness"` (now `harness-unpriced`); the tier rows would have said "no `--model`" over a priced lane (they now read the same fallback the resolver does); and the comment justifying the dropped `--max-budget-usd` cited `attemptExhaustion`, which runs for implement/repair only — so a **review or triage pass has no in-turn ceiling on a priced lane**. That one is a real loss, corrected in the comment rather than papered over, with the reasoning for keeping the trade.

The two container acceptance criteria are still blocked and I re-checked both today: `node scripts/openrouter-lane-check.mjs --model z-ai/glm-5.3-flash` still returns `402 billing_error — this account never purchased credits`, and there is still no Docker daemon in this environment.

`tsc --noEmit` clean, 1359 tests passing, changed files lint-clean against the known baseline.

lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/quota-tile-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/execution-lane-settings.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/quota-tile.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0026_quota_state_per_lane.sql` |
| persistence | `drizzle/**` | `drizzle/0026_quota_state_per_lane.sql` |

lennons301-reviewer (dismissed):
Approve. Both blocking findings from round 1 are properly closed — I re-measured each the way I found it — and the rest of the change holds up. This PR is gated (`human-signoff`, auto-merge off), so this approval satisfies branch protection only; the two unmet acceptance criteria below are the human's call, not mine to wave through or to hold the diff for.

## Round 1's blockers, re-checked

**1. The priced-lane mechanism now engages in the *unset* state.** This was the real defect: `tier` was null in the default configuration (no `AGENT_MODEL`, no stored tier) — the state one press of the lane chip leaves behind — so the fleet charged the CLI's fiction *and* re-added `--max-budget-usd` against it. The fix splits the two questions the way the round-1 note hoped it might:

- the flag reads `ResolvedLane.declaresPrices` (a fact about the endpoint, true before any tier resolves), so the pinned-model case keeps the guard — `claude-code.test.ts` pins that case explicitly, which is the one that regressed;
- the tier falls back to the lane's own `standard` via `laneFallbackTier`, and `resolveLane` applies it only when nothing else named a tier *and* no raw model is pinned.

Pinned where it counts: `settings-overrides.test.ts` drives the real PATCH route against the real checked-in `lanes.yaml`, switching to `openrouter-glm` and nothing else, and asserts `tier: "standard"`, `model: "z-ai/glm-5.3-flash"`, non-null prices, `declaresPrices: true`, and the same identifier on the settings row the screen renders. That is the exact configuration I measured as broken.

**2. `drizzle/meta/0026_snapshot.json` is present and correct.** Re-run the way I found it — a scratch copy of `drizzle/` plus `src/db/schema.ts` through `drizzle-kit generate` — and it now reports `6 tables … No schema changes, nothing to migrate`, with no rename prompt. The `RENAME COLUMN` cannot be re-asked of the next author, which was the boot-failure path.

## What else I verified

- `npx tsc --noEmit` clean. Full suite: **73 files / 1359 tests passing** (fresh `local.db` removed first, per CLAUDE.md).
- Lint on every changed file: only the documented baseline — the 9 `require()` errors in `output-parser.test.ts`, and I confirmed from the diff that this PR touches only the `createOutputHandler(TASK_ID, LANE_ID)` call sites, not those lines. Matches the PR body.
- **The evidence reproduces.** `909×$1 + 61×$5 + 13979×$0.10 + 6711×$2 = $0.0160339`, exactly the CLI's own `total_cost_usd` on the subscription fixture — so `modelUsage` really is the aggregate, and `usage.input_tokens: 10` beside it really is the last iteration. On the OpenRouter fixture, `38387×$5 + 122×$25 = $0.194985` to the digit, i.e. Anthropic list rates applied to a free MiniMax model; at the lane's own GLM Flash prices the same tokens are `$0.0029095`. The 67× is not rhetoric. (The exact match also settles a question the fixture raises but the comments don't: `thinkingTokens` is inside `outputTokens`, not additional — otherwise the reproduction would miss by $0.0002.)
- **Substitution happens once.** Every cost consumer reads `turnResult.costUsd` — `tasks.totalCostUsd` (both the first-turn and follow-up sites), `attemptExhaustion` at both call sites, `runs.totalCostUsd` via `syncRunCost` — and `runTurn` is the only place that rewrites it. Nothing reads `total_cost_usd` behind it.
- **The per-lane invariant is tested end to end against the shipped catalog**, not a fixture: `per-lane-gate.test.ts` records a live `rejected` window on `claude-subscription`, then shows the gate open on `openrouter-glm`, closed again on the switch back, and the subscription's own row untouched by the trip. That is the wedge #182 would otherwise have introduced here, and the one-directional fix (per-lane keying plus the gate's existing "no observation is not a closed gate") is the right shape — no new branch in the gate.
- Secrets: `lanes.yaml` carries variable names only (its own test asserts the file contains nothing secret-shaped), `FleetLaneRow` is deliberately three fields, `LaneView` adds `prices` and no credential, and the adapter passes `lane.id` — not the lane — into `createOutputHandler`, so auth values stay off the logging path. `scripts/openrouter-lane-check.mjs` reads the key by variable name and never prints it.
- The dropped `--max-budget-usd` really is the only in-turn ceiling review/triage had: `DEFAULT_REVIEW_BUDGET_USD`/`DEFAULT_TRIAGE_BUDGET_USD` appear in `turn-manager.ts` (the flag) and `fleet-view.ts` (display) and nowhere else. The disclosure is accurate, and so is the argument that converting the ceiling into the CLI's currency would mean modelling its pricing.

## Findings — none blocking

**1. The "what still bounds a priced turn, exactly" comment omits interactive passes, and they are the least bounded case.** `claude-code.ts:118` enumerates what survives (`attemptExhaustion` for implement/repair, the daily cap) and names review/triage as the loss. But an **interactive** task has no run — so no `attemptExhaustion` — and is exempt from the daily autonomous cap by construction. Its $20 ceiling reached the harness through `config.maxBudgetUsd` and this same flag (`turn-manager.ts:473`, and `:1056` for follow-up turns), so on a priced lane an interactive turn now has **no dollar ceiling at all**, only `--max-turns`. On `openrouter-glm` that is pennies; on the pre-existing `openrouter` lane, priced at Anthropic's $15/$75, it is real money in a 50-turn session. The trade is defensible and identical in kind to the review/triage one — but a comment that says "exactly" should say this too, and the human signing off should know it.

**2. The `openrouter` lane's prices are Anthropic list prices, asserted rather than measured.** The comment says the figures "happen to match Anthropic's list prices, which is the point". If OpenRouter carries any margin on those slugs, the fleet *under*-charges by it — the one direction `lane-cost.ts` says it refuses to fail in. Same listing-only provenance as the GLM rows, but without their explicit "verified only as far as the model listing" caveat. Worth re-checking with the script (`--model anthropic/claude-sonnet-4.5`) once the account is funded, alongside the GLM slugs.

**3. Possible data clump: `tierModels` + `fallbackTier`.** They are both derived from the primary lane, produced together by `laneState` in the overrides route, and threaded together through `describeModelTierSettings` into `resolveModelTierField` — now four positional optional parameters. One small lane-context object would stop the next lane-derived field being a fifth. Suggestion only; the current shape is tested and readable.

**4. The fallback *rule* is shared; its *condition* is written twice.** `resolveLane` says `chosen.tier ?? (pinnedModel === null ? laneFallbackTier(lane) : null)`; `resolveModelTierField` says `envTier ?? (envValue === null ? fallbackTier : null)`. I walked the four states (override / env tier / pinned id / unset) and they agree today, and each has a test — but they express "unset and not pinned" in two vocabularies, and a third caller is where that drifts. Noting, not asking.

**5. Two nits.** `lane-config.ts` and `lane-config.test.ts` credit the degrade ladder to "#164's ladder" while CLAUDE.md's new bullet says "#170's ladder" — #170 is the ladder ticket, #164 the parent. And `per-lane-gate.test.ts`'s last case, "gates nothing when no lane resolves", also asserts that an override naming an undeclared lane falls back to `claude-subscription`; that is a pre-existing `choosePrimaryLane` behaviour and a slightly surprising one (the settings route refuses undeclared ids, so it only bites if a lane is deleted from `lanes.yaml`), and it deserves its own test name rather than riding along with the null case.

## The two unmet criteria — the human's call, not a code gap

- [ ] *A full implement pass runs to completion on the lane in a real container, producing a branch and a PR*
- [ ] *A review pass and a triage pass each run on the lane*

Both blocked outside the diff: the OpenRouter account has never purchased credits (every paid slug 402s, so all measurement was on a free model), and this environment has no Docker daemon. The ticket explicitly asks for a stop-and-report rather than a shim, and `lanes.yaml`'s note above the lane is honest that the three GLM slugs are listing-verified only — it does not imply a call was made. I would not hold the diff for either, but I would not switch the primary lane on the strength of it: after merge, add credits, run `node scripts/openrouter-lane-check.mjs --model z-ai/glm-5.3-flash` to confirm the slug and its price, then arm one small ticket on `openrouter-glm` and read the feed's lane-charge line before anything larger. Note finding 1 while doing so — interactive sessions on a priced lane are the uncapped case.

Nothing in this change touches the subscription lane's live behaviour: it declares no prices, so it keeps `--max-budget-usd`, keeps the harness's (correct) figure, and its `quota_state` row is re-keyed in place by the migration. The `openrouter` lane's unset state does change, which is the intended consequence of finding 1 in round 1.

The measurement discipline is still the strongest part of this: four `result` events checked in verbatim, both cost claims reproducible from them to the digit, `readTurnUsage` refusing the `usage` fallback with the fixture that proves why, and a standing script so the next person can re-ask the endpoint rather than re-read the prose.

lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/quota-tile-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/execution-lane-settings.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/quota-tile.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0028_quota_state_per_lane.sql` |
| persistence | `drizzle/**` | `drizzle/0028_quota_state_per_lane.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0028_snapshot.json` |

lennons301-reviewer (approved):
## Approved — gated for human sign-off

Reviewed against issue #175's ten acceptance criteria, `CLAUDE.md`, `docs/agents/`, and `~/code/platform/standards/review-gates.md`. Objective signals, run locally (no CI checks are configured on this branch):

- `pnpm test` — **1500 passed / 80 files**, clean run after `rm local.db*` (per the migration convention). No cross-graph-state/queue-slots flakiness this run.
- `npx tsc --noEmit` — clean.
- `pnpm lint` — 16 errors, **exactly the known `main` baseline**: 9 + 6 `no-require-imports` in `output-parser.test.ts` (on lines this PR did not touch — verified against `git show origin/main:` — plus the pre-existing unused `result`) and one `set-state-in-effect` in `preview-pane.tsx`, untouched here. No new lint findings in any changed file.

### What I verified in the diff

**Cost (AC6) is the load-bearing part and it holds up.** `chargeForTurn` is applied once, in `runTurn`, so every budget in the fleet reads one figure — I confirmed `createOutputHandler`/`chargeForTurn` have exactly one non-test call site each, and that the charged `costUsd` is what flows into `updateTask` → `runs.totalCostUsd` → `metered_spend`. The three-way basis (`harness` / `lane-prices` / `harness-unpriced`) fails in the safe direction on every branch, and `readTurnUsage` reading `modelUsage` *only* is the right call: I recomputed the checked-in fixture and it does what the comments claim — Haiku list prices over `modelUsage` reproduce `total_cost_usd` to `0.0160339` exactly, and `38387×$5 + 122×$25` is exactly the `0.194985` the CLI billed a free model. Evidence checked in beside the rule is the right shape for a claim about a specific observation.

**Quota (AC7).** `quota_state` keyed by lane, one reader (`getQuotaObservation(lane)`), and the sweep + dashboard both take the lane from the single `readMoneyGuards`/`primaryLaneOf` resolution — so "whose quota is judged", "whose spend is capped" and "which lane is on the screen" cannot diverge. `per-lane-gate.test.ts` runs against the real shipped `lanes.yaml` rather than a fixture, which is the version of that test worth having, and it pins both directions (metered lane open under a standing subscription wall; closed again on switching back). The gate's own "no observation is not a closed gate" rule does the rest — per-lane keying really is the whole fix.

**Migration.** `0028` is a faithful `RENAME COLUMN` plus a re-attribution of the single `fleet` row to `claude-subscription` (which matches the lane id in `lanes.yaml` — I checked). Snapshot `0028_snapshot.json` present with correct `prevId` chain to `0027`, journal `when` bumped exactly +1 day per `CLAUDE.md`. Every test DB applies the migration, so the rename is exercised.

**Lane declaration + secrets (AC2, AC3, AC10).** Variable *names* only in `lanes.yaml`, on the API payload and on the settings panel; the fixture and the probe script carry no key. AC10 is pinned end-to-end through the real route (`switches to the open-weights lane and back, with no redeploy`), including the unset-tier state a single chip press actually leaves behind.

**The two round-2 fixes are the right shape.** Keying `--max-budget-usd` on `ResolvedLane.declaresPrices` (the lane definition) rather than on the resolved per-tier `prices` is correct — it is a fact about the endpoint, and the pinned-model case is exactly where a reinstated ceiling would truncate a turn with nobody watching. And `laneFallbackTier` is consistent across both readers: the pass path (`resolve.ts`) and the settings rows (`resolveModelTierField` via `describeModelTierSettings`) reach the same answer from one exported function, so a row cannot say "the account default" over a lane running a named slug. Default-lane behaviour is untouched throughout: `claude-subscription` declares no prices, keeps the flag, and still resolves unset to no `--model`.

No smell-baseline findings worth raising: one funnel per fact, no duplicated cost logic, no speculative hooks (every exported member of `lane-cost.ts` has a caller).

### For the human merging this — three things to look at, none of them code defects

1. **Two acceptance criteria are unmet and the PR says `Closes #175`.** The container criteria (a full implement pass on the lane producing a branch/PR; a review and a triage pass on the lane) were not executed — blocked by an OpenRouter account with no credits (`402 billing_error`) and no Docker daemon in the worktree. That is honestly reported in the body, on the issue and in `lanes.yaml`'s own note, which is the right behaviour, but merging as-is closes the ticket with two boxes unticked. Please either drop the `Closes` keyword, or plan to reopen / file the follow-up to execute them once credits are on the account (`node scripts/openrouter-lane-check.mjs --model z-ai/glm-5.3-flash`, then `Settings → Execution lane → openrouter-glm` and arm a small ticket). Consequently the three GLM slugs and their prices are listing-verified, not call-verified.
2. **A priced lane gives review and triage passes no in-turn spend ceiling.** Their $5/$2 budgets reached the harness through `--max-budget-usd` and nowhere else, and `attemptExhaustion` covers implement/repair only — so on `openrouter*` those passes are bounded by `--max-turns` and the daily caps at pickup, not within a turn. The trade is argued well (a ceiling in the CLI's currency truncates a review at ~$0.08 of real spend, mid-work and invisibly) and is stated in the code, the PR and `CLAUDE.md`, so this is a *conscious* residual rather than a gap — but it is the one consequence of this PR I would want an owner to have consciously accepted, since it only bites on the lanes this ticket exists to make usable.
3. **`runs.totalCostUsd` mixes currencies if a lane is switched mid-run.** Flagged in the PR body and on the issue, harmless while a switch is a deliberate press between attempts, but #173/#176 make mid-run switching routine and there is no filed ticket for it yet. Worth filing (or noting on #176) rather than leaving it only in a PR body.

### Non-blocking suggestions

- `harness-unpriced` (a priced lane whose turn reported no tokens, or a pinned model at no priced tier) charges the harness's over-statement *and* stays silent — `noteLaneCharge` fires only on `lane-prices`. The feed then shows the fiction as if it were the price. A one-line note on that basis too ("charged the harness's figure; this lane's prices could not be applied") would keep the "written down where the operator looks" property the rest of this change earns.
- The migration's *data* half — a pre-upgrade DB with a `fleet` row coming out keyed `claude-subscription` — has no test, though `create-test-db` exposes the raw handle for exactly that (`runs-ledger` does it). Low risk, since the row is latest-wins and the next turn rewrites it.
- `docs/runbook.md` §"Spending real money" is now slightly behind the operator-facing reality: what a priced lane's two cost lines in the feed mean, the "bounded by spend" quota tile, and point 2 above. `CLAUDE.md` is thoroughly updated; the runbook is the other half of that story.
- Pre-existing but adjacent: in `src/db/schema.ts` the long `quota_state` doc comment (which this PR edits) sits immediately above `meteredSpend`'s comment rather than above `quotaState`, so both tables' docs are stacked over the wrong one. Came in with #174's merge, not here — worth a drive-by whenever that file is next touched.

PR #185, opened 2026-09-02
Tier degrade ladder: step down instead of pausing when a tier allowance is exhausted

### LINKED ISSUE #170: Tier degrade ladder: step down instead of pausing when a tier allowance is exhausted

lennons301:
Closes #170

Losing one tier's allowance now steps the fleet down a tier and carries on
rather than stopping it. The two halves of #164 turn out to be one mechanism:
the CLI's `rate_limit_event` names the window that refused a pass, and a
**tier-scoped** one (`seven_day_opus`, `seven_day_sonnet`) says the account
still has quota — just not for that tier. So the run steps
`heavy → standard → light` and retries in place, and only an account-wide
rejection (or one at the bottom of the ladder) still reaches #168's pause.

## What changed

- **`src/lib/quota/tier-ladder.ts`** — a pure leaf module with the two rules.
  The window's tier scope is **derived, not enumerated**: the trailing segment
  of the limit type goes through the same `normalizeModelTier` a ticket's
  `model:` directive does, so a `seven_day_haiku` a later CLI adds is
  understood on arrival, while `seven_day_overage_included` reads as
  account-wide — the same cautious answer an unrecognised window gets, which
  is `rate-limit-event.ts`'s verbatim-enum rule read from the other end. The
  step is to the first rung strictly **below both** the tier that ran and the
  tier the window names, because a `seven_day_sonnet` wall observed on a heavy
  pass would otherwise step straight into the window that just refused.
- **`decideNext`** expresses the decision (AC 6), ahead of the blocked-marker
  and empty-pass readings for the same reason the pause is: a refused turn
  never reached the model, so its empty diff is the wall's and its "final
  message" is the CLI's own session-limit line.
- **`detectQuotaRejection` now reports a reset-less rejection** instead of
  hiding it. It was hidden while pausing was a wall's only possible
  consequence; a degrade waits on no clock, so what a missing reset means is
  the reducer's call now. An account-wide reset-less wall still takes the
  ordinary path and spends the attempt, exactly as before.
- **`runs.degraded_from`** (migration `0026`) records the tier the run was
  *asked* for, written once on the first step. `runs.model` keeps meaning the
  tier that **actually ran** — it is what the spend is read against and what
  every later pass of the attempt resolves through — hence two columns rather
  than one (AC 4).
- **The retry** is a fresh queued task of the same kind under the same run,
  carrying the same prompt, branch and PR, so a repair degrades as a repair.
  The run's status is deliberately untouched: an `implementing` run with a
  queued task is a state the sweep already reads correctly
  (`implementPassSettled` calls a queued pass unsettled, `hasInFlightRepairTask`
  sees the queued repair) and boot leaves alone. No attempt and no interruption
  is spent. The ladder bounds its own retries — `runs.model` only moves
  downward — so there is no counter.
- **Dashboard + digest** (AC 5): a degraded run stays under **Running** looking
  like work (mode chip, green phase) with one quiet line naming both tiers. The
  digest reads the same `degraded` field, so the two cannot disagree.
- Both refused-pass endings now share `endRefusedPass`, which owns the ordering
  that is load-bearing in each (container handle taken *before* the terminal
  write, #159; the replacement pass queued *before* the teardown; removal, not
  the blocked run's stop, #93 / 2026-08-04) and is the only writer of the run
  patch, so neither ending can quietly spend an attempt.

## Tests

Table tests in the reducer spanning every limit type this build knows plus two
it does not, crossed with the tier that ran (AC 6); the ladder's rules pinned
separately; and the ledger writes driven through the real pass-completion seam
over a migrated in-memory database, since none of the accounting promises are
observable from the reducer. 1358 tests pass; `eslint` is clean on every
changed file.

## Self-review findings acted on

Ran `/code-review` against `origin/main` with the issue as the spec. Two
findings were real and are fixed in `7e3722f`'s parent:

1. **The follow-up drain loop dropped the new decision.** It broke on a *list*
   of outcomes (`paused` / `blocked` / `finalized`), so a fix-up or repair turn
   refused on a tier allowance fell through — finishing a pass that had just
   been degraded, or execing the next queued message into a container
   `removeTaskContainer` had already deleted. Fixed by stopping on *anything
   but* `proceed`, which is the only decision that leaves the loop work: a
   decision added later now stops the drain by default, which is the safe
   direction and the one the list got wrong.
2. **An attempt could buy a second budget.** The budget check is per *task*, so
   the retry — a second implement task on the same run — would have received a
   fresh `run.budgetUsd` on top of whatever the refused pass spent. The
   implement branch now passes the attempt's *remaining* budget. This is a
   no-op everywhere else: an interrupted run is re-claimed as a **new** run
   (fresh row, `totalCostUsd` 0), review and repair passes carry their own
   budgets, and fix-up turns go through the follow-up path, which already
   subtracted the task's spend. The degrade retry is the only second
   implement-kind task a run can own.

Also fixed: the dead narrowing clause in the reducer, and a typo.

## Where I disagree / deliberately left things

- **"Nothing tests the pickup end to end."** True — `quota-degrade.test.ts`
  stops at the queued row, because running it would need a Docker daemon. The
  chain it relies on (`startTask` resolves the lane from `runs.model`, which
  the degrade rewrote) is pinned at both ends instead. The drain-loop defect
  above was likewise not unit-testable without a container, so it was fixed
  structurally — the bug class is now unrepresentable — rather than by adding a
  case to a list a test would have to remember to cover.
- **The step is one-way.** Nothing restores `runs.model`, so the rest of the
  run stays a rung down even past the window's reset; re-testing the wall would
  cost an agent turn on every later pass. A later attempt is a fresh run and
  starts at the configured tier. Now stated in the runbook and at the write
  site rather than implied.

## Interaction worth a reviewer's eye (not fixed here)

`main` gained #182's quota admission gate while this was in flight (merged in
cleanly, one import conflict). That gate closes on **any** `rejected`
observation, whatever window it names — so after a `seven_day_opus` rejection
no new tickets are claimed for up to seven days, even though the fleet can
still work a rung down. This ticket's ACs are all run-scoped and are met, but
the issue's framing ("steps the *fleet* down a tier and carries on") arguably
wants the gate to be tier-aware too. I have not touched #182's behaviour: a
tier-aware gate needs a fleet-level notion of "which tier are we on", which the
ladder deliberately keeps per-run, and inventing that policy unilaterally
seemed worse than naming it. Probably a follow-on under #164.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/running-list-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/running-list.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0026_run_degraded_from.sql` |
| persistence | `drizzle/**` | `drizzle/0026_run_degraded_from.sql` |

lennons301-reviewer (dismissed):
## Approve — sound, well-tested, and gated for your merge

Reviewed against issue #170's six ACs, the repo's CLAUDE.md conventions and the #167/#168 rules this builds on. All six ACs are met, and the objective signals are green locally (this repo has no PR CI — only deploy-on-merge-to-main — so I ran everything the merge would otherwise discover in production):

- `pnpm vitest run` — 72 files, **1358 tests pass**
- `tsc --noEmit` — clean
- `eslint` on every changed file — clean
- `pnpm build` — succeeds
- migration `0026`: journal `when` is exactly +1 day past `0025`, `prevId` matches `0025`'s id, and the snapshot diff adds `runs.degraded_from` and nothing else

### What I verified beyond the tests

- **The degrade actually sticks.** `startTask` rewrites `runs.model` on every implement-shaped pass (`model: passLane.tier ?? passModel`), so the load-bearing question is whether the retry resolves back to the configured tier. It doesn't: `requireLane(kind, run.model)` → `resolveAgentModelChoice` treats the run's `model` column as the ticket directive, which outranks override and env for work-carrying kinds, so `standard` resolves to `standard` and the write is idempotent. The one-way claim holds.
- **The queued retry is picked up and nothing else drives the run.** The poll loop selects any `queued` task (kind-priority ordering only), so the retry is not subject to #171/#182's pickup gate — correct, since that gate holds *new* pickup. `workingTaskOf` returns the newest implement/repair task, so `implementPassSettled` reads the queued retry as unsettled and gate evaluation waits; `hasInFlightRepairTask` includes `queued`; `finalizeDanglingRuns` skips a run owning a queued task, so a restart mid-degrade doesn't ghost the run. Every claim in the PR body checked out against the code.
- **`endRefusedPass`** removes the duplication the second ending would have created and keeps the #159 ordering (handle before the terminal write) and the #93/2026-08-04 removal-not-stop rule in one place. Making it the sole writer of the run patch is the right guard against an ending quietly spending an attempt.
- **The budget fix is real and correctly scoped.** A second implement task on one run would have received a fresh `run.budgetUsd`. Worth noting for the record that the `Math.max(0, …)` floor is unreachable in practice, because `attemptExhaustion` is checked *before* `evaluatePassOutcome` in both the initial-turn and drain paths — so a pass that spent the attempt fails it rather than degrading.
- **Stopping the drain on `!== "proceed"`** is the right structural fix, not just the smaller diff: the bug class is now unrepresentable.

### Non-blocking findings, for you or a follow-up

1. **Misplaced doc block** (`src/lib/orchestrator/turn-manager.ts`, lines ~1654–1712). `degradeRunTier`'s JSDoc is stacked immediately above `endRefusedPass`'s own JSDoc, so the helper appears to carry both and `degradeRunTier` (line 1734) is left undocumented. Comments are load-bearing in this repo, so this is worth moving — a one-line fix at merge or in a follow-up.
2. **A fix-up/repair turn refused on a tier wall loses its message.** `processQueuedMessages` stamps `deliveredAt` *before* the turn, so when a degrade fires mid-drain the delivered message (e.g. the reviewer's requested-changes feedback from `buildFeedbackTurn`) is gone, the task is failed, and the retry re-runs `task.description` — the original implement prompt — instead. The pre-#170 pause path lost it too, and the review cycle re-requests the changes so it is bounded and self-correcting, but it is neither tested nor documented for the degrade path, which unlike the pause carries on unattended. Worth a follow-up.
3. **`attemptExhaustion` wasn't given the same treatment as `maxBudgetUsd`.** On the initial turn it still compares *this pass's* cost against the full `run.budgetUsd`, so for a degrade retry the attempt-level ceiling is enforced only by the harness bound you now pass. Same outcome in practice; the asymmetry is just a little surprising to read next to the new comment.
4. **A restart during a degraded retry re-pays one refused pass.** The retry runs as a `running` task, so `markInterruptedRuns` marks the run `interrupted` and the sweep re-claims a *fresh* run at the configured tier — which walls again and degrades again. Self-correcting and costs no attempt, but the runbook's "a later attempt is a fresh run" doesn't quite cover the interruption case.
5. Cosmetic: mis-indented `tier: null,` in `review-tasks.test.ts` (no prettier in this repo, so lint won't catch it).

### The interaction you should weigh

I agree with the PR body's own flag: #182's admission gate closes on **any** `rejected` observation, so after a `seven_day_opus` rejection the fleet degrades the *in-flight* run but claims no new tickets for up to seven days — even though it demonstrably can still work a rung down. This ticket's ACs are all run-scoped and are met, and I think declining to invent a fleet-level "which tier are we on" unilaterally was the right call, but it does mean #170's headline benefit is partly withheld until a tier-aware gate lands. Worth filing under #164 before the next quota wall.

None of the above changes my read: this is a careful, well-argued change with tests that pin the promises where they live (reducer table tests for the decision, a migrated-DB executor test for the ledger writes). Approving — the `human-signoff` label stands and the merge is yours.

lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/running-list-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/running-list.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0028_run_degraded_from.sql` |
| persistence | `drizzle/**` | `drizzle/0028_run_degraded_from.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0028_snapshot.json` |

lennons301-reviewer (dismissed):
Approving. Reviewed with fresh eyes against issue #170, the repo conventions in `CLAUDE.md`/`docs/runbook.md`, and the #167/#168/#169/#171 designs this sits between. This PR is gated (`human-signoff`, auto-merge off), so this approval only satisfies branch protection — a human merges.

## Objective signals

No CI checks are reported on the branch (`gh pr checks 185` → "no checks reported"), so I ran them locally on the merge head `7370d43`:

- `pnpm vitest run` — 79 files, **1500 tests pass** (after `rm local.db*`, per the migration gotcha).
- `npx eslint` on all 22 changed files — clean.
- `npx tsc --noEmit` — clean.

## Acceptance criteria

1. **Tier-scoped rejection steps down and retries** — `planTierDegrade` (`src/lib/quota/tier-ladder.ts`) + the `degradeRunTier` branch in `decideNext`, executed by `degradeRunTier` in `turn-manager.ts`. ✔
2. **Account-wide pauses** — `limitScopeTier` derives the scope from the trailing segment through `normalizeModelTier`, so `five_hour`/`seven_day`/`overage`/`seven_day_overage_included` all yield null and fall to `pauseRunOnRateLimit`. Deriving rather than enumerating is the right call and matches `rate-limit-event.ts`'s verbatim-enum rule read from the other end. ✔
3. **Bottom of the ladder pauses** — the "first rung strictly below *both*" rule (`Math.max` of the two indices) covers both the bottom rung and the `seven_day_sonnet`-observed-on-heavy case that would otherwise step into the window that just refused. ✔
4. **Tier recorded and legible in the ledger** — `runs.degraded_from` (migration `0028`, journal `when` bumped a day past `0027` per the documented drizzle gotcha) alongside `runs.model`. Two columns is right: `model` is what the spend is read against and what `requireLane(task.kind, run.model)` resolves every later pass through, including repair (`isWorkPassKind` covers it), so the retry genuinely runs a rung lower. ✔
5. **Dashboard shows it** — `RunningCard.degraded` from `buildFleetView`, rendered quietly under Running; the digest reads the same field. The "both tiers or neither" guard against a half-written row is a nice touch and is tested. ✔
6. **Reducer + table tests spanning each limit type** — the `LADDER` table in `decide.test.ts` covers all six known types plus an unknown one plus `null`, crossed with the tier that ran and with a null tier. ✔

## What I checked beyond the ACs

- **No attempt, no interruption spent.** `endRefusedPass` is the only writer of the run patch and touches neither `attempt` nor `interruptionCount`; verified through `quota-degrade.test.ts`, which drives the real pass-completion seam over a migrated in-memory DB.
- **Ordering in `endRefusedPass`** is preserved for both endings: container handle taken before the terminal write (#159), transcript copied before teardown (#169), `queueNext` before teardown, and removal rather than the blocked run's stop (#93 / 2026-08-04). Folding the pause and the degrade into one ending is the right shape — it is exactly the kind of divergence that costs an attempt when it drifts.
- **The retry can't accidentally resume a session.** `restoreSessionTranscript` returns early on `task.sessionId === null`, and the degrade insert leaves it unset, so the "fresh conversation on the same branch" claim holds even on a run that had an earlier pause with a transcript on disk.
- **Sweep/boot don't misread the intermediate state.** `workingTaskOf` returns the newest implement/repair task, so post-degrade `implementPassSettled` sees a `queued` pass and gate evaluation waits; `hasInFlightRepairTask` sees a queued repair; boot's interrupt path only touches `running` tasks and the dangling-run finalizer only fires when all tasks are terminal. Leaving `runs.status` alone is correct, and the writes are synchronous between the terminal update and the insert so no poll can observe the gap.
- **Only implement-shaped passes reach it** — review and triage return before `evaluatePassOutcome`, consistent with #168.
- **The drain-loop fix** (`decision !== "proceed"`) is the right structural fix, and the reasoning for inverting the list is sound.
- **Budget lineage.** `resumedFromTaskId` on the retry means `spendCarriedIntoPass` nets the refused pass's spend off the attempt allowance, so a degrade doesn't buy a second budget. Good catch in self-review.

## Non-blocking findings

1. **Orphaned doc block** (`src/lib/orchestrator/turn-manager.ts:1820-1850`). Two JSDoc blocks sit adjacent: the "Step a run down the tier ladder and retry it (issue #170)" block is immediately followed by `endRefusedPass`'s own block and then `async function endRefusedPass`. So the degrade's documentation is attached to `endRefusedPass` (and discarded by tooling), while `degradeRunTier` at ~line 1928 has no doc comment at all. In a repo where this prose is the design record, that's worth a one-line move.
2. **`spendCarriedIntoPass`'s walk bound now has to cover two mechanisms.** The loop is bounded by `MAX_RESUMES_CEILING` (5) "because a legitimate chain is at most `MAX_RESUMES_CEILING` long" — true when only resumes wrote `resumedFromTaskId`. A chain can now be resumes **plus up to two degrade steps**, so a worst-case lineage exceeds the bound and the walk silently under-counts carried spend, handing the last pass more allowance than the attempt has left. Only reachable at the resume ceiling, and bounded by one pass's budget, but the constant's stated justification is no longer accurate. Suggest `MAX_RESUMES_CEILING + MODEL_TIERS.length` (or similar) with the comment updated.
3. **Duplicated shape** (Fowler baseline, judgement call): the continuation-task insert in `degradeRunTier` and the one in `sweep.ts`'s resume executor are the same row-shape with different prompt/session handling. A shared `queueContinuationPass(...)` would stop the two drifting on a field a third continuation later needs. Not urgent; they're in different modules and differ meaningfully.
4. **A degrade during a fix-up turn loses the reviewer's feedback context** — deliberately no session, and any still-undelivered messages stay on the row now marked `failed`, so the retry starts from the original brief and re-enters the review cycle from scratch. It converges (bounded by `reviewCycleCount`) and the pause path has the same shape, so this is not a regression — just naming it as a known edge.
5. **Cosmetic:** the PR body says migration `0026`; the file is `0028_run_degraded_from.sql` (renumbered after merging main).

## On the #182 interaction the author flagged

Agreed it's out of scope, and I'd add that this PR strictly improves the situation rather than creating it: before it, a `seven_day_opus` wall both paused the run *and* held pickup; now the run carries on and only the pickup hold remains. Whether the hold actually persists depends on which window the CLI names as the representative claim on a subsequent standard-tier call — if it keeps reporting the exhausted opus window at 100% utilization, `evaluateQuotaGate` stays closed on the utilization threshold even though the fleet can work. Worth a follow-on ticket under #164 for a tier-aware gate; not a reason to hold this.

For the human merging: nothing here needs re-work before merge in my view — findings 1 and 2 are cheap follow-ups.

lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/running-list-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/fleet/running-list.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0029_run_degraded_from.sql` |
| persistence | `drizzle/**` | `drizzle/0029_run_degraded_from.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0029_snapshot.json` |

lennons301-reviewer (approved):
Approving. GATED for human sign-off (`human-signoff`, auto-merge off), so this satisfies branch protection and a human merges after looking.

## Objective signals

No CI checks are reported on the branch, so I ran everything locally at `2eca6d4`:

- `pnpm vitest run` — 82 files, **1557 tests, all passing**
- `npx tsc --noEmit` — clean
- `npx eslint` on all 12 changed `.ts`/`.tsx` files — clean (no reliance on the known `main` lint baseline)
- Migration hygiene per CLAUDE.md: `drizzle/0029_run_degraded_from.sql` has its `drizzle/meta/0029_snapshot.json` beside it, the snapshot's `prevId` chains to `0028`'s `id`, the journal `when` is exactly one day past `0028`, and the snapshot's `runs` column count (34) matches the schema's — so the next author's `drizzle-kit generate` won't re-ask about this column.

## Acceptance criteria

All six are met, and each is pinned by a test rather than by prose:

1. **Tier-scoped weekly rejection steps down and retries** — `planTierDegrade` + the `degradeRunTier` action; `quota-degrade.test.ts` asserts the queued replacement exists and `runs.model` moved.
2. **Account-wide rejection pauses** — `five_hour` / `seven_day` / `overage` / `seven_day_overage_included` all land on `pauseRunOnRateLimit` in the reducer table.
3. **Bottom of the ladder pauses** — `["light", "seven_day_opus", "pause"]`, and driven through the real seam with `bootWith("light")`.
4. **The tier that actually ran is on the run and in the ledger** — `runs.model` keeps meaning "ran at", `runs.degraded_from` records "asked for", written once and provably stable across a second step (`heavy → standard → light` still reads `degradedFrom: "heavy"`).
5. **Dashboard shows a run below its configured tier** — `RunningCard.degraded`, rendered under Running with the both-tiers-or-neither rule in `buildFleetView`, and the digest reads the same field.
6. **Reducer decision + table tests spanning each limit type** — `LADDER` covers every window this build knows plus two it does not (`thirty_day_enterprise`, `null`) crossed with the tier that ran, including the "never step onto a rung the wall covers" case (`heavy` + `seven_day_sonnet` → `light`).

## What I checked beyond the ACs

- **The ladder actually closes the loop.** The claim rests on `startTask` re-resolving the lane from `runs.model`, which the degrade rewrote — the one link the tests stop short of (admitted in the PR body). I traced it: `requireLane(task.kind, run?.model ?? null)` feeds the run's tier in as the directive, and the implement-shaped run patch writes back `model: passLane.tier ?? passModel`, so the degraded rung is idempotent rather than being overwritten with the configured tier on the retry. The failure mode I was most worried about — a retry silently re-running at `heavy` — is not present.
- **`degradeRunTier` is unreachable for review and triage passes.** `evaluatePassOutcome` is only called under `isImplementShaped`; review and triage return earlier at `finishReviewPass` / `finishTriagePass`. Matches the stated "implement-shaped only" design.
- **Loosening `detectQuotaRejection` to report reset-less rejections has exactly one consumer.** `detectQuotaRejection` and `PassOutcome.rateLimited` each have a single call site, and the reducer's `resumeAfter !== null` guard reproduces the old fall-through for the account-wide reset-less case bit for bit (pinned by the "spends the attempt on a reset-less account-wide wall" test). No other reader inherits a newly non-null value.
- **No window where a degrading run looks dangling.** In `endRefusedPass` there is no `await` between `updateTask(status: "failed")` and `queueNext()`, so no sweep can interleave and see a run whose tasks are all terminal — which would have been `finalizeEmptyPass`/boot-finalize bait. Single-threaded by construction rather than by luck, and the queue-before-teardown ordering is durable across a crash as documented.
- **The second-budget hazard is closed by reuse, not by a new mechanism.** The retry carries `resumedFromTaskId`, so `spendCarriedIntoPass` nets the refused pass's spend off the attempt allowance — same lineage #169 established, unchanged. Exhaustion is still judged ahead of the wall at line ~601, so a pass that walled having already spent the attempt fails rather than degrading into an empty allowance; consistent with the documented #169 ordering.
- **The drain-loop fix is the right shape.** `if (decision !== "proceed") break` inverts a list that was already wrong the moment a fourth outcome existed. Stopping by default on an unrecognised decision is the safe direction, and it removes the bug class rather than adding a case a future author must remember.
- Smell baseline: nothing I'd act on. `endRefusedPass` is a genuine behavioural unifier (it owns the load-bearing ordering) rather than a Middle Man, its optional params are both exercised, and holding `limitType` as a verbatim string is the repo's stated rule read from the other end.

## For the human merging: the fleet-level interaction

The PR body names this and I agree it is the one thing worth a human's judgement rather than mine. `evaluateQuotaGate` closes on **any** `rejected` observation without looking at `rateLimitType`, so a `seven_day_opus` wall still holds *new* pickup for up to seven days even though this ticket has just proven the fleet can work a rung down. The in-flight run degrades and retries fine (the gate holds pickup only, and a queued retry is not pickup), so nothing here regresses — but the issue's framing, "steps the *fleet* down a tier and carries on", is only true run-scoped after this PR.

I think declining to fix it here was the right call: a tier-aware gate needs a fleet-level notion of "which tier are we on", which this ladder deliberately keeps per-run, and inventing that policy inside this ticket would be worse than naming it. Worth filing as a follow-on under #164 either way.

## Two nits, neither blocking

1. `src/lib/orchestrator/turn-manager.ts` — the `degradeRunTier` doc block (`Step a run down the tier ladder and retry it (issue #170)`) sits immediately above `endRefusedPass`'s own block, so two consecutive JSDoc comments precede `endRefusedPass` and `degradeRunTier` itself is declared with none. Both texts are correct; they are just in the wrong place, which in this repo (where the comments are the design record) is worth a one-line move.
2. The degrade's `tasks` insert and `executeResumeRun`'s in `sweep.ts` are now the same ~15-field continuation-pass shape in two places, differing only in the prompt transform, the session and `resumeCount`. A future column a continuation must carry would need both edits. Suggestion only — the differences are real enough that I would not insist on extracting it.

PR #186, opened 2026-09-02
Resume a rate-limited run losslessly once its window resets

### LINKED ISSUE #169: Resume a rate-limited run losslessly once its window resets

lennons301:
Closes #169

A paused run now resumes by itself once its quota window resets, continuing the
same agent conversation on the same branch.

## What it does

- **The sweep is the driver.** No new scheduler: paused runs are gathered every
  30-second tick like everything else that waits, and a resume is subtracted
  from the claimable slots exactly as a queued review is — which is what makes
  *resuming existing work outranks claiming new work* true when slots are
  scarce, rather than a matter of which action happens to execute first.
- **Lossless where it can be.** The pass's session transcript is copied out of
  the container before the pause tears it down, kept beside the SQLite database
  (so it inherits `/data` with no migration, and survives a restart), and
  restored into the fresh container so the pass continues the same session on
  the same branch. This rests on the #165 spike's measurement: one file is the
  whole transcript, a pass killed mid-tool-call resumes knowing how far it got,
  and the CLI finds a session by id rather than by path.
- **Every step degrades rather than fails.** A transcript that cannot be copied
  out costs the pause nothing — the pause protects the *attempt*, the
  conversation is the saving — and a restore that fails takes the spike's
  declared fallback (same branch, prior context lost). `--resume` against a
  session the container has never heard of would fail the pass outright, so it
  is the one thing the restore refuses to do. The resume prompt carries the
  pass's original brief behind its preamble for the same reason: it has to
  stand on its own whichever way the restore went.
- **Jitter is a hash of the run id**, not a random draw. The reducer is pure and
  re-decides every 30 seconds, so a random offset would move between ticks and
  make a run eligible at 12:01 and not at 12:02. The dashboard and the digest
  show the run's *eligible* instant through the same function, so no countdown
  hits zero minutes before anything moves.
- **The bound.** `resumeCount` is its own counter (a pause spends neither an
  attempt nor an interruption, and bumping either would change what those
  numbers mean), bumped when a resume is *decided* so a resume whose task never
  starts cannot retry every sweep forever. Past the bound the ticket routes to
  `ready-for-human` the way exhaustion does — but the run lands in `exhausted`,
  not `failed`, because attempts are counted from failed runs: re-arming later
  gets the attempts the ticket never used. Settable at **Settings ▸ Quota**
  (0–5, default 3, `MAX_RESUMES_PER_ATTEMPT` underneath).

## Two decisions worth an explicit look

1. **A resume is not held by the kill switch or the daily cap.** Both gate
   *pickup* — "no claim, no triage pass" — while everything already in flight
   is decided regardless (verdicts posted, fix-ups delivered, conflicts
   repaired), and a paused run is the middle of an attempt the fleet already
   started; abandoning it at a window boundary would spend the very attempt the
   pause protected. The consequence is real, though: an operator who engages the
   kill switch can still see a fresh agent container start hours later. The
   runbook says so and points at cancelling the run. Happy to invert it if you'd
   rather the switch mean "no new containers, full stop".
2. **A fix-up turn walled mid-review-cycle** resumes from the *original* brief
   when its transcript did not survive — the reviewer's feedback was delivered
   as a message, not as the task description. The review cycle then re-reviews
   what comes back, so it self-corrects, but it is a narrower fallback than the
   implement case.

## Self-review

Ran `/code-review` against `origin/main` with the issue as the spec. Acted on
everything I agreed with; the two real defects it found are fixed in
`e0b5e7c`-onwards and each has a test:

- a second resume wrapped the first resume's prompt (three stacked preambles by
  resume 3) — the preamble now ends at a marker the next one cuts on;
- the copy-out could persist a half-read transcript, because waiting on an exec
  races the process exiting against its output arriving and
  `Buffer.from(halfALine, "base64")` decodes happily — the container now prints
  the file's size after the payload and a short read is reported as no read.

Disagreed with one: it flagged the `MAX_RESUMES_PER_ATTEMPT` environment
variable as scope creep against "the bound is settable in the UI". The settings
layer's own contract (#166) is that an unset override *falls through to its own
environment default* and reports its provenance, so a UI-settable field without
a variable behind it would be the odd one out. It parses through the settings
field's own validator, so the two accept exactly the same values.

Not done, deliberately: the needs-you card for a quota-exhausted ticket still
reads "Attempts exhausted". Distinguishing it needs a discriminator the run row
does not carry, and the issue comment, the Discord embed and the label all say
what actually happened. Happy to add a column if you'd rather the card be exact.

## Integrating #182

`main` moved under this branch while it was in flight: #182 (the quota
admission gate, issue #171) landed the settings layer's *first* non-model field
while this one was landing its second. Merged in, and the resume bound adopts
that shape rather than keeping a parallel one — its own resolver and view
(`resolveResumeBound` → `ResumeBoundView`), the same three layers ending in a
built-in default, and its own panel under the **Quota** section the gate
created. The two read in order there: when the fleet stops *starting* work on a
spent window, and how long work it already started may keep riding one. The
sweep reads both off one per-tick read of the settings row, and the two agree by
construction — #171 holds pickup only and says a parked run still resumes, which
is what this does.

## Verification

`npx vitest run` — 1312 pass (72 files), `pnpm build` clean, `eslint` clean on
every changed file. The Docker transfer functions themselves are the one part no
test covers — they need a daemon — so both are bounded, both report rather than
throw where the caller has an answer, and the seams either side of them are
tested against a real store on disk.

---

## Review round 2 — the blocking finding, fixed

**A resumed pass no longer restarts the attempt's budget from zero.** The
finding was right and it was a miss, not a decision: a resume inserts a new
task row, every budget control in the turn manager is scoped to the task, so
one attempt's real ceiling was `(1 + resumes) x run.budgetUsd` — $80 by
default, $300 with a `budget:` directive — on the one path with no pickup gate
in front of it. Closed the way the review's option 1 asks, judging the bound
against the attempt's accumulated spend.

- `tasks.resumedFromTaskId` (migration `0027`) records which pass a resume
  continues, and `src/lib/orchestrator/pass-budget.ts` is the one place the
  three controls — the harness cap, the pre-turn check, `attemptExhaustion` —
  read: **a resume of an attempt already $18 into $20 gets $2**, pinned as a
  table test alongside the walk itself (16 tests, against a real migrated DB).
- **Lineage rather than the run**, which is the one place I diverged from the
  suggested shape. Summing the run's implement/repair tasks over-counts: a run
  can own two *distinct* repair passes, each entitled to its own $5, and a
  first repair pass would have the implement pass's spend subtracted from its
  $5 and start exhausted. Only the chain tells "the same pass, continued" from
  "another pass of the same kind".
- **Only the attempt's budget is netted** — a repair pass keeps its own $5
  ungross. My first cut netted it too, and my own `/code-review` pass caught
  two defects in that: the pre-provision refusal covered repair and routed
  through `failImplementAttempt`, which drives the run `failed` and so would
  burn one of the ticket's three attempts on a pass that is *never an attempt*,
  abandoning the PR unrepaired and bypassing the conflict escalation; and
  netting the $5 made the ceiling $5 in `startTask` but the run's budget in the
  drain loop, so a chain at $4.90 got ten cents to merge a branch with. Fixed
  in `fa9466c`. The residual is stated where the decision is: a resumed repair
  pass gets its $5 again, at most `MAX_RESUMES_CEILING` times, and closing it
  properly means giving repair one allowance per conflict episode across rows —
  including the fix-up turns that already charge against `run.budgetUsd` rather
  than the $5 — which is #54's accounting, not this ticket's.
- The pre-provision refusal is implement-only and is **defence in depth, not a
  live path**: exhaustion is judged ahead of the pause, so every link in a
  resume chain is strictly under the ceiling. It sits before the container so
  the alternative isn't ~2 GiB built to run a turn with no money in it.

## Non-blocking findings

- **`readContainerFile` enforcing the ceiling after the memory is spent** —
  taken. The container now refuses an over-size file itself (`wc -c` before
  `base64`, its own exit code so the reason can be logged), so the base64 line,
  the JS string and the decoded buffer — roughly 3x the file — never
  materialise. `saveTranscript` keeps its own check as the store's invariant.
- **Stray blank line in `settings-resolver.ts`** — gone.
- **`resolveCountSetting` at three copies** — not taken, as offered: there are
  two here, and the third (the overflow daily cap) is the right moment for the
  shared helper. Worth noting #183 has since landed `meteredDailyCapUsd` with
  its own resolver too, so the case is stronger than it was.
- **No direct test of `executeResumeRun` / `executeExhaustPausedRun`** — still
  none, consistent with `sweep.ts` having no test file. The "most recent
  implement|repair task" selection you flagged is the one bit I'd most like
  covered; what is covered now is the consuming half (the lineage walk, the
  chain, a distinct same-kind pass, a missing predecessor, a cycle).
- **The kill-switch/daily-cap exemption and the fix-up fallback** — left for
  the human call, as agreed.

## Integrating #183

`main` moved again: #174's billing-kind money guards landed the settings
layer's third non-model field and a migration at the same index this branch
used. Merged in — this branch's two column additions are regenerated as one
`0027_quota_resume_counters` on top of main's `0026`, both settings keys share
the allowlist rather than one replacing the other, and the resumed pass's
budget check sits above main's lane/billing write. The two money mechanisms
don't overlap: #174 guards *pickup* on billing kind, which is the same layer
#171's gate sits in and the same layer a resume is deliberately above.

## Verification

`npx vitest run` — 1443 pass (77 files), `pnpm build` clean, `eslint` clean on
every changed file.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/quota-resume-settings-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/quota-resume-settings.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0026_run_resume_count.sql` |
| persistence | `drizzle/**` | `drizzle/0026_run_resume_count.sql` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/resume-jitter.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/resume-jitter.ts` |

lennons301-reviewer (changes requested):
Reviewed against issue #169, the repo's CLAUDE.md conventions and `docs/agents/review-gates.yaml`. Objective signals first: no CI runs on this branch, so I ran them locally — `npx vitest run` is **1367 pass / 74 files, green**, and `eslint` is clean on all 27 changed TS/TSX files. The migration follows the documented Drizzle gotcha (journal `when` bumped exactly +1 day past 0025), and the `human-signoff` gate is correct (`src/db/schema.ts`, `drizzle/**`, `src/lib/docker/**`, `src/lib/orchestrator/autonomy/**`, `src/config.ts`, components — five gate categories).

The design is strong and matches the ticket closely: the sweep as the driver with no new scheduler, the resume subtracted from `claimableSlots` alongside queued reviews (so the priority is structural rather than ordering luck), jitter as a hash of the run id (the right call for a pure reducer re-deciding every 30s), the bound checked before the clock, `exhausted` rather than `failed` so the ticket keeps its unspent attempts, and a store keyed by run beside the SQLite database so "survives a restart" needs no migration. `decide.test.ts` covers every acceptance criterion as table tests, including the two I'd have written myself (a spread across the jitter window from one shared reset, and a resume proceeding while the kill switch holds pickup). The `readContainerFile` size check added in the self-review is a real defect caught — a half-decoded base64 line restored into the next container would have been very unpleasant to debug.

One blocking finding.

### A resumed pass restarts the attempt's budget from zero

`executeResumeRun` inserts a **new task row** for the same run, and every budget control in the turn manager is scoped to the *task*, not the run:

- `src/lib/orchestrator/turn-manager.ts:560` — `attemptExhaustion(run, turnResult.costUsd, ...)` judges the fresh pass's own cost against `run.budgetUsd`; the new task starts at `totalCostUsd = 0`.
- `src/lib/orchestrator/turn-manager.ts:489` — `maxBudgetUsd: run?.budgetUsd` hands the harness the *whole* per-attempt budget again.

So one attempt's real spend ceiling becomes `(1 + resumes) x run.budgetUsd`: **4 x $20 = $80 by default**, and up to **4 x $75 = $300** for a ticket carrying a `budget:` directive. `budgets.ts` calls `MAX_ATTEMPT_BUDGET_USD` a *hard* ceiling ("Issue text is semi-trusted input; nothing it says can exceed this") and CLAUDE.md states "$20 per attempt (ticket-directive override capped at $75)". This PR is also the one path where that spend accrues with no pickup gate in front of it, since resumes are deliberately exempt from the daily cap.

What makes it a defect rather than a stated trade-off: `runs.totalCostUsd` already knows the true attempt total (`syncRunCost` sums every task the run owns), so the number is sitting right there unused — and the PR body/CLAUDE.md carefully record what the pause does *not* touch (`finishedAt`, `attempt`, `interruptionCount`) while saying nothing about the budget. Either it was missed, or it is a decision that deserves to be written down like the kill-switch one below it.

Mitigating, but not enough to leave silent: a pass refused at the wall exits in ~2s having spent nothing (#165's finding), and rate-limit telemetry is subscription-only (#167), so today's dollars are notional. But the mid-flight wall — a pass that spends $15 and *then* gets refused on a later API attempt — is precisely the case the pause was built for, and those same numbers feed the $500/day cap and the dashboard.

Two ways to close it, either is fine by me:

1. Judge the bound against the attempt's accumulated implement/repair spend rather than the fresh task's — e.g. sum the run's `implement`/`repair` tasks (not review/triage, which carry their own allowances) and pass `run.budgetUsd - spentSoFar` as the resumed pass's `maxBudgetUsd`, with a table test that a resume of an attempt already $18 into $20 gets $2 rather than $20.
2. If a resumed pass is *meant* to get a fresh budget (defensible: the paused pass's spend largely bought work that is already pushed), say so explicitly — in the `executeResumeRun` docstring, the CLAUDE.md #169 bullet and the PR body — and name the resulting real ceiling, so the next reader of `MAX_ATTEMPT_BUDGET_USD` isn't misled about what "hard" means.

### Non-blocking

- `readContainerFile` materialises the whole file (base64 chunks -> one string -> decoded Buffer, roughly 3x the file) before `saveTranscript` applies `MAX_TRANSCRIPT_BYTES`. On the box with a documented OOM history this is the one place the ceiling is enforced *after* the memory is spent; having the container refuse to emit the payload when `wc -c` is over the ceiling would enforce it at the source, which is this repo's usual instinct.
- `resolveResumeBound` is `resolveQuotaThreshold` with three tokens changed (view interface included). CLAUDE.md endorses the shape deliberately, so I'm not asking for it here — but the `SettingKey` comment already names a third such field (the overflow daily cap), and at three copies the shared `resolveCountSetting(spec, overrides, builtIn)` is worth taking.
- `src/lib/settings-resolver.ts` gains a stray extra blank line after `SETTINGS_FIELDS`.
- `executeResumeRun` / `executeExhaustPausedRun` have no direct test (resumeCount bump, run left `rate_limited`, kind/branch/PR inheritance, session id only when the transcript is on disk). Consistent with `sweep.ts` having no test file at all, so I read it as convention rather than a gap — noting it because the "most recent implement|repair task" selection is the one bit of executor logic here that isn't obviously right by inspection.
- The kill-switch/daily-cap exemption and the fix-up-turn fallback are correctly flagged for a human, and this PR being gated is the right place for that call. No objection from me on either.

Everything else — the reducer/executor split, the settings-resolver shape (own resolver, own view, three layers, digits-checked before `Number()`, ceiling deriving both chips and validator), the `globalThis`/two-graph re-read in the executor, the boot-time prune, the fleet view and digest showing the *eligible* instant through the same function the reducer decides with — reads as done properly.

lennons301:
Blocking finding fixed — a resumed pass no longer restarts the attempt's budget from zero. Took option 1 (judge the bound against the attempt's accumulated spend), with one divergence from the suggested shape: the netting keys off **lineage** (`tasks.resumedFromTaskId`) rather than summing the run's implement/repair tasks, because a run can own two *distinct* repair passes and a first repair pass would otherwise start with the implement pass's spend subtracted from its $5.

Only the *attempt's* budget is netted — a repair pass keeps its own $5 gross. My first cut netted that too, and `/code-review` caught two defects in it: the pre-provision refusal covered repair and routed through `failImplementAttempt`, which would burn one of the ticket's three attempts on a pass that is never an attempt and abandon the PR unrepaired; and it made the ceiling $5 in `startTask` but the run's budget in the drain loop. Both fixed in `fa9466c`, with the bounded residual stated at the decision.

Also took the `readContainerFile` ceiling-at-the-source point and the stray blank line; left `resolveCountSetting` as offered. Details, and #183 integration, in the PR body.

lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/quota-resume-settings-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/quota-resume-settings.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0027_quota_resume_counters.sql` |
| persistence | `drizzle/**` | `drizzle/0027_quota_resume_counters.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0027_snapshot.json` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/resume-jitter.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/resume-jitter.ts` |

lennons301-reviewer (approved):
Re-reviewed against issue #169, CLAUDE.md and the round-1 findings. This PR is **gated** (`human-signoff` present, auto-merge off), so this approval satisfies branch protection only — a human merges.

Objective signals first, since no CI runs on this branch. Locally, from a clean `local.db` (the documented Drizzle gotcha): `npx vitest run` is **1443 pass / 77 files, green**, `npx tsc --noEmit` is clean, and `eslint` is clean on all 29 changed TS/TSX files. Migration `0027_quota_resume_counters` sets its journal `when` exactly +1 day past `0026`'s, as the repo's rule requires, and it is regenerated on top of main's `0026` rather than colliding with it.

## The blocking finding is closed, and closed better than the suggestion

Round 1's finding was that a resumed pass restarted the attempt's budget from zero, making one attempt's real ceiling `(1 + resumes) x run.budgetUsd`. `src/lib/orchestrator/pass-budget.ts` closes it at the one place the three controls read: the harness cap (`maxBudgetUsd: passBudget.remainingUsd`), the pre-turn check in `processQueuedMessages`, and `attemptExhaustion(run, carried + turnCost, ...)`. A resume of an attempt $18 into $20 gets $2, pinned as a table case.

Two divergences from what I suggested, and I agree with both:

- **Lineage (`tasks.resumedFromTaskId`) rather than summing the run's implement/repair tasks.** My version would have charged a first repair pass the implement pass's spend against its own $5 and started it exhausted, and would have merged two *distinct* repair passes into one allowance. The chain is the only thing that distinguishes "the same pass, continued" from "another pass of the same kind"; `spendCarriedIntoPass` is bounded by `MAX_RESUMES_CEILING` and cycle-guarded, with tests for a missing predecessor, a cycle, an over-long chain and a same-kind sibling.
- **Only the attempt's budget is netted; a repair pass keeps its $5 ungross.** The reasoning holds: a repair pass is never an attempt, so a refusal on the way in would have to fail an attempt it may not spend, and the residual is bounded (its $5 again, at most `MAX_RESUMES_CEILING` times) and is stated where the decision is. `fa9466c` also caught the two defects that first cut had — `failImplementAttempt` on a repair pass, and a $5 ceiling on the way in against `run.budgetUsd` in the drain loop.

The pre-provision refusal is correctly implement-only and correctly placed before `createWorkspaceContainer` rather than after ~2 GiB is built. Confirmed unreachable in the live path as claimed (exhaustion is judged ahead of the pause), so it reads as the defence in depth it says it is.

Behaviour for everything that is not a resume is unchanged: `spendCarriedIntoPass` returns 0 on a null `resumedFromTaskId`, and the drain loop's `spentUsd > 0 && spentUsd >= budgetUsd` preserves the old truthiness guard exactly.

## The other round-1 findings

- `readContainerFile` now refuses an over-size file **inside the container** (`wc -c`, own exit code 3, size logged) so the base64 line, the JS string and the decoded buffer never materialise — taken, and done at the source as asked. `saveTranscript` keeping its own check as the store's invariant is right.
- Stray blank line: gone.
- `resolveCountSetting` at three copies: not taken, as offered. I still think the third copy is the moment, but CLAUDE.md endorses the per-field resolver shape explicitly, so it stays a suggestion.
- `executeResumeRun`/`executeExhaustPausedRun` remain untested directly, consistent with `sweep.ts` having no test file. I read the "most recent implement|repair task" selection again: `.sort(asc createdAt).pop()` picks the latest, which for a second resume is the previous *resumed* row, so the lineage chains correctly rather than always pointing at the original pass. The PR-number/branch/kind inheritance is what keeps `shouldOpenDraftPr` from opening a second PR and `checkout: "adopt"` from racing a fresh branch.

## What I verified myself this round

- **Idempotency under two module graphs.** `hasLiveTask` is gathered as a task fact and re-read inside `executeResumeRun` before the insert, which is the right latch given the run deliberately stays `rate_limited` until its pass starts, and given a webhook-triggered sweep runs on the app-router graph (#159).
- **The restart story.** `resumeCount` bumped at decision, `resumeAfter` cleared only in `startTask`, so a restart between the two leaves a visibly paused run; a restart *during* a resumed pass finds the run `implementing` and takes the existing interrupted-run path, no attempt consumed. `ACTIVE_RUN_STATUSES` includes `rate_limited`, so the boot prune keeps a waiting run's transcript, and the queue has no run-status filter that would strand a queued resume.
- **The transfer primitives.** Paths travel in the environment and are quoted, so a session id cannot reach `bash` as syntax; the read is `Tty: true` + `base64 -w0` + a trailing declared size, and a short read is reported as no read. Path escape via a hostile session id would stay inside a container the agent already owns, and the host side validates the run id before joining — no host escape. Both are bounded on `raceWithTimeout` per #151, and both degrade to the declared fallback.
- **Every acceptance criterion has a test**: the reducer's 14 new cases cover eligibility, jitter spread from one shared window, the null-clock case, idempotency, the bound before the clock, bound zero, resume-outranks-claim via `claimableSlots`, and a resume proceeding while the kill switch holds pickup; the store's cover the restart criterion; the settings resolver and the panel render cover "settable in the UI and falls through when unset".

## Non-blocking, for a follow-up

- `awaitExecExit` is the extracted form of a wait-for-exec pattern that already existed twice in `container-manager.ts` (setup, commit-and-push), and this PR adds it as the third rather than collapsing the other two. Migrating them is out of scope for a quota ticket on a gated path, but the helper now exists and the older sites are the ones worth moving.
- `awaitExecExit` resolves 200 ms after the exec stops running, so a large transcript could occasionally lose its trailing size and be discarded as a partial read. It degrades correctly (same branch, context lost); a single re-read on a size mismatch would recover it cheaply.
- The Docker transfer functions have no automated coverage, by admission — the one part of this that is argued rather than demonstrated. Worth a deliberate first walled run being watched.

## For the human merging

Two decisions the author flagged, and I agree they are a human's rather than mine, not blockers:

1. **A resume is not held by the kill switch or the daily cap.** It is consistent with what both mechanisms already mean (they gate pickup; in-flight work is decided above them) and with #171's stated rule that a parked run still resumes, and the runbook says so. The consequence is real: an operator who engages the kill switch can still see a fresh agent container start hours later. If you want the switch to mean "no new containers, full stop", that is a one-line change to the reducer's ordering and should be said out loud.
2. **A fix-up turn walled mid-review-cycle resumes from the original brief** when its transcript did not survive, since the reviewer's feedback was a message rather than the description. Self-correcting through the re-review, narrower than the implement case.

Everything else — the sweep as the only driver, jitter as a hash of the run id, the bound checked before the clock, `exhausted` rather than `failed` so unspent attempts survive a re-arm, the transcript store beside the database with a boot-time prune, and the dashboard/digest showing the *eligible* instant through the same function the reducer decides with — reads as done properly and matches what CLAUDE.md now records.

PR #187, opened 2026-09-02
Interactive overflow to a metered lane, with confirm-then-cap guardrails

### LINKED ISSUE #173: Interactive overflow to a metered lane, with confirm-then-cap guardrails

lennons301:
Closes #173

When the subscription window **refuses** a pass, autonomous work pauses (#168/#169) and attended work **crosses onto a paid lane** — a human is sitting there waiting. That asymmetry is the ticket; everything else here is what makes it safe.

## What's in it

**One pure decision, three consumers** (`src/lib/lanes/overflow.ts`). `decideLaneCrossing` answers, for one pass: which lane it runs on, how its spend is booked, and whether it may start at all. The turn manager routes a pass with it, the queue loop declines to start one with it, and the task screen offers the confirmation from it — the #148 rule, because a gate computed twice eventually disagrees with itself. It owns nothing #174 owns: the cap, the confirm-once gate and cap-outranks-confirmation are evaluated through that ticket's own `evaluateMeteredSpend`/`resolveMeteredCap`.

**Overage as a paid lane.** `overagePaysNow` counts a request that drew on overage *and* a refused window whose overage window is still serving — the state `scripts/rate-limit-stub.mjs --scenario overage-active` reproduces (`status: rejected`, `overage-status: allowed`, HTTP 200: the wall is up, the request succeeded, the card paid). **The ticket said to check this rule before building it, and one part of it did not survive the check:** `overageInUse` alone must *not* count, because the real captured event from a healthy account carries `overageInUse: true` with `status: allowed` and `isUsingOverage: false` — overage billing merely being *available*. Keying off it would have held the fleet, every day, for a confirmation nobody owed. Pinned by a test against that fixture's shape.

That reclassification enters through `readMoneyGuards`, which now reports the **effective** billing kind, so the sweep, the dashboard and the crossing all see it through one mechanism rather than four. Consequence worth stating: on an overage the money guards hold new autonomous pickup too, which is the ticket's "subject to the same guard".

**The at-the-keyboard confirmation.** `MeteredCrossing` on the session screen says what is happening and offers the press, with the numbers that make it decidable. The press is #174's fleet-level one, so the strip says so — autonomous passes may spend on it too for the rest of the day. At the cap there is deliberately **no** button: no press helps before midnight, so the session is told it is capped and pointed at the cap. With nowhere to overflow, the message names the variable that would fix it.

**Held, not failed.** A refused session stays `queued` with the reason on its feed, and the pickup loop **steps over the whole interactive kind** while the guards refuse it. Interactive work sorts first (#15), so one held session at the head would otherwise stop the review passes and resumes behind it — work already paid for, held up by work not yet authorised. Follow-up turns needed nothing new: `laneForFollowUp` already resolved the lane before dequeuing the message, so the owner's turn isn't burnt and it runs on the poll after the press.

**`tasks.tier`** (migration 0028) joins `tasks.lane`/`laneBilling`. A run row has carried lane + tier since #172; interactive work — the only kind that crosses — has no run row, so without it the one figure the fleet couldn't account for was the one it spent real money on.

## Verifying the interactive-cap hole

The ticket asks for #174's closing of the interactive exemption to be verified *on this path*. It couldn't be: the rule lived inside the turn manager's private `updateTask`, reachable only through a container. It's now `bookTaskCost` beside the ledger it writes, with tests: an overflowed session counts against the cash cap while staying exempt from the quota-funded one, a session an active overage pays for counts even though its lane declares itself a subscription, and subscription work counts for nothing however much it spent.

## Self-review (`/code-review` vs `origin/main`, spec #173)

Both axes found real defects; all are fixed in `3ad3a35`:

- **A refused pass was failed, not held** — `requirePassLane` threw from inside `startTask`'s failure boundary, so the manual run route (`POST /api/tasks/[id]/run`, which never passes the queue's check) and the withdrawn-confirmation race failed a task that must wait. The crossing is now read outside that boundary.
- **The feed notice re-posted every turn** — it quotes the day's running spend, so the last-message dedup never matched twice. `payerChanged` is the honest comparison, and it catches what a lane-only one misses: the lane holding still while an overage starts paying.
- **An unknown overage status decided something** — read as "serving", it suppressed the wall and left the session on the walled lane. It now goes through the shared severity map, so an unheard member decides nothing (#171's rule).
- **The "who pays" sentence was built three times**, and the dashboard's copy omitted the subscription condition, so a metered primary observed mid-overage would have read "an active overage on openrouter". `overageIsThePayer` is now the one predicate.
- Presentation split from fetching in the panel (so each state is render-tested), a dead `live` prop removed, `TaskData.kind` typed as the pass-kind union.

One finding I'm answering rather than acting on: **holding interactive dispatch on a metered *primary*, not only on an overflow**, was flagged as mild scope creep. I've kept it deliberately — the guard keys off billing kind rather than off having overflowed, exactly as #174's do, and "the first metered use in a local day is confirmed by the human in the UI" / "at the cap, interactive work is told it is capped" read as facts about metered interactive work, not about the overflow transition. #174's own briefing says the attended confirm UX is this ticket's.

## Known limitation

A session dispatched **from Discord** while the guards refuse gets its explanation on the task feed only; nothing is posted back to the channel. The ticket scopes the confirmation to the UI ("surfaces it where the human already is — in the UI"), and #174's once-a-day Discord ping still covers "the fleet is held on money", so this is left as-is rather than widened here.

## Checks

`pnpm test` 1511 passing (80 files), `pnpm build` clean, `pnpm lint` clean on every changed file (`main`'s pre-existing errors in `preview-pane.tsx` / `output-parser.test.ts` untouched).

---

## Integrated with `origin/main` (#184 / issue #175)

#175 landed under this branch and keyed `quota_state` by **lane**, which changes who may read it: *which* row to read depends on the lane the money guards resolve, so an observation handed in from outside would have made each caller pick the lane — the divergence per-lane keying exists to prevent. `readMoneyGuards` now owns that read and hands the observation back, so the admission gate, the crossing's overage rule and the dashboard's tile judge one instant of one row on the one lane they all resolved.

Two of #175's mechanisms compose with the crossing unchanged, and both improve it:

- a turn's cost is charged from the lane's declared `prices` *before* it reaches `bookTaskCost`, so an overflowed session's cash is booked at what it actually costs rather than at the CLI's Anthropic-list fiction (16.7× on a measured turn);
- `laneFallbackTier` means an overflow onto a priced lane runs at that lane's own `standard` rather than passing no `--model`, because the target is resolved through the same `resolveLane`.

Migration renumbered 0028 → **0029** (`0028` is now `quota_state_per_lane`), journal `when` a day past it. Post-merge: `pnpm test` 1568 passing (83 files), `pnpm build` clean, lint clean on changed files.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/metered-crossing-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/metered-crossing.tsx` |
| visual-ui | `**/components/**` | `src/components/metered-spend-settings.tsx` |
| visual-ui | `**/app/**/page.*` | `src/app/tasks/[id]/page.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0029_task_tier.sql` |
| persistence | `drizzle/**` | `drizzle/0029_task_tier.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0029_snapshot.json` |

lennons301-reviewer (dismissed):
Approving. Reviewed with fresh context against issue #173, the repo's CLAUDE.md conventions, `docs/agents/review-gates.yaml` and the `#171`/`#174`/`#175` sections this ticket sits on. This PR is **gated** (`human-signoff` label, auto-merge off), so this approval satisfies branch protection only — a human still merges.

## Objective signals

- `pnpm test` (fresh `local.db` removed first): **1568 passing, 83 files**, no flakes on the known cross-graph-state / queue-slots specs.
- `npx tsc --noEmit`: clean.
- `npx eslint` over every changed `.ts`/`.tsx` (source and tests): clean. `main`'s known baseline errors are in files this PR doesn't touch.
- No CI checks configured on the branch, so the above was run locally.
- Migration hygiene per CLAUDE.md: `0029_task_tier` journal `when` is exactly one day past `0028`, `0029_snapshot.json` is present and its `prevId` matches `0028`'s `id`, and the DDL is a nullable `ADD COLUMN`.

## Acceptance criteria, checked against the code and its tests

Every one of the ticket's ten boxes is implemented and pinned by a test, most of them at two levels (the pure decision and the surface that consumes it):

- **Autonomous work never overflows** — `decideLaneCrossing` returns early for every non-interactive kind, keeping `walled` as a reported fact while emitting no refusal and no overflow; pinned per-kind in `overflow.test.ts` and end-to-end through the real poll loop in `metered-pickup.test.ts` ("does not hold an autonomous pass").
- **Interactive overflow on a wall** — target is the first *available metered* lane in the file's own `primary` preference order, with declaration order as the tiebreak; the excluded lane is the one being crossed off. Tested for ordering, for falling past a lane whose credential is missing, and for staying put on a healthy window.
- **Refused with a reason when there is nowhere to go** — `no-metered-lane` names the missing variables via `laneMissingEnv` (one reader of availability, not a second reading of `auth`), and the message reaches both the feed and the session panel.
- **Confirm-once in the UI, then automatic, then capped** — all three routed through #174's `evaluateMeteredSpend`/`resolveMeteredCap` rather than restated; cap-outranks-confirmation and yesterday's-confirmation-is-no-confirmation both tested; the panel deliberately offers no button at the cap (render test asserts the absence).
- **Metered interactive spend counts, subscription interactive spend doesn't** — the previously untestable rule inside the turn manager's private `updateTask` is now `bookTaskCost` beside the ledger, with the increment-only semantics preserved and five cases pinned, including the one that matters most here: an overflowed session counts against the cash cap while staying at `0` on `todayAutonomousSpendUsd`.
- **Active overage is a paid lane** — `overagePaysNow` counts `isUsingOverage`, and counts a `rejected` window whose overage window still serves (the `overage-active` stub shape). The ticket asked for this rule to be *checked* before building; the check produced a real correction, and I agree with it: `overageInUse` alone must not count, because the captured healthy-account event carries it with `status: allowed` / `isUsingOverage: false`. Keying off it would have held the fleet daily for a confirmation nobody owed. `overageIsServing` goes through the shared `quotaSeverity` map, so an unheard status decides nothing and the wall stands — the safe direction on both counts, and tested (`throttled_pending_review` -> still walled, session still crosses).
- **Ledger records lane and tier** — `tasks.tier` added and written on pass start and on each follow-up turn, beside the effective `laneBilling`.

## Correctness of the pure module and its callers

- **One decision, three consumers** holds in fact, not just in prose: the turn manager, the queue loop and `/api/settings/metered-spend` all reach `decideLaneCrossing` through `readLaneCrossing`, which is built *on* `readMoneyGuards` rather than beside it. After #175 keyed `quota_state` by lane, moving the quota read into `readMoneyGuards` is the right call — which row to read depends on the lane it resolves, so an observation handed in from outside would have put lane selection back in each caller. The sweep and `rows.ts` now take `money.quota` instead of re-reading, so gate, crossing and tile judge one instant of one row.
- **Held, not failed** is correct on every path I could find into a pass: the crossing is read *outside* `startTask`'s failure boundary (everything inside that `try` fails the task by design), the manual `POST /api/tasks/[id]/run` route therefore also holds rather than fails, and `laneForFollowUp` holds a follow-up turn *before* dequeuing the message so the owner's turn isn't burnt. Verified by reading the queue loop that the early return from `startTask` still releases both `inFlightTasks` and `slotReservations` in the `.finally`, so a held pass leaves no phantom slot behind — the #151/#159 failure this platform keeps meeting.
- **The queue's precedence is right, and tested for the wedge shape.** Stepping over the whole interactive *kind* is exactly the correct width given the hold is fleet-wide for that kind, and putting it ahead of the slot/memory checks matches #171's stated precedence. `metered-pickup.test.ts` drives the real loop with a real lane file, quota row, settings row and ledger and asserts a review pass starts while the held session waits — a genuine regression test, not a mock-shaped one.
- **`payerChanged` is the honest dedup.** An exact-text comparison could not work here (the sentence quotes the day's running spend), and comparing the recorded `(lane, laneBilling)` pair catches the case a lane-only comparison misses: the lane holding still while an overage starts paying. It compares the same pair the ledger keys off, so what is announced and what is charged cannot part company.
- **`resolveLane(laneId)`** is a parameter rather than a second resolver, and everything after `findLane` uses the resolved lane, so the target's own credentials, base URL, tier map and #175 `laneFallbackTier` all apply; `choice` still reports the primary. `primaryLaneOf` and `resolveLane` both go through `choosePrimaryLane`, so passing the crossing's `laneId` explicitly on a non-crossing pass is a no-op rather than a behaviour change — including the unknown-choice and unavailable-primary cases, which still produce #172's report verbatim.
- **Scope judgement I agree with:** holding an interactive session on a *metered primary*, not only on an overflow. The criteria are written about metered interactive work ("the first metered use in a local day…"), #174's briefing explicitly assigns the attended confirm UX here, and keying off billing kind rather than off having overflowed is the same rule #174's own guards follow.

## Findings (none blocking)

1. **The dashboard's new overage wording is unpinned.** No test sets `primaryLaneOverage: true`, so the `payer` branch in `buildFleetView` — "an active overage on X" in `pickupPaused.body` and in both needs-you cards — is never rendered by a test, even though the self-review names that copy as a defect it fixed. I traced the branch by hand and it is correct (the `?? "a metered lane"` fallbacks still behave as before, and `overageIsThePayer` is tested in `overflow.test.ts`), so this is coverage rather than a bug — but it's the one rule in this PR that isn't held down, in a repo whose convention is that each rule is.
2. **`attendedPickupIsHeld` writes to the database.** A predicate-named function that inserts a feed note via `crossingHoldsPass` is mildly misleading at the call site; the docstring discloses it, and `crossingHoldsPass` carries the same shape deliberately so the sentence and its dedup are identical wherever a refusal is met. Naming only — the behaviour is what you want.
3. **The crossing's cap and the dashboard's cap can be different numbers, legitimately.** The crossing resolves the *overflow target's* declared cap while `readMoneyGuards` resolves the *primary's*, so a session overflowing off a subscription lane can be capped at the target's `$20` while the dashboard's metered gauge reports the fleet as not-in-force (primary is subscription). It is documented at the call site and the session's own message quotes its own numbers, so nothing lies; it's an observability seam worth remembering when reading the two surfaces side by side.
4. **A Discord-dispatched held session is explained on the task feed only.** Disclosed in the PR body, consistent with the ticket scoping the confirmation to the UI, and #174's once-a-day Discord ping still covers "the fleet is held on money". Left as-is is the right call for this ticket; a follow-up would be cheap if it bites.

Docs (`CLAUDE.md`, `docs/runbook.md`, the `lanes.yaml` comments) match the code as landed, including the two surprises an operator most needs warning about — an active overage moving the real-money gauge while the lane still says `claude-subscription`, and a metered primary asking on the first interactive turn of the day.

Gate categories matched: `persistence` (`src/db/schema.ts`, `drizzle/**`), `autonomy-control` (`src/lib/orchestrator/autonomy/sweep.ts`), plus the estate default UI globs (`src/components/**`, `src/app/**/page.tsx`) — correctly labelled `human-signoff` with auto-merge off.

lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/__tests__/metered-crossing-render.test.tsx` |
| visual-ui | `**/components/**` | `src/components/metered-crossing.tsx` |
| visual-ui | `**/components/**` | `src/components/metered-spend-settings.tsx` |
| visual-ui | `**/app/**/page.*` | `src/app/tasks/[id]/page.tsx` |
| data-migrations | `**/*.sql` | `drizzle/0030_task_tier.sql` |
| persistence | `drizzle/**` | `drizzle/0030_task_tier.sql` |
| persistence | `drizzle/**` | `drizzle/meta/0030_snapshot.json` |

lennons301-reviewer (approved):
Approving. Fresh-eyes review against issue #173, the repo's CLAUDE.md conventions and the surrounding #167/#168/#171/#172/#174/#175 machinery. This PR is **GATED** (`human-signoff`, auto-merge off), so this approval satisfies branch protection only — a human still merges.

## Objective signals

No CI checks reported on the branch, so I ran them locally after `rm local.db*`:

- `pnpm test` — **1625 passing, 85 files**, exit 0
- `pnpm build` — exit 0
- `eslint` on all 27 changed `.ts`/`.tsx` files — clean (main's known baseline errors in `preview-pane.tsx` / `output-parser.test.ts` untouched)

## Acceptance criteria — each traced to code and a test

| AC | Where |
|---|---|
| Autonomous passes never overflow; they pause | `decideLaneCrossing` returns `base` before the overflow branch for any non-`interactive` kind (`isAttendedPass`); `refusal` is structurally unreachable for them. Tests: *"autonomous passes never overflow"*, and end-to-end through the real queue loop in `metered-pickup.test.ts` (*"does not hold an autonomous pass"*) |
| Interactive work overflows when the primary is walled | `laneIsWalled` + `meteredOverflowCandidates`, honouring the file's own `primary` preference order so the card charged stays a version-controlled decision |
| Refused with a reason when nowhere to overflow | `describeUnavailableMeteredLanes` — built on `laneMissingEnv`, so "what makes a lane unavailable" still has exactly one answer (#172) — surfaced on the feed *and* in the panel |
| First metered use confirmed in the UI via #174's fleet gate | `MeteredCrossing`/`CrossingPanel` + `evaluateMeteredSpend`; the strip states plainly that the press is the fleet's |
| Rest of the day automatic; capped rather than silent | `money.hold` precedence taken straight from #174; at `cap-reached` there is deliberately **no** button |
| Metered interactive spend counted against the cash cap | `bookTaskCost` inside the `updateTask` funnel |
| Subscription interactive work stays exempt | early return at `billing === "subscription"`, pinned by *"counts nothing for subscription work, whatever it spent"* |
| Active overage classified as a paid lane, same guard | `overagePaysNow` → `effectiveBilling` → `readMoneyGuards.billing`, which the sweep snapshot and `loadFleetRows` now both take |
| Ledger records lane and tier | `tasks.tier`, migration `0030_task_tier` |

## What I checked hardest

**The overage rule actually was checked, and one half of it did not survive.** The ticket said *"stated rather than confirmed in #164 — check it before building"*, and the PR reports that `overageInUse` alone had to be dropped because a real captured healthy-account event carries it with `status: allowed` / `isUsingOverage: false`. Keying off it would have held the fleet daily for a confirmation nobody owed. That is the ticket being executed rather than assumed, and it is pinned by *"does not count overage merely being available on a healthy account"*.

**No second opinion about money.** `decideLaneCrossing` re-runs #174's own `resolveMeteredCap`/`evaluateMeteredSpend` for the lane it picks rather than restating the policy; the queue's `attendedPickupIsHeld`, `startTask` and the API route all funnel through the same `readLaneCrossing`. `readMoneyGuards` now owns the per-lane quota read, which is the right consequence of #175: `sweep.ts` and `rows.ts` both drop their own `getQuotaObservation` call, so the lane whose window is judged, whose spend is capped and whose overage is read can no longer be three lanes. I verified `primaryLaneOf` and `resolveLane` share `choosePrimaryLane`, so `crossing.laneId` is the same lane `resolveLane` would have picked unaided — the new `laneId?` parameter only ever *overrides* on a genuine overflow.

**Fail-safe directions.** `overageIsServing` goes through `quotaSeverity`, so an unheard status word decides nothing and the wall stands (#171's rule, and the direction that crosses off the walled lane rather than sitting on it). An *unavailable* lane is correctly not a wall. `laneIsWalled` is false for a metered lane and for a null/spent observation, so silence never closes anything. `bookTaskCost` reads the stored `laneBilling` (already the *effective* kind), so no overage opinion is duplicated in the ledger.

**Held, not failed.** The self-review caught the real defect here: `requirePassLane` throwing inside `startTask`'s failure boundary would have failed a task reachable via `POST /api/tasks/[id]/run` without passing the queue's check. The crossing is now read outside that `try` and answered by `crossingHoldsPass`. I traced the queue's `.finally` — an early return still releases `inFlightTasks` and `slotReservations`, so a held pass leaks no slot.

**`nextQueuedTask(skipInteractive)`** is the right width for the hold: the guard is fleet-wide for interactive passes, so skipping the kind (rather than the row) is exactly what stops one unpressed session at the head of a kind-first queue from starving review passes and resumes behind it. Exercised through the real loop, not a re-implementation, in *"never holds the work behind it — a review pass starts instead"*.

**`payerChanged`** is the honest dedup — a text comparison could never match twice against a sentence quoting running spend. I confirmed `processQueuedMessages` re-reads `task` at the top of each drain iteration, so the recorded pair is fresh on iteration 2 and the notice does not re-post mid-drain.

**Security / conventions.** No secret is stored or served: `env` reaches `decideLaneCrossing` for `laneIsAvailable`/`laneMissingEnv` only, and refusal text carries variable *names* — the same rule #172 set. The `metered-spend` route serves numbers, lane ids and billing kinds. No `globalThis` state is needed here (everything is DB- or file-derived), so #159's two-graph trap is sidestepped rather than papered over. Migration `0030` is a single `ADD COLUMN`, journal `when` bumped exactly one day past `0029`, and I verified `0030_snapshot.json`'s `prevId` chains to `0029`'s `id` with `tasks.tier` as the only diff — the CLAUDE.md rule that keeps the *next* author's `drizzle-kit generate` honest.

**Smell baseline:** nothing I'd act on. `CrossingLane` is deliberately structural rather than a fourth lane type; the three "who pays" sentences share one predicate (`overageIsThePayer`) while keeping surface-appropriate prose, which is the right side of the duplication line. `overflow.ts` is long but cohesive and the docs carry their own argument.

## Non-blocking observations (suggestions, not requests)

1. `tasks.tier` is written by both write sites but read by no surface — it is a ledger column by design (CLAUDE.md says so), yet unlike `TaskListRow` nothing type-checks it onto a screen, so a rename would go unnoticed until someone queried it.
2. While the primary lane is a healthy-looking subscription, `MeteredSpendPanel` hides its confirm strip (`state.metered` is false), so during an *overflow* the only UI route to the press is the held session's own panel. That is the ticket's intent, and the runbook documents the flow — worth an operator knowing that the settings screen is not a fallback there.
3. `noteOnceOnFeed`'s exact-text dedup can admit a second note on a long-held session if another in-flight run moves the day's spend figure quoted in the message. Bounded by turn completions, not a storm.
4. `attendedPickupIsHeld` is evaluated before the `!inFlightTasks.has(next.id)` check, so a narrow race could write a refusal note onto an interactive task already dispatched. Harmless, and it fails in the direction of not blocking.

The self-review is unusually honest — five real defects found and fixed in `3ad3a35`, plus one finding argued rather than silently dropped (holding a metered *primary*, which I agree with: #174's own briefing hands the attended confirm UX to this ticket, and keying off billing kind rather than off having overflowed is the rule that ticket set). The Discord-dispatch limitation is correctly scoped out and stated.

PR #192, opened 2026-09-02
Cross-lane cost routing: pick the cheapest capable lane, and fail over instead of pausing

### LINKED ISSUE #176: Cross-lane cost routing: pick the cheapest capable lane, and fail over instead of pausing

lennons301:
Closes #176

Cost routing and wall-triggered failover, as the ticket's two halves sharing one
pure ranking.

## Cost-first selection

`src/lib/lanes/lane-selection.ts` is the ranking, and it is the *only* place
lane choice is decided — the crossing routes a pass with it, the reducer decides
a failover with it, and the settings screen shows what it would pick, for the
#148 reason that a selection computed twice eventually disagrees with itself.

Two orderings come out of one price table, because cost and capability are
different axes and only one of them is what the fleet pays:

- **Cost** is the marginal cash rate. A subscription lane is **zero** — its
  quota is already bought — which is why routing is a *no-op in the shipped
  configuration until a wall*, and that property is what made it safe as a
  default. A priced lane costs its own blended rate; a lane billing per token
  with no declared `prices` is *unread* and ranked **last**, the cautious answer
  `lane-cost.ts` already argues for. An active overage moves a subscription lane
  into that class too: the card is being charged at a rate nothing writes down.
- **Capability** is what the provider charges for what the lane runs, with an
  unpriced lane at the *top* (it is Anthropic-direct, running first-party
  Claude — the same fact that puts it last on cost, read back). This axis exists
  only to give a pass kind's **minimum lane** a meaning: a floor is inclusive,
  so a floor of `openrouter` still allows the free subscription above it, and a
  floor of `anthropic-api` reads as "first-party Claude only".

Ties break on the file's own `primary` order — #173's reviewed overflow ranking,
surviving as the tie-break rather than the rule.

The rate is a **ranking key, not a forecast**: `RANKING_PASS_USAGE` is a stated
1M-token mix (cache-read dominant, off #175's captured turns) priced through the
same `priceTokens` the ledger charges from. Nothing here is booked against a
budget.

## Wall-triggered failover

`failOverRunLane` is asked *after* #170's tier ladder (a rung down is free — the
account still has quota — where a move may cost real money and will certainly
cost a fresh conversation's worth of orientation) and *before* #168's pause,
which becomes the fallback. It spends neither an attempt nor an interruption,
shares `endRefusedPass`'s ordering, and — unlike the degrade — **carries the
session**, so a move is lossless the way #169's resume is (and only when the
transcript actually landed, which is why `queueNext` is now handed `preserved`).

The reset-less wall is the case this helps most: a pause needs a clock and had
none, so before this such a pass took its ordinary path and **spent one of the
ticket's three attempts**.

The target lane is advisory, not a pin: the replacement pass re-asks the same
ranking when it starts, so a window that lifted in the intervening half-minute
sends it back to the cheaper lane.

## What routing may not do

- **Spend cash the fleet is not permitted.** #174's cap and confirm-once press
  are evaluated *inside* the ranking, per lane against that lane's own declared
  cap, so a money-held lane is simply not a candidate — and the *fall-back* goes
  through the same evaluation, because when the ranking picks nothing the pass
  still runs on the lane in force.
- **Route around an unavailable lane.** A missing credential is #172's report;
  papering over a misconfiguration by spending at another provider is what that
  ticket exists to refuse. The pass stays where it was sent and dies naming the
  variable.
- **Override a pin.** An explicit `primaryLane`/`AGENT_LANE` choice turns the
  ranking off — the escape hatch, and deliberately not a new setting, since #172
  already distinguishes a decision from the unset default that walked the
  preference order. A **wall releases it**, because #173 crossed an attended
  session off a walled lane whether or not it was pinned and that has to keep
  being true: a pin says "do not choose for me", where a wall says the chosen
  lane cannot serve the request at all.
- **Become a route around a human stop.** The kill switch and the per-project
  toggle gate pickup exactly as before; a failover is in-flight continuation,
  like #169's resume.

## Acceptance criteria

- [x] Selection is a pure function of (available lanes, pass kind, resolved
  tier, settings, quota and spend state), table-tested with no Docker or network
  — `lane-selection.test.ts`, 36 cases
- [x] A pass runs on the cheapest available lane that can serve its pass kind
  and tier — `decideLaneCrossing` routes through the ranking
- [x] Pass kinds carry a configurable minimum lane, settable in the settings UI
  — four `minLane*` fields, `Settings → Execution lane`
- [x] A walled run prefers moving to another permitted lane over pausing
- [x] Autonomous failover onto a metered lane respects #174's cap and confirm
  gate — three executor tests pause instead when either holds
- [x] A lane emitting no quota telemetry is admitted on spend, never on another
  lane's utilization — `getQuotaObservations` keyed by lane; absent reads as null
- [x] A lane move reuses #169's transcript restore and costs neither an attempt
  nor an interruption
- [x] Lane moves are bounded by the existing quota-pause counter
  (`runs.resumeCount` against `maxResumesPerAttempt`)
- [x] Pausing remains the fallback when no lane is both available and permitted
- [x] The fleet can be pinned to a single lane from settings, disabling cost
  routing
- [x] The kill switch and per-project autonomy toggle still stop all work
- [x] The run ledger records the lane each pass actually ran on, including after
  a mid-run move — `startTask` writes `runs.lane`/`tasks.lane` at pass start, and
  a move is a fresh pass start

## Deliberate behaviour changes a reviewer should look at

1. **An overflow target changes on the shipped file.** #173 sent a walled
   attended session to `anthropic-api` (preference order); it now goes to
   `openrouter-glm` (~40x cheaper). Where the floor excludes GLM it goes to
   `openrouter` rather than `anthropic-api`, even though the two charge the
   same, because only one of them writes its price down. Setting a kind's floor
   to `anthropic-api` restores first-party-only routing for it.
2. **Autonomous work now crosses onto a paid lane.** #173's asymmetry was
   "interactive overflows, autonomous pauses"; the ticket asks for exactly this
   change, and it stays bounded by #174's cap and confirm-once press — the most
   common reason a run still pauses is that nobody confirmed the day's spend.
3. **A pass is routed off an overage-paying subscription lane** onto one whose
   price is declared and capped. The notice and the refusal still name the
   overage as the cause, because that is the news the operator needs.

## Known boundary, not widened here

#171's admission gate and #174's pickup guards still read the **primary** lane's
row rather than the lane routing would pick, so new *pickup* under a walled
primary still waits where an in-flight run fails over. The gate is fleet-wide by
construction and the snapshot carries one observation; making it per-pass-kind
needs its own ticket, and no acceptance criterion here asks for it.

## Validation

- `pnpm test` — 87 files, 1703 tests, all passing (baseline was 1625)
- `npx tsc --noEmit` clean; `pnpm build` succeeds
- `npx eslint` clean on every changed file (the repo's known baseline errors in
  `output-parser.test.ts` are untouched)
- **Not verified live:** no lane move was exercised against a real provider —
  the OpenRouter account holds no credits (#175's finding) and Docker is
  unavailable here. To finish the proof once funded, arm a ticket with
  `scripts/rate-limit-stub.mjs --scenario exhausted` in front of the
  subscription lane and confirm the day's spend: the run should comment "Moved
  execution lane" and retry on `openrouter-glm` rather than going
  `rate_limited`.

## Self-review

Ran `/code-review` against `origin/main` with #176 as the spec. It found three
real defects, all fixed in `91714ee` with regression tests, plus the doc
contradiction they caused:

1. the fall-back skipped the money guards (a way onto a paid lane unconfirmed
   and past the cap);
2. a pin disabled #173's interactive overflow outright;
3. the minimum lane was silently ignored on fall-through while the settings row
   claimed the pass would be refused.

Nothing was left unaddressed.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/lane-routing-settings.tsx` |

lennons301-reviewer (approved):
Approving. Reviewed against issue #176's acceptance criteria, `CLAUDE.md`, `docs/agents/review-gates.yaml` and the #170/#171/#173/#174/#175 conventions this ticket builds on. This PR is **gated** (`human-signoff`, auto-merge off): my approval only satisfies branch protection — a human still looks and merges, and there are three things below I'd want their eye on.

## Objective signals

No CI runs on this branch, so I ran everything locally on `agent/issue-176`:

- `pnpm test` — 87 files, 1703 tests, all passing (after `rm local.db*`, per the migration gotcha in `CLAUDE.md`)
- `npx tsc --noEmit` — clean
- `npx eslint` on all 26 changed `.ts`/`.tsx` files — clean (baseline errors elsewhere untouched)
- `gh pr view` — `mergeable: MERGEABLE`, blocked only on review

## Ticket coverage

Every acceptance criterion is met, and the ones that could have been faked are not:

- **Pure, table-tested selection.** `lane-selection.ts` takes `(catalog, env, kind, tier, pin, floor, observations, config, overrides, spend, confirmedAt, now)` and returns the ranking whole; 42 cases in `lane-selection.test.ts` with no Docker and no network.
- **One ranking, three callers.** `decideLaneCrossing` routes with `selectLane`, `planLaneFailover` is a thin read of the same function with the refused lane excluded, and `routingState` in the overrides route reads it for the screen. I checked the #148 property holds end-to-end: `startTask` hands `crossing.laneId` to `requirePassLane` → `resolveLane`, and both `readLaneCrossing` and `requirePassLane` are called with the same `ticketModel` at both call sites (lines 341/365 and 1034/1037), so the tier a lane is *ranked* at is the tier the pass *runs* at.
- **Money guards inside the ranking, including the fall-back.** `evaluateMeteredSpend`/`resolveMeteredCap` are called per candidate against that lane's own `caps.daily_budget_usd`, and `selection.inForce` is a judged candidate — so the fall-back path carries its money state rather than `null`. This was the first of the three self-review fixes in `91714ee` and it was a real hole; it is closed and pinned by `overflow.test.ts`'s "falling back to the lane in force never skips the money guards".
- **Failover ordering and bound.** In `decideNext` the order is degrade (#170) → `failOverRunLane` (#176) → pause (#168), each `continue`ing, all ahead of the blocked-marker and empty-pass readings — verified by reading the loop, not just the tests. The bound is `pass.resumesMade < snapshot.maxResumesPerAttempt` with `resumeCount` bumped in `endRefusedPass`'s `runPatch`; neither `attempt` nor `interruptionCount` moves.
- **Losslessness, honestly conditioned.** `queueNext(preserved)` is the right change: the retry carries `sessionId` only when the transcript actually landed, which is the difference between a resumed conversation and a pass that dies inside `--resume`.
- **Per-lane telemetry.** `getQuotaObservations()` keys by `quota_state.lane`; an absent row reads as null and such a lane is admitted on spend, never on the subscription's wall. `laneIsWalled` still returns false for any metered lane, so a leftover subscription observation cannot be held against one.
- **Escape hatch.** The pin is `primaryLaneInForce().pinnedLaneId`, non-null only when an override or `AGENT_LANE` names a *declared* lane (`choosePrimaryLane`'s explicit layers) — a preference-order answer stays a default. The wall-releases-the-pin rule is decided from the turn's own rejection in `planLaneFailover` rather than from a second read of the row, which is the right fresher fact to trust.

The executor tests are the strongest part of the change: `lane-failover.test.ts` drives the real pass-completion seam over a migrated in-memory DB against the real `lanes.yaml`, stubbing only Docker, GitHub and the transcript file — so "pauses instead when the day's cash is unconfirmed / the cap is spent / the floor excludes every target" are assertions about the ledger, not about a mock.

## For the human's eye before merging

1. **The default routing target for a walled autonomous implement pass is now `openrouter-glm`.** Two shipped facts combine: `anthropic-api` declares no `prices`, so it ranks *last* on cost as "unread"; and the minimum lane resolves through two layers to `null` — no floor — on a fresh deployment. So with `OPENROUTER_API_KEY` provisioned and the day's cash confirmed once, a walled subscription sends implement work to GLM open weights rather than to Anthropic-direct. The reasoning is written down and the control exists (`minLaneImplement`, or `AGENT_MIN_LANE`), and the ticket does ask for cheapest-capable with the floor as the knob — but it is a quality/provider shift that arrives by default, and #175's own note says the GLM slugs are verified only against OpenRouter's listing, never exercised. Worth deciding whether a floor is set at deploy time.
2. **A lane move proceeds while the global kill switch is engaged.** `decide.test.ts`'s "never becomes a route around a human's stop" asserts pickup is refused *and* the move still happens. That is exactly the documented kill-switch semantics ("no pickup actions at all … nothing else changes: in-flight runs … carry on") and exactly #169's resume precedent, so I'm not treating it as a defect — but the ticket's wording is "still stop everything", and a move is the one continuation that can newly begin spending cash (bounded by the confirm-once press already given that day and the cap). A human should confirm that reading is the intended one.
3. **The branch is two commits behind `main`** (#188 `adoptParkedContainer`, #191). GitHub reports no conflict and the two touch different parts of `turn-manager.ts`, but my green suite was the branch alone — worth letting CI/the merge tree re-run before landing.

Also noting the PR is honest about what it could not prove: no live lane move was exercised (no OpenRouter credits, no Docker here), and it states the exact command to finish the proof. That matches what #175 attempt 1 did and the driver briefing asked for.

## Minor, non-blocking

- `CLAUDE.md` line 64 lost a sentence boundary: "…rather than getting a second one to keep true A pass refused on a *tier-scoped* window never reaches it (issue #170)".
- `src/app/api/settings/overrides/route.ts` imports `getFleetSettings` twice from the same module, the second time aliased as `readFleetSettings`.
- Two restatements of the min-lane registry outside it: `MIN_LANE_KEYS` in `settings-overrides.tsx` and `KIND_BY_FIELD` in `lane-routing-settings.tsx` duplicate `MIN_LANE_FIELD_ORDER` / `MIN_LANE_FIELD_BY_KIND`. A fifth field added later would silently route its save error to the tiers panel — the repo's own `chipColumns` convention (derive the mapping from the map that classifies) would cover it.
- `docs/runbook.md` says "Three practical consequences" above four bullets.
- `routingState` calls `readMoneyGuards` + `getQuotaObservations` once per pass kind (4x per settings response) where one read would do; harmless on SQLite, but the module's own argument for sharing one read applies.

No smells worth raising beyond those: the extraction of `lane-wall.ts` out from under `overflow.ts`/`money-state.ts`/`lane-selection.ts` is the right direction (and the re-export keeps callers naming the policy module), `meteredOverflowCandidates` was removed rather than left dead, and the one place I looked hardest for Speculative Generality — `capabilityRank` as a second axis — is load-bearing: without it the minimum lane has no vocabulary.

PR #205, opened 2026-09-04
Parked runs re-check lanes on every sweep

### LINKED ISSUE #199: Parked runs re-check lanes on every sweep
## Parent

Spec: #196

## What to build

A run parked on a quota wall stops being stranded beside a lane that could serve it.

Today the reducer's paused-run path resumes on exactly one condition: the run's clock expiring. So authorizing a paid lane a minute after a run parked re-evaluates nothing — the run sits out the window regardless. With one container slot and a dependency chain, a run parked for five hours beside a lane costing a fortieth as much stalls the whole frontier.

Give that path a second condition: if a lane other than the walled one can now serve the run, it resumes there instead of waiting.

An early lane resumption is a lane move like any other. It counts against the same resume bound, so a run still walks its lanes and then exhausts to a human rather than thrashing. It answers to the same daily confirmation and cap, so unattended work still never starts spending real money on its own. And it is announced with the lane it moved to and what that lane costs — a crossing is never silent.

The existing wall ordering is unchanged: tier degrade first, then lane failover, then a clock-driven pause.

Also record a stated limit while in this area: every declared lane currently runs the same harness adapter, differing only in endpoint, credential variable and model identifiers, which is why a lane move can carry the session across providers. A move across two different adapters could not carry it and would have to fall back to starting again on the branch with work already pushed.

## Acceptance criteria

- [ ] A paused run whose wall still stands, and which no other lane can serve, stays parked
- [ ] A paused run that another lane can now serve resumes there rather than waiting out its clock
- [ ] Authorizing a paid lane takes effect at the next sweep tick for an already-parked run
- [ ] An early lane resumption counts against the same resume bound a clock-driven resume does
- [ ] At the resume bound, the run exhausts to a human rather than moving again
- [ ] The money guards hold an unconfirmed or capped crossing, so unattended work does not start spending real money on its own
- [ ] The crossing is announced with the target lane and its cost
- [ ] The existing wall ordering — tier degrade, then lane failover, then pause — is unchanged
- [ ] The cross-adapter session limit is written down where a future adapter author will find it

## Blocked by

lennons301:
Closes #199

## What

A run parked on a quota wall stops being stranded beside a lane that could serve it. The reducer's paused-run path gains a second resume condition: while the wall still stands by its clock, a lane other than the walled one that can serve the run **now** resumes it there (`resumeRunOnLane`) instead of waiting the window out. Authorising a paid lane after the pause (confirming the day's spend, raising the cap, lifting a floor) takes effect at the next 30-second sweep.

## How

- **Reducer** (`decide.ts`): `PausedRun` carries the walled lane and a `laneFailover` option; the new branch sits after the bound check and before the clock check. It counts against the same resume bound, reserves a slot like any resume, and is not held by the kill switch or the daily cap. The wall ordering of a *finishing* pass (degrade, then failover, then pause) is untouched.
- **Gatherer + executors** (`paused-runs.ts`, new, extracted from the sweep so they run over a real DB with the real `lanes.yaml` under test): the option is ranked through the same `readLaneFailover` a refused pass is offered at its wall, with the walled lane excluded. The money guards (cap, confirm-once) are evaluated *inside* that ranking, so an unconfirmed or capped lane is never offered and a null option means "stay parked". The early resume shares the clock-driven resume's body (lineage, session carry, `resumeCount` bump, run left visibly paused until the pass starts) with the lane move's prompt in front, and announces the lane and what it costs on the issue.
- **Shared cost sentence** (`src/lib/lanes/lane-rate.ts`, new leaf): one formatter and one sentence read by the settings screen's routing row, #176's failover comment and the early-resume comment, so the rate is quoted the same way everywhere. The failover action now carries the rate it announces.
- **Stated limit**: every declared lane runs the one `claude-code` adapter, which is why a lane move can carry the session; a move across two adapters could not and would fall back to restarting on the branch. Written at the foot of `harness/adapter.ts`, pointed to from `registry.ts`, `lanes.yaml`, `session-transcript.ts` and `restoreSessionTranscript` (the seam that would enforce it), plus CLAUDE.md and the runbook.

## Judgement calls worth a look

- **Jitter floor.** The early resume fires only while the wall stands by more than the jitter window (5 min by default). Once the window has reset the run's own lane is free again, and a wall lifting inside that window is not worth paying a lane move to skip. Strictly, the AC says a run "resumes there rather than waiting out its clock"; a run under five minutes from reset waits. Pinned by a test; easy to drop if you disagree.
- **Target is advisory**, exactly as #176's failover: `startTask` re-asks the ranking, so a wall that lifts first sends the pass back to the cheaper lane. The issue comment says so and `tasks.lane` is the record.
- **Announcement channel** is an issue comment plus the orchestrator log, matching the failover precedent. No Discord embed: Discord stays push-only for things that need a human, and a lane move needs nothing.

## Self-review findings not acted on

From the `/code-review` pass against `origin/main`:

- *Data clumps*: the lane fields are flattened onto `resumeRunOnLane` rather than nesting `LaneFailoverOption`, and `readLaneFailover` gained a sixth positional parameter. Both follow the existing `failOverRunLane` / `readLaneFailover` shape; changing one without the other would be the inconsistency.
- *`resume/maxResumes` vs the failover's `move/maxMoves`*: kept, because the executor shares the resume's body and the comment reads "resume k/max".
- *The limit is restated in several files*: `adapter.ts` is canonical and the rest are pointers, which is where an adapter author, a lane author and the seam's maintainer each look first.

## Testing

- `decide.test.ts`: new describe for the early lane resume (stays parked, resumes elsewhere, bound first, exhausts at the bound, ordinary resume once reset, jitter floor, clockless row, live task, slot reservation, kill switch / daily cap, same-tick wall ordering).
- `paused-runs.test.ts` (new, DB-backed, real `lanes.yaml`): the gather offers nothing while unconfirmed and offers GLM once confirmed; nothing when capped, when no other lane has a credential, or when the floor excludes them; a pin to the walled lane is released, a pin to a third lane is honoured; ranks at the run's actual tier; both executors' writes and the announcement's wording.
- Full suite: 92 files, 1764 tests pass. `tsc --noEmit` clean; eslint clean on every changed file.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/lane-routing-settings.tsx` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/__tests__/paused-runs.test.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/paused-runs.ts` |

lennons301-reviewer (approved):
Approving against #199. This PR is gated (`human-signoff`, auto-merge off; paths match `autonomy-control` and the estate `visual-ui` gate), so this approval satisfies branch protection only — a human merges.

**Objective signals** (no CI reported on the branch, so run locally from the worktree): vitest on the touched areas — 60 files, 1106 tests pass; eslint clean on all 12 changed source/test files; `tsc --noEmit` clean. Branch contains `origin/main` HEAD; mergeable.

**Against the ticket's acceptance criteria**, each verified in code and pinned by a test:
- stays parked when no other lane can serve it — `laneFailover: null` is "stay parked", no second guard read (`decide.ts` paused loop)
- resumes elsewhere while the wall stands — new `resumeRunOnLane` branch, ordered *after* the bound check and *before* the clock check
- authorising a paid lane takes effect next tick — `gatherPausedRuns` re-ranks every sweep off the tick's single `readMoneyGuards`; the DB-backed `paused-runs.test.ts` flips `meteredSpendConfirmedAt` and sees the offer appear
- same resume bound; exhausts rather than moving at the bound — shared `queueResumedPass` bumps `resumeCount`; bound judged first
- money guards hold unconfirmed/capped — evaluated inside `planLaneFailover` -> `selectLane` (`cap-reached` / `unconfirmed`), covered against the real `lanes.yaml`
- announced with lane and cost — `describeLaneCost` quotes the ranking's own $/Mtok or says the lane declares no prices
- wall ordering of a finishing pass unchanged — completed-pass path touched only to carry `toLaneRateUsdPerMTok`; the pre-existing `parkedRunIds` skip covers the same-tick case
- cross-adapter session limit written at `harness/adapter.ts` (canonical), pointed to from `registry.ts`, `lanes.yaml`, `session-transcript.ts`, `restoreSessionTranscript`, CLAUDE.md and the runbook

**Consistency:** the gatherer ranks through the same `readLaneFailover` the turn manager calls at the moment of the wall, passing `run.model` exactly as the turn manager does, so the two paths cannot disagree (#148 rule). `lane-rate.ts` is a true leaf (`import type` only), safe for the client component. Reducer stays pure; extracting gatherer + executors to `paused-runs.ts` is what makes the real-DB/real-`lanes.yaml` test possible. Self-review findings not acted on (flattened lane fields on the action, 6th positional param) are consistent with the existing `failOverRunLane` shape — I agree with leaving them.

**For the human merging, two deliberate behaviour choices to confirm:**
1. **Jitter floor.** A run within `resumeJitterMs` (5 min default) of its reset does *not* move lanes — a stated narrowing of AC "resumes there rather than waiting out its clock", argued in the PR body and pinned by a test. Worst-case added wait is under 2x jitter. I find it sound; it is the one place the PR reads the ticket non-literally.
2. #176's failover comment now **always** carries a cost sentence (previously silent for a subscription target) and quotes a rate for priced lanes. Benign and in the spirit of "a crossing is never silent"; `lane-failover.test.ts` only asserts the comment's lead, so the rendered wording there is pinned via `describeLaneCost`'s own tests rather than at that call site.

Suggestion only (not tied to the ticket): a one-line assertion in `lane-failover.test.ts` that the failover comment contains "per million tokens" would pin the shared sentence at both call sites.

PR #227, opened 2026-09-05
Widen the harness contract and drive the orchestrator through a fake adapter

lennons301:
Closes #214

## What this does

Widens the harness adapter contract (`src/lib/harness/`) to what a second harness needs and drives the orchestrator through a fake adapter under test, with no change to what the fleet does on the Claude lane.

- **Contract.** An adapter now declares `image`, static `capabilities` (`userInvokedSkills`, `quotaTelemetry`, `reportsCost`, `sessionResume`), `composeSkillInvocation`, `sessionArtifactPaths` and `mapEffort` (null = no equivalent, flag omitted). `TurnResult` carries a **normalised outcome** — `completed | turn-limit | refused { quota | auth | other, resumeAfter, limitType } | failed` — beside the vendor's verbatim `terminalResult`, which stays only for the stream recorder.
- **Orchestrator reads the outcome only.** Attempt exhaustion is `outcome.kind === "turn-limit"`; the wall ordering (degrade → failover → pause) reads `refused { quota }` through one helper, `quotaRefusalOf`; the review-verdict and triage-exit readers take a `PassTurn` (`outcome` + `finalMessage`) rather than re-parsing a stream. `rate-limit-rejection.ts` is gone; no module outside `src/lib/harness/claude-code/` reads `error_max_turns`, `is_error` or an API status (the only remaining mentions are comments in `stream-recorder.ts`).
- **Claude Code adapter is a directory.** `claude-code/index.ts` implements every member; `stream-parser.ts` is the parser moved from `orchestrator/output-parser.ts`; `outcome.ts` holds `classifyClaudeExit` (one API-error predicate, the pre-#214 one) and `readFinalMessage` (one final-message rule shared by the parser and the test bridge); `image.ts` is a leaf the image builder reads.
- **Descriptor table.** `descriptors.ts` (id + capabilities, no imports) is what the pure lane parser validates adapter ids against; `descriptors.test.ts` pins it to the registry both ways, and `registry.ts` types its production table `Record<HarnessAdapterId, …>` so registering without describing is also a compile error.
- **Fake adapter.** `src/test/fake-harness.ts` is registered only under test and never described in the production table, so `lanes.yaml` can never name it. `fake-harness-turn.test.ts` runs the real `startTask` over a real DB and a one-lane catalog: completion, `refused { quota }` reaching the reducer as a pause and as a degrade, turn-limit exhausting the attempt, a null outcome interrupting, and a non-quota refusal taking the ordinary path.

[...]

PR #228, opened 2026-09-05
Skills reach the agent image through the open Agent Skills standard, pinned at build

### LINKED ISSUE #215: Skills reach the agent image through the open Agent Skills standard, pinned at build
## Parent

Spec: #213

## What to build

The estate's generation and workflow skills reach the agent image through the open Agent Skills standard, pinned, at image build — replacing the Claude-plugin install that runs at every container setup today.

Today the container setup script adds the `mattpocock/skills` marketplace and installs the plugin with the Claude CLI, unpinned, on every container start, and parses the resolved version out of the setup output. That is Claude-only by construction, costs wall-clock on every start, and lets an upstream skills release change a pass mid-day.

After this ticket the agent Dockerfile installs `mattpocock/skills` with the Agent Skills installer at global scope for the Claude Code agent, pinned to a ref recorded in the Dockerfile, so the canonical skill folders live in the standard location and the Claude Code skill directory links to them. The plugin steps leave the setup script. The pinned ref is stamped on the image as a label; the orchestrator lifts it from the image into the run ledger's existing skills-version column and into the task feed's system note, so the version stays traceable. Because a skill installed this way is a plain skill rather than a namespaced plugin, every session skill the schema names (grill-me, grill-with-docs, triage, to-spec, to-tickets, wayfinder) and every vendored workflow name must resolve in the image — the build asserts it, so a `workflow:<skill>` label or a generation session can never name a skill the container does not have (the estate's capability-contract rule).

## Acceptance criteria

- [ ] The agent Dockerfile installs the skills with the Agent Skills installer, pinned to an explicit ref, at global scope for the Claude Code agent
- [ ] The build fails if any schema-declared session skill is missing from the installed set
- [ ] The container setup script no longer runs any `claude plugin` command, and setup is measurably faster on a warm image
- [ ] The pinned skills ref is an image label; the run ledger's skills-version column and the task feed's system note carry it for every run and session
- [ ] A `workflow:tdd` implement pass and a `/grill-me` generation session behave exactly as before on the Claude lane
- [ ] The image builder's staleness check rebuilds when the pinned ref changes
- [ ] AGENTS.md's description of how skills reach a container is updated

## Workflow

model: standard

## Blocked by

lennons301:
Closes #215

## What changed

- **`Dockerfile.agent`** installs `mattpocock/skills` with the Agent Skills installer, pinned by the `/tree/<tag>` URL to `SKILLS_REF=v1.2.3`, at global scope for the Claude Code, Codex and OpenCode agents. Canonical folders land under `/home/node/.agents/skills/<name>/SKILL.md`; `/home/node/.claude/skills/<name>` links to them for Claude Code. The installer itself is pinned too (`SKILLS_INSTALLER_VERSION=1.5.23`) so a release of the tool cannot change what the build does.
- **The build asserts the required set**: every schema-declared session skill plus every vendored workflow (`REQUIRED_SKILLS`) must be present, linked for Claude Code, and recorded at the pinned ref in the installer's lock file. `image-builder.test.ts` pins that list to `SESSION_SKILLS` and `docs/agents/workflows/*.md` so the two cannot drift.
- **The pinned ref is an image label** (`co.interlude.agent-skills-ref`), stamped by the Dockerfile from the same build arg. `ensureImage` reports it off the inspect its staleness check already makes (no second daemon round trip); `createWorkspaceContainer` carries it on the handle, and the turn manager writes it to `runs.skillsVersion` (first-write-wins, as before) and the feed's system note, now reading `Skills v1.2.3 (mattpocock/skills, pinned at image build)`.
- **The setup script runs no `claude plugin` command**; the marketplace add, plugin install, version parse and marker are gone. `execSetup` no longer returns a version.
- **Staleness** needs no new mechanism: the Dockerfile hash already covers the arg, and a new test pins that a Dockerfile with a different `SKILLS_REF` is stale.
- **AGENTS.md** describes how skills reach the image; the bind-mount rationale no longer says "plugins".

## Verification

- Real build of the new Dockerfile on this box: passes; label reads `v1.2.3`; 35 canonical skills and 35 Claude Code links; all seven required skills resolve by exact `name:`; lock records `v1.2.3`; no `~/.claude/plugins` directory.
- Negative build with `--build-arg REQUIRED_SKILLS="grill-me not-a-real-skill"` fails: `ERROR: skill 'not-a-real-skill' is missing from the installed set`.
- Setup wall-clock removed: the three plugin steps measured **8.7s** on the current `interlude-agent:latest` here (warm image, live network). The new setup script runs none of them.
- Upstream CHANGELOG 1.2.0 → 1.2.3 checked, and `git compare v1.2.0...v1.2.3`: none of the six session skills or `tdd` changed, so the pin is a safe bump from the laptop's 1.2.0 plugin.
- `vitest run`: 99 files, 1915 tests pass. `tsc --noEmit` clean. ESLint clean on every changed file.

## Notes for the reviewer

- The AC says "for the Claude Code agent"; the install targets all three agents because the owner's empirical finding on the issue says a single target makes the installer copy into that agent's directory and never create the canonical folder. The Claude Code link is still what the assertion checks for each skill.
- `docs/agents/workflows/tdd.md` still says "(v1.2.0)". Left alone: the spec says the vendored workflow files are unchanged, and `tdd` itself did not change through v1.2.3.
- `RunningContainer.skillsRef` is optional: handles rebuilt for a container that already started (boot re-adoption, follow-up reconnect) carry none, because the ref was recorded when that container started. The ledger column keeps its `skillsVersion` name; the handle field says what it holds, a git ref.
- First deploy after merge rebuilds the agent image (the Dockerfile hash changed); the install runs once at build over the network, never again at container start.

## Self-review (`/code-review`, fixed point `origin/main`, spec #215)

**Standards axis** — 6 findings, all read and acted on except two:
- Acted on: the new `imageSkillsRef` was a second unbounded Docker inspect of an image `ensureImage` had just inspected (#151) — folded into `ensureImage`, which now reports the ref off the inspect it already makes; the two three-line label reads collapsed into one helper; the null-path comments claimed a state that is unreachable after a rebuild and now say what is reachable; `skillsVersion` on the handle renamed `skillsRef`; "the six session skills" in AGENTS.md replaced with a reference to `REQUIRED_SKILLS` so it cannot rot.
- Disagree, kept optional field: the reviewer wants `skillsRef: string | null` required. The field is meaningful only on the handle `createWorkspaceContainer` returns; a rebuilt handle not carrying it is what `undefined` says, and requiring it would add `skillsRef: null` to seven unrelated test fixtures that fake a container with `container: never`.
- Disagree, kept test: "rebuilds when the pinned ref changes" is the existing stale-hash test with a specific stale hash. It pins acceptance criterion 6 in its own words — that the mechanism covers the ref — which the generic test does not name.

**Spec axis** — 3 "missing" items and 2 scope notes:
- The measurement, the changelog check and the real build all existed but were unrecorded; they are in *Verification* above, and the ~9s figure in the comments now cites its measurement.
- Scope notes accepted: the pinned installer version stays (reproducibility); the two extra build assertions (Claude Code link, lock-recorded ref) stay, and the Dockerfile now says the lock check couples the build to the installer's lock shape and why that is the right failure.
- Data path confirmed by the reviewer: one container-creation site, note + first-write-wins ledger for every task kind, `REQUIRED_SKILLS` pinned to the schema and the workflows directory, the hash covering the ref.


lennons301:

| gate | glob | path |
|---|---|---|

lennons301-reviewer (approved):
Approving. Gated PR (`Dockerfile.agent`, `src/lib/docker/**`, `src/db/schema.ts` match the agent-sandbox and persistence gates), so this satisfies branch protection and a human merges. One item below is worth the merger's judgement before they do.

## Verified

- **Branch:** full suite green here (99 files, 1915 tests, `--maxWorkers=2`); ESLint clean on all six changed files. The branch is one commit behind `main` (#227). Trial merge onto `origin/main` in a scratch worktree: no conflicts, `tsc --noEmit` clean, 101 files / 1944 tests pass. Safe to merge as-is.
- **Real build of the PR's `Dockerfile.agent`** (scratch tag, Dockerfile-only context as `buildImage` uses): label `co.interlude.agent-skills-ref=v1.2.3`; 35 canonical folders under `/home/node/.agents/skills`, 35 relative symlinks under `/home/node/.claude/skills`; all seven `REQUIRED_SKILLS` present with matching `name:` frontmatter and `ref: v1.2.3` in `~/.agents/.skill-lock.json` (v3, `.skills[<name>]` keyed as the jq expects); no `~/.claude/plugins`; npx cache removed.
- **Negative build** with `--build-arg REQUIRED_SKILLS="grill-me not-a-real-skill"`: the installer ran uncached (`Installed 35 skills`) and the build failed with `ERROR: skill 'not-a-real-skill' is missing from the installed set`, exit 1. AC "build fails if a required skill is missing" holds.
- **Setup runs no `claude plugin` command:** `buildSetupScript` has no install fragment; `execSetup` returns void; the only remaining `claude plugin` text in the tree is comments. Timing the removed steps on the pre-#215 image here gave ~5.0s (the PR says 8.7s; network variance either way, and the new setup runs none of them).
- **Ledger and feed:** `createWorkspaceContainer` carries `skillsRef` off `ensureImage`'s single inspect (warm path: one round trip, as claimed); `startTask` writes the `Skills v1.2.3 (...)` note and the first-write-wins `runs.skillsVersion`. Rebuilt handles (boot re-adoption, follow-up reconnect, `init.ts`) never reach that write site, so the optional field is honest.
- **Staleness:** the hash covers the whole Dockerfile; the new test pins a changed `SKILLS_REF` as stale.
- **Tests pin what the description says:** `REQUIRED_SKILLS` is asserted equal to `SESSION_SKILLS` + `docs/agents/workflows/*.md` stems; `-a claude-code`, `-g`, the pinned installer version and the `/tree/${SKILLS_REF}` URL are all read from the Dockerfile text with comments stripped for the "no `claude plugin`" check.
- **External claims:** `skills@1.5.23` exists on npm (also current latest); upstream tag `v1.2.3` exists; `skills add --help` confirms there is no ref flag (AGENTS.md's claim); `compare v1.2.0...v1.2.3` touches 34 files, none under the seven required skills' folders (the PR's "safe bump" claim). The plugin manifest at v1.2.3 is skills-only (no hooks/commands/agents/MCP), so swapping plugin for skills changes nothing in kind.

## For the human merger to weigh (not a blocker)

**The image installs a superset of what the plugin exposed.** The `.claude-plugin/plugin.json` at v1.2.3 lists 25 skills; `-s '*'` installs 35. The ten extras are upstream's `skills/in-progress/*` (six, all `disable-model-invocation: true`, so user-invoke only) and `skills/misc/*` — four that are **model-invocable**: `git-guardrails-claude-code` ("Set up Claude Code hooks to block dangerous git commands (push, reset --hard...)"), `setup-pre-commit` (Husky + lint-staged + Prettier), `migrate-to-shoehorn`, `scaffold-exercises`. The plugin author excluded these from the plugin; every agent container will now carry them as auto-triggerable skills. `/grill-me` and `workflow:tdd` themselves behave exactly as before (AC 5 holds as written), but the pass's environment is wider than it was, and the PR body does not mention it. The owner's issue comment prescribed `-s '*'` and noted "35 skills", so the implementer followed instruction — this is the owner's call. If narrowing is wanted, `-s` takes a list of names, so the fix is one line in the Dockerfile (the plugin's 25, or a curated subset that still includes the seven asserted plus e.g. `code-review`).

## Minor notes

- After merge, `src/lib/orchestrator/__tests__/fake-harness-turn.test.ts` (from #227) still mocks `execSetup: async () => ({ skillsVersion: "9.9.9" })`. Harmless (the return is ignored; tests pass on the merged tree) but the mock's shape no longer matches the real signature — a one-line tidy for whoever touches that file next.
- Suggestion, not tied to a standard: the pin is a tag name, and tags can move upstream. The lock records `skillFolderHash` per skill; if content pinning is ever wanted, the `/tree/<sha>` URL form or asserting the hash would do it. The ticket and the owner both asked for a tag, so this is fine as shipped.
- `docs/agents/workflows/tdd.md` still says "(v1.2.0)". Defensible — the file is a gated vendored derivative, the spec says the workflow files are unchanged, and `tdd` did not change through v1.2.3 (verified above).

PR #229, opened 2026-09-05
Lane-file rules follow the adapter's capabilities, and each lane shows its harness

lennons301:
Closes #219

## What changed

- **Parser follows capabilities.** `parseLaneConfig` already took the descriptor table (#214); a parsed lane now *carries* its adapter's capabilities (`LaneDefinition.capabilities`, through `ResolvedLane` and `LaneView`), so no downstream reader looks an adapter up, and a test lane on the fake adapter carries the fake's capabilities. New rule at parse: a **metered** lane on an adapter without `reportsCost` must price all three tiers or the file is refused naming the adapter and the rule; a **subscription** lane on the same adapter may omit prices; a Claude Code lane is unaffected (keyed on the capability, not billing or endpoint). An unregistered adapter is refused listing the registered ids.
- **Quota telemetry follows the harness.** `readMoneyGuards` (the one quota read shared by the sweep, dashboard and settings route) reads a lane's `quota_state` row only where its harness declares telemetry (`laneReportsQuota`), so the admission gate, the overage check and the tile all see null for such a lane. The failover ranking's per-lane observation map is filtered the same way (`actionableObservations`). The tier ladder reads only the refusal's own window, never a stored row — pinned by a new reducer test.
- **Quota tile.** "cannot report" keys on the adapter's `quotaTelemetry`, whatever the billing kind, and names the harness; "nothing observed yet" for a lane that could report and has not (with "bounded by spend" kept as a note on a metered one).
- **Settings screen.** Per lane: harness, image state and credential availability. Image state is one bounded Docker probe per adapter (`readHarnessImageStates` / `probeImageBuilt`), which tells a positive 404 from a daemon that did not answer — the latter renders *unknown*, never "not built". The payload still carries variable names only; the existing no-secret test covers the new fields.
- Docs: AGENTS.md convention bullet, runbook note, `lanes.yaml` comment.

## Acceptance criteria

- Unregistered adapter refused with registered ids — `lane-config.test.ts`
- Metered lane on non-cost-reporting adapter refused without prices; subscription accepted; Claude Code unaffected — `lane-config.test.ts`
- Tile "cannot report" vs "nothing observed" keyed on the capability — `fleet-view.test.ts`, `quota-tile-render.test.tsx`
- Gate and ladder never act on an observation from a no-telemetry lane — `quota-capability.test.ts` (over a real DB with a fake-adapter lane), `decide.test.ts`
[...]

PR #230, opened 2026-09-05
Session artefacts belong to the adapter; a cross-adapter lane move restarts on the branch

### LINKED ISSUE #217: Session artefacts belong to the adapter; a cross-adapter lane move restarts on the branch
## Parent

Spec: #213

## What to build

A paused or moved pass carries its conversation across lanes only when both lanes run the same adapter; across two different adapters it starts again on the branch with the work already pushed, and says so.

Today the session-transcript module hardcodes Claude Code's transcript path under the container user's Claude directory, and a resume always passes `--resume`. The limit that a lane move cannot carry a session across adapters is written down (issue #199) but nothing enforces it — there was no second adapter to enforce it against.

After this ticket the adapter owns the artefact paths for a session id, and preserving a paused run copies exactly those out. Restoring reads the adapter of the lane the paused pass ran on (via its task row's lane) and the adapter of the lane it is resuming on: same adapter, the artefacts are copied back and the adapter's resume shape is used; different adapter, the restore is refused, the session id is cleared, the pass starts fresh on the branch, and a system note tells the owner the conversation was not carried and why. The reducer's lane ranking is unchanged — a cross-adapter lane remains a legal failover target; it costs the conversation, not the attempt.

## Acceptance criteria

- [ ] The session-transcript module no longer names a vendor path; the container paths come from the adapter's artefact-path member
- [ ] A same-adapter lane move restores the transcript and resumes the conversation exactly as today (existing paused-run and lane-move tests pass)
- [ ] A cross-adapter lane move, exercised with the fake adapter over a real database, starts fresh on the branch, clears the session id, and posts a system note naming both lanes
- [ ] An adapter declaring no session resume never has artefacts copied out, and its runs pause and resume as a fresh start with the same note
- [ ] The stated limit at the foot of the adapter contract's module note is updated to say it is now enforced, and where

## Workflow

model: standard

## Blocked by

- #214

lennons301:
Closes #217

## What this does

A paused or moved pass carries its conversation across lanes only when both lanes run the same harness adapter. Across two different adapters it starts again on the branch with the work already pushed, and says so.

- **The store names no vendor path.** `src/lib/quota/session-transcript.ts` keeps one directory per run holding the artefacts the adapter named (`sessionArtifactPaths`) verbatim, plus a manifest (adapter, session id, the container path each came from). Claude Code's mangled-cwd derivation moves into its adapter (`claudeTranscriptPath`). Writes land beside the run's directory and are renamed into place, so a failed second save keeps the first transcript; prune removes only directories with a run id's shape, plus the pre-#217 one-file `.jsonl` entries nothing reads now.
- **Preserve asks the adapter.** `preserveSessionTranscript` resolves the refused pass's lane to its adapter and copies exactly those paths out, all or none. An adapter declaring `sessionResume: false` has nothing copied out.
- **Restore compares adapters.** `restoreSessionTranscript` takes the lane the pass is starting on, reads the predecessor's lane off `tasks.resumedFromTaskId`, and decides through the pure `decideSessionCarry` (`src/lib/harness/session-carry.ts`, a leaf). Same adapter: artefacts go back to their recorded paths and the harness resumes by id, as before. Different adapter, a harness that cannot resume, or a predecessor lane the file no longer declares: the session id is cleared, a plain-text system note names both lanes and why, and the pass starts fresh. A continuation queued with no session says nothing (a degrade, or a pause whose copy already failed and said so), except onto a harness that cannot resume, which the spec asks to be noted.
- **Announcements stop over-promising.** The failover note on the refused pass's feed and the early-resume issue comment (`describeSession`) say when the target lane runs a different harness, hedged because the target is advisory. Same-adapter wording is unchanged.
- **The ranking is unchanged.** A cross-adapter lane remains a legal failover target; it costs the conversation, not the attempt.
- **The stated limit** at the foot of `adapter.ts`, in `registry.ts` and in `lanes.yaml` now says it is enforced and where. AGENTS.md and the runbook read as current.

## Tests

- `cross-adapter-session.test.ts` drives the real `startTask` over a real database, a real on-disk store and fake lanes on three fake adapters (a second fake and one that cannot resume, added to `src/test/fake-harness.ts`): a cross-adapter move starts fresh with the id cleared (observed at exec time, since the turn's own id overwrites it afterwards) and a note naming both lanes; a same-adapter move restores; a retired predecessor lane refuses; a wall on the no-resume adapter fails over onto another adapter with nothing copied out; a wall with the session preserved fails over cross-adapter, is announced as not carrying, and is refused at start; alone in the catalog the no-resume run pauses with nothing copied out and resumes as a fresh start with the same note; an early resume onto another adapter is announced as not carrying.
- `session-carry.test.ts` pins the decision and its wording; `session-transcript.test.ts` covers the multi-artefact round trip, the safe write, the ceiling over the whole set and the shape-checked prune.
- Existing paused-run, lane-move, lane-failover and quota-resume suites pass; `quota-resume.test.ts` now seeds the predecessor row and passes the resolved lane.

Validation: `tsc --noEmit` clean; `vitest run`: 105 files, 1992 tests passing (`rm local.db*` first, per AGENTS.md); ESLint clean on every changed file. `origin/main` (PR #229) merged in; the one AGENTS.md conflict was the harness bullet both sides edited.

## Self-review (`/code-review`, fixed point `origin/main`, spec #217)

Acted on:
- `saveTranscript` removed the old directory before writing, so a failed second save could destroy a good first transcript. Now writes beside and renames into place; pinned by a test.
- Prune removed any directory not named for a live run with no shape check. Now only run-id-shaped directories (and interrupted `.tmp` ones).
- The decision's doc said a no-session continuation says nothing, but the code posted a `different-adapter` note for one. Reordered so it is silent, with the no-resume note kept as the one stated exception; tests updated.
- The `describeSession` issue comment and the failover feed note promised a continuation onto a different adapter that the restore would refuse. Both now say when the target runs a different harness.
- `AGENT_WORKDIR`'s doc orphaned `buildSetupScript`'s doc block, and the Claude adapter still hardcoded the path. The constant is a leaf (`src/lib/docker/workdir.ts`) both sides import.
- The cross-adapter test could not detect the session-id clear being removed. It now records every row's id at exec time. The failover test asserts the reducer's choice through the move's own announcement and no longer overclaims in its name.
- `SessionCarryRefusal` renamed `FreshStartReason`; `laneAdapter`'s doc corrected; a runbook line notes that runs parked across this deploy resume without their context.

Disagreed, kept:
- **Manifest-versus-lineage rule and `lane-unknown` refusal** (flagged as policy the spec did not state). The spec's rule needs an answer when the predecessor's lane is undeclared, and a lane file that re-pointed a lane id would make lineage say "same adapter" while the artefacts on disk are another harness's. Both refusals take the fallback the spec already names rather than resuming against the wrong harness, which fails the pass outright. The manifest's paths are needed regardless, so a restore puts each artefact back where it was read from.
- **`AGENT_WORKDIR` used in the container scripts** (flagged as scope). The constant had to leave the store module; a constant the scripts did not use would be a second statement of the same path, which the standards axis flagged from the other side.
- **Hand-built `ResolvedLane` literal in `quota-resume.test.ts`** (suggested resolving through `resolveLane`). It follows `claude-code.test.ts`'s existing fixture shape; the seam under test is the restore, not lane resolution.

## Review gates

Touches `src/lib/orchestrator/autonomy/**` and `src/lib/docker/**`, so this PR gets `human-signoff` and is not auto-merged.


lennons301:

| gate | glob | path |
|---|---|---|
| agent-sandbox | `src/lib/docker/**` | `src/lib/docker/workdir.ts` |
| autonomy-control | `src/lib/orchestrator/autonomy/**` | `src/lib/orchestrator/autonomy/paused-runs.ts` |

lennons301-reviewer (approved):
Fresh-eyes review against #217 and the repo standards. Gated PR (`human-signoff`, auto-merge off): this approval satisfies branch protection only; a human merges.

## Verified

- **Objective signals** (no CI on the branch, so run locally): `vitest run --maxWorkers=2` after `rm local.db*` — 105 files / 1992 tests green; ESLint clean on all 16 changed `.ts` files; `tsc --noEmit` clean.
- **Acceptance criteria**, each to code and a test:
  1. Store names no vendor path — `session-transcript.ts` has no `/home/node/.claude`; the only `.jsonl` mention is the prune's recognition of its *own* pre-#217 shape. The mangled-cwd derivation now lives in the adapter (`claudeTranscriptDir`/`claudeTranscriptPath`), pinned in `claude-code.test.ts`.
  2. Same-adapter move restores exactly as before — `cross-adapter-session.test.ts` ("restores and resumes … same adapter") plus the untouched paused-run / lane-move / lane-failover suites.
  3. Cross-adapter move over a real DB — real `startTask`, tmpdir store, fake catalog on three fake adapters, only Docker/GitHub/Discord stubbed. Session id observed cleared **at exec time** (the turn's own id overwrites it after), note names both lane labels and both adapter ids, attempt untouched.
  4. `sessionResume: false` — nothing read out at the pause (`readContainerFile` never called, `hasTranscript` false), resume queued with no session, resumed pass gets the `no-session-resume` note.
  5. Stated limit updated at the foot of `adapter.ts`, and in `registry.ts` and `lanes.yaml`, naming `restoreSessionTranscript`/`decideSessionCarry`.
- **Docs**: AGENTS.md harness/#169/#199 bullets, runbook (incl. the one-off that a run parked across this deploy resumes without context), `lanes.yaml` header all read as current.
- **Design**: `decideSessionCarry` is a pure leaf; the decision sits at the one seam that knows both ends; store writes are beside-and-rename with a test that fails the second write; prune is shape-checked. Lane ranking untouched.
- **Gate paths touched**: `src/lib/docker/**` changes are `"/workspace/repo"` literal → `AGENT_WORKDIR` (same value, new leaf `workdir.ts`), nothing about mounts/network/limits. `autonomy/paused-runs.ts` changes are announcement wording only.

## Non-blocking findings (for the human and a follow-up, not a reason to hold this)

1. **Manifest `sessionId` is stored but never compared with the queued id.** Chain: pause on lane A saves session S1 → a fresh-start continuation (degrade, or a cross-adapter move) starts S2 → S2 is walled and its copy *fails* → `hasTranscript(run.id)` is still true from S1 → `queueResumedPass` queues the resume carrying S2 → on a same-adapter lane `decideSessionCarry` says restore, S1's artefacts are written back and the harness is told `--resume S2` — the one outcome the module doc says it refuses. The old store had the same hole in a different shape; the difference now is the manifest carries the datum to close it (`stored.sessionId !== sessionId` → `not-kept`, one branch in `decideSessionCarry`). Suggest filing.
2. **Duplicated Code** (Fowler): `!catalog.ok || lanesShareAdapter(catalog.catalog, from, to) !== true` appears as `moveCrossesAdapter` in `paused-runs.ts` and inline in `failOverRunLane`. One helper beside `lanesShareAdapter` would keep the two announcements' rule from drifting. Suggestion only.
3. `laneAdapter()` (→ `getHarnessAdapter`, which throws on an unknown id) runs outside the `try` in `preserveSessionTranscript`, so a throw there would fail the pause the doc says it must never cost. Unreachable today (the parser validates adapters against the descriptor table, which is pinned to the registry), so defence-in-depth only.
4. **Policy the ticket did not state**, flagged for the human since the PR chose it deliberately: `lane-unknown` refuses a transcript even when the manifest names the same adapter as the destination, so retiring a lane from `lanes.yaml` while a run is parked on it costs that run its conversation. Cautious and documented; noting it so the choice is seen rather than inherited.

Issue #219, opened 2026-09-05
Lane-file rules follow the adapter's capabilities, and each lane shows its harness


lennons301:
## Parent

Spec: #213

## What to build

The lane file's rules follow each adapter's declared capabilities, and every surface that lists lanes says which harness runs them.

Today the lane parser accepts exactly one adapter id, requires nothing of a lane because of its adapter, and the quota tile decides "this lane cannot report quota" from the lane's billing kind — which is wrong the moment a subscription lane runs a harness that emits no rate-limit event.

After this ticket the parser takes the adapter descriptor table and refuses a document naming an adapter that is not in it, listing the registered ids. A **metered** lane on an adapter that does not report cost must declare prices for all three tiers, or the whole file is refused with the reason (the rule #175 argued for, now enforced by capability); a **subscription** lane on such an adapter may omit them — its marginal cash cost is zero, tokens are still recorded, nothing books to metered spend. The quota tile's "cannot report" state keys on the adapter's telemetry capability rather than billing kind, and the admission gate and tier ladder ignore lanes whose adapter reports none. The execution-lane settings screen shows, per lane, the harness, whether its image is built and whether its credentials are present, so an unavailable lane is explained before a pass tries it.

## Acceptance criteria

- [ ] A lane file naming an unregistered adapter is refused with the registered ids in the reason
- [ ] A metered lane on a non-cost-reporting adapter without three-tier prices is refused; the same lane declared subscription is accepted; a Claude Code lane is unaffected by the rule
- [ ] The quota tile says "cannot report" for a lane whose adapter declares no telemetry, whatever its billing kind, and "nothing observed" for one that could report but has not
- [ ] The admission gate and tier ladder never hold or step the fleet on an observation from a lane whose adapter declares no telemetry
- [ ] The settings screen lists harness, image state and credential availability per lane; no secret value is ever in the payload (existing test)
- [ ] The checked-in lane file parses unchanged

## Workflow

model: standard

## Blocked by

- #214

lennons301:
🤖 Attempt 1/3 starting.

Issue #214, opened 2026-09-05
Widen the harness contract and drive the orchestrator through a fake adapter


lennons301:
## Parent

Spec: #213

## What to build

The harness adapter contract grows to what a second harness actually needs, and the orchestrator is driven end to end through a fake adapter under test — with no change to what the fleet does on the Claude lane.

Today the contract is three members (exec environment, command, output handler), and the orchestrator still reads Claude Code's own vocabulary around them: turn exhaustion is the `error_max_turns` subtype, a quota refusal is an error flag plus a 429 status, and the stream parser lives in the orchestrator rather than under the adapter. A second harness cannot be written against that.

After this ticket an adapter also declares: the image its containers run; its static capabilities (`userInvokedSkills`, `quotaTelemetry`, `reportsCost`, `sessionResume`); how to compose a skill invocation for a named skill and agenda; the container paths holding a session's replayable artefacts; how to map a fleet effort level onto its own (or that it cannot). The turn result carries a **normalised outcome** — `completed`, `turn-limit`, `refused` (with a kind of quota / auth / other and an optional resume time), or `failed` — beside the vendor's verbatim fields, and the turn manager and reducer read only the normalised outcome. The Claude Code adapter implements every new member (effort maps one to one; its stream parser moves under the adapter), and a static descriptor table of adapter ids and capabilities is shared by the lane parser and the registry so the two cannot drift.

A **fake adapter**, registered only in tests, emits scripted normalised outcomes so the turn manager can run a whole turn without Claude Code or Docker. It is the test double every later ticket in this milestone builds on.

## Acceptance criteria

- [ ] The adapter interface declares image, capabilities, skill invocation, session artefact paths, effort mapping and a normalised turn outcome; the Claude Code adapter implements all of them
- [ ] No module outside the Claude Code adapter reads `error_max_turns`, the error flag or an API status code; the turn manager and reducer branch on the normalised outcome only
- [ ] The Claude Code stream parser lives under its adapter; the orchestrator does not import it directly
- [ ] A static adapter descriptor table (id, capabilities) exists; a test pins it to the registry so an adapter cannot be registered without a descriptor or described without being registered
- [ ] A fake adapter registered under test runs a whole turn through the turn manager, and a scripted `refused { quota }` reaches the reducer as a refusal
- [ ] Every existing turn-manager, reducer, lane and fleet-view test passes unchanged in behaviour; the recorded Claude stream fixture still parses to the same messages, cost and session id
[...]

PR #231, opened 2026-09-05
One agent image per adapter, built from a shared base

### LINKED ISSUE #216: One agent image per adapter, built from a shared base
## Parent

Spec: #213

## What to build

Each harness adapter gets its own agent image, built from a shared base, and a container runs the image its lane's adapter names.

Today there is one image, one Dockerfile, and one hash the image builder uses to decide whether to rebuild. A second harness would either bloat that image or fork the whole file.

After this ticket the agent Dockerfile is a **base** (git, gh, yq, jq, pnpm, the pinned skills, the workspace user) plus a thin **per-adapter layer** that installs one harness and pre-accepts its headless mode; the Claude Code layer is the only one so far. The image builder keys by adapter, names each image by adapter, and stamps the hash of both files so a base change rebuilds every adapter's image and a harness bump rebuilds one. Container creation asks the resolved lane's adapter which image to run. Container naming, the network and its aliases, memory and CPU caps, the reaper's prefix match and preview routing are untouched.

## Acceptance criteria

- [ ] The agent Dockerfile is split into a base and a Claude Code layer; the Claude Code image contains exactly what the single image contained before
- [ ] The image builder builds and inspects one image per registered adapter, stamped with a hash covering both the base and that adapter's layer
- [ ] Changing the base rebuilds every adapter image on next boot; changing one adapter layer rebuilds only that image
- [ ] Container creation uses the image the lane's adapter declares
- [ ] The reaper, capacity derivation and preview routing tests pass unchanged
- [ ] The production image build copies whatever the runtime needs to build adapter images, as it already does for the lane file

## Workflow

model: standard

## Blocked by

- #214
- #215

lennons301:
Closes #216

## What this does

Each harness adapter gets its own agent image, built from a shared base, and a container runs the image its lane's adapter names.

- **Dockerfile split.** `Dockerfile.agent-base` carries everything that is not the harness (git, gh, yq, jq, pnpm, the workspace user, the pinned skills). `Dockerfile.agent-claude-code` is `ARG BASE_IMAGE=interlude-agent-base:latest` / `FROM ${BASE_IMAGE}`, installs the Claude CLI, pre-accepts its headless mode, ends as `node`. A test pins that the base installs no harness and the layer repeats nothing the base did.
- **Image builder keyed by adapter image** (`src/lib/docker/image-builder.ts`). `ensureImage(image: HarnessImage)`: the base is built as `interlude-agent-base:latest` and stamped with its file's SHA-256; an adapter image is stamped with `imageStamp(baseHash, layerHash)`, a hash over **both** files, under the one `co.interlude.agent-dockerfile-sha256` label. A base change therefore rebuilds every adapter image (the base once — the first stale adapter rebuilds it, the next finds it current); a layer change rebuilds only that image; the warm path is still one inspect. The skills-ref label is the base's and reaches the adapter image by inheritance. Tests cover both clauses over two real layer files (the Claude layer and a fixture under `src/test/fixtures/`).
- **Container creation uses the adapter's image.** `createWorkspaceContainer` takes `image`; the turn manager passes `getHarnessAdapter(passLane.adapter).image`, so the container manager stays harness-agnostic. Pinned through the fake adapter (`fake-harness-turn.test.ts`) and in `container-manager.test.ts`.
- **Production image** `COPY`s `Dockerfile.agent-*` by glob; the `agent-sandbox` review gate matches the same glob.
- **Deploy** prunes every unused image carrying the stamp label instead of `docker rmi interlude-agent:latest` (a name that no longer exists — left alone, the forced rebuild that refreshes the unpinned CLI/gh installs each deploy would have silently stopped). Unused-only semantics as before; the legacy image is cleaned up by the same line. The label is pinned by a test.
- **Docker settings panel** lists one row per registered adapter through #219's bounded `readHarnessImageStates`, so it and the lane screen read one fact. New route test.

Container naming, the network and its aliases, memory/CPU caps, the reaper's `interlude-task-` prefix and preview routing are untouched; their suites pass unchanged.

## Verification

- **"Exactly what the single image contained before":** built the pre-split `Dockerfile.agent` from `origin/main` as a reference and the new base + layer on this box, then diffed `find / -xdev -printf "%M %u %g %p -> %l"` listings and `docker inspect .Config` (User, WorkingDir, Env, Cmd, Entrypoint, Labels): **10,644 entries each, zero differences**; same Claude Code version (2.1.261), 35 skills + 35 Claude links, same `.claude.json`. Against a *fresh* (uncached) build the only size deltas were apt/dpkg logs and `/usr/bin/gh` — upstream `stable` moved between the cached and fresh installs, not the split.
- **The production build path:** ran the real `ensureImage(CLAUDE_CODE_IMAGE)` through dockerode against the local daemon: built base then adapter (classic builder, `ARG` before `FROM`, `buildargs`), 54s; a second call made no build and returned the same `{ skillsRef: "v1.2.3" }`. Stamps on both images equal the hashes computed from the files on disk.
- `vitest run`: 104 files, 1981 tests, all passing (`rm local.db*` first). `tsc --noEmit` clean. ESLint clean on every changed file.

## Self-review (`/code-review`, fixed point `origin/main`, spec #216)

**Acted on (Standards):** the builder's and AGENTS.md's stated reason for taking a `HarnessImage` ("so the settings probe can import it without the registry") was not true and now states the real one; the deploy's third statement of the stamp label is pinned by a test; the docker route's daemon-down branch that restated `HarnessImageState`'s shape is gone (the probe's own rule yields unknown); the base header no longer contradicts its skills step; `/api/settings/docker` gained a test.

**Acknowledged, not changed:**
- *"Rebuilds on next boot"* is met at container creation, exactly where `origin/main` rebuilt — there was no boot-time build before either. A boot-time warm is `budgets.ts`'s stated follow-up and not this ticket's.
- The base image is never referenced by a container, so the deploy prune removes it every deploy and the first pass after a deploy does a full rebuild — the same cost as today's forced `rmi`.
- The split writes `.claude.json` after the skills install rather than before; the installer does not touch that file, and the order is the safer one for the bypass flag. Byte-identical result verified above.
- Two adapters both stale on one base change and ensured *concurrently* would build the base twice (last tag wins) — the same shape as two concurrent builds of the single image today.
- Test-only exports (`imageStamp`, `DOCKERFILE_HASH_LABEL`, `BASE_IMAGE_BUILD_ARG`) follow the repo's pin-by-test habit.

**Follow-up worth a ticket:** the Claude CLI is still installed unpinned (`npm install -g @anthropic-ai/claude-code`, as before), so a "harness bump" today happens through the deploy prune rather than a file change. Pinning it behind a build arg in the layer would make the per-adapter stamp the whole mechanism — out of scope here, since AC 1 asks for the image as it was.

## Review gates

Touches `Dockerfile.agent-*` and `src/lib/docker/**` (agent-sandbox), `.github/workflows/**`, `src/components/**` and `docs/agents/review-gates.yaml` (estate defaults), so this PR gets `human-signoff` and is not auto-merged.


lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/docker-status.tsx` |
| ci-secrets | `.github/workflows/**` | `.github/workflows/deploy.yml` |
| ci-secrets | `**/review-gates.yaml` | `docs/agents/review-gates.yaml` |
| agent-sandbox | `Dockerfile` | `Dockerfile` |

lennons301-reviewer (dismissed):
Fresh-context review of PR #231 against issue #216 and the repo standards (AGENTS.md, docs/agents/review-gates.yaml, platform review-gates.md). This PR is **gated** (`human-signoff`, auto-merge off): this approval satisfies branch protection only; a human merges.

## Objective signals (no CI on this branch, so run locally at 46469b6)

- `vitest run --maxWorkers=2` after `rm local.db*`: **104 files, 1982 tests, all passing**.
- `tsc --noEmit`: clean.
- ESLint on every changed `.ts`/`.tsx`: clean.
- The local daemon holds `interlude-agent-base:latest` and `interlude-agent-claude-code:latest` built through the real `ensureImage` path; their stamp labels equal, to the byte, `sha256(Dockerfile.agent-base)` and `imageStamp(baseHash, layerHash)` computed from the files at commit ab9a7ef. The final commit only edited comments in the base header, which is exactly the change that should make both images stale next time. The stamp formula is therefore confirmed against a real build, not only the fake daemon in the tests.

## Acceptance criteria

1. **Split into base + Claude Code layer; same contents.** Every instruction of the pre-split `Dockerfile.agent` on `origin/main` lands in exactly one of the two files: the base keeps apt tooling, yq, gh, pnpm, `/workspace`, `~/.claude`, `USER node`, the pinned skills install, the skills-ref LABEL and `WORKDIR`; the layer carries only the CLI install and the `.claude.json` pre-acceptance, root for the install and `node` at the end. The only ordering change (CLI + `.claude.json` after the skills install) is documented, and the PR body records a full filesystem/inspect diff with zero differences against a build of the old file. Met.
2. **Builder keyed by adapter image, stamped over both files.** `ensureImage(image: HarnessImage)`; base stamped with its file hash, adapter image with `imageStamp(base, layer)` under the one `co.interlude.agent-dockerfile-sha256` label; the warm path is still one inspect and the base is not inspected for a current adapter image. Tested over two real layer files (the Claude layer plus `src/test/fixtures/Dockerfile.agent-fixture`). Met.
3. **Base change rebuilds every adapter image (base once); layer change rebuilds one.** Both clauses pinned, including the "does not treat the base's own stamp as a current adapter image" edge. The rebuild happens at container creation rather than literally "on next boot" -- the same place rebuilds have always happened (there was no boot-time build before either); I agree with the PR body's reading and the human should be aware of it. Met in substance.
4. **Container runs the lane adapter's image.** `createWorkspaceContainer` takes `image`; the turn manager -- the one production caller -- passes `getHarnessAdapter(passLane.adapter).image`; `ensureImage` and `Image:` read the same value. Pinned in `container-manager.test.ts` and through the fake adapter in `fake-harness-turn.test.ts` (the container ran the fake's image, not the Claude one). Met.
5. **Reaper, capacity, preview routing tests unchanged.** None of those files are in the diff; the full suite passes. Met.
6. **Production image copies what the runtime needs.** `COPY --from=build /app/Dockerfile.agent-* ./` lands both files beside `custom-server.js` (process.cwd()); `.dockerignore` excludes neither; the fixture layer under `src/test/fixtures/` is outside the root glob, as intended. Met.

## Standards

- **Deploy prune matches the stamp label.** `docker image prune -a -f --filter label=co.interlude.agent-dockerfile-sha256` is pinned to `DOCKERFILE_HASH_LABEL` by a test (right call: a renamed label would be a silent no-op behind `|| true`). I checked the semantics: unused-only, so a parked container's adapter image survives exactly as `docker rmi` left it before; the pre-#216 `interlude-agent:latest` carries the same label and is cleaned up by the same line. One nuance for the operator, not a defect: when the base is iterated before its child, Docker untags it but keeps the body until the child goes -- the next pass still sees `interlude-agent-base:latest` as missing and rebuilds both, which is the intended forced rebuild; the leftover dangling image is collected by the following deploy's `docker image prune -f`.
- **Review gate** updated to `Dockerfile.agent-*`, matching the production `COPY` glob (bash `[[ == ]]` semantics match both files at the root).
- **AGENTS.md** kept current: tech-stack line, project-structure line, the #215 bullet re-pointed at the base, the #214 bullet's `image.ts` sentence, and a new #216 bullet stating the design and the invariants (base never installs a harness, pinned by a test; the label the deploy prunes by, pinned by a test). `docs/runbook.md` has no stale image references.
- **Claude Code lane unchanged**: `buildExecEnv`/`buildCommand` untouched; no module outside `src/lib/harness/claude-code/` gained knowledge of the image beyond the turn manager asking the adapter, which is where the resolved lane is. `image-builder.ts` now imports only a type from `harness/adapter`, so the Docker-side module has no runtime coupling to the adapter graph.
- Smell baseline: nothing to raise. The test-only exports (`imageStamp`, `BASE_IMAGE_BUILD_ARG`, `DOCKERFILE_HASH_LABEL`) follow the repo's pin-by-test habit and are acknowledged in the PR body.

## Non-blocking notes (suggestions)

1. `src/lib/harness/claude-code/image.ts` still says it is a leaf "because the image builder is Docker-side code with no business importing the stream parser". After this PR the image builder no longer imports that module at all -- its importers are the adapter's `index.ts` and three tests. The leaf still earns its keep (the tests import the image without pulling the DB), but the sentence names the wrong beneficiary. A one-line comment fix, fine as a follow-up.
2. Agree with the PR body that pinning the Claude CLI behind a build arg in the layer is a follow-up ticket, not this one: AC 1 asks for the image as it was, and today the deploy prune is what refreshes the unpinned installs.

## For the human merging

- First pass after the deploy pays a full base + layer build (same cost as today's forced `rmi`); the 20-minute `OCCUPANCY_DIVERGED_MINUTES` / provisioning grace still covers it.
- Visual change on the settings screen: the Docker panel now shows one `<adapter> image` row per registered adapter (one today), each with the image name and a ready / not built / unknown chip.

lennons301:

| gate | glob | path |
|---|---|---|
| visual-ui | `**/components/**` | `src/components/docker-status.tsx` |
| ci-secrets | `.github/workflows/**` | `.github/workflows/deploy.yml` |
| ci-secrets | `**/review-gates.yaml` | `docs/agents/review-gates.yaml` |
| agent-sandbox | `Dockerfile` | `Dockerfile` |

lennons301-reviewer (approved):
## Review — approve (gated: this only satisfies branch protection; a human merges)

Fresh-context review against #216 and the repo standards (AGENTS.md, `docs/agents/review-gates.yaml`, `~/code/platform/standards/review-gates.md`). No CI checks are reported on this branch, so every objective signal below was run locally on the worktree.

### Objective signals
- `vitest run --maxWorkers=2` after `rm local.db*`: **106 files, 2007 tests, all passing**.
- `tsc --noEmit`: clean.
- ESLint on every `.ts`/`.tsx` this PR touches: clean.
- Branch already contains `origin/main` (merge commit 37fd98e); no file main changed since the fork point overlaps a file this PR changes.

### Acceptance criteria
1. **Split, contents preserved** — verified structurally against `origin/main:Dockerfile.agent`: the base is every pre-split instruction minus `npm install -g @anthropic-ai/claude-code` and the `.claude.json` write; the layer is exactly those two, root only for the install, ends `USER node`, inherits `WORKDIR /workspace` and the `co.interlude.agent-skills-ref` label. The only change is the order (CLI install now after the skills install); the PR body reports a byte-identical `find` listing and identical `.Config`, which is consistent with what the files say. `image-builder.test.ts` pins that the base installs no harness and the layer repeats nothing the base did.
2. **One image per adapter, stamped over both files** — `ensureImage(image)` stamps `imageStamp(baseHash, layerHash)` under the one label; the adapter build's `labels` override the label inherited from the base, and the test "does not treat the base's own stamp as a current adapter image" pins the case where it would not.
3. **Base change rebuilds all, layer change rebuilds one** — pinned over two real layer files (the Claude layer and `src/test/fixtures/Dockerfile.agent-fixture`), including "base built once for N stale adapters", "missing base under a stale adapter", and "unstamped adapter image". Warm path is one inspect; the base is not inspected for a current adapter image. Note for the human: the ticket says "on next boot"; the PR rebuilds at the next container creation, which is where `origin/main` rebuilt too (I checked `instrumentation.ts`: there is no boot-time image build to attach to). The PR body states this reading; I find it the correct one.
4. **Container creation uses the adapter's image** — `createWorkspaceContainer` takes `image` and uses `options.image.name`; its single caller (`turn-manager.ts:559`) passes `getHarnessAdapter(passLane.adapter).image`; the fake-harness turn test asserts the container ran the fake's image, not the Claude one.
5. **Reaper / capacity / preview tests unchanged** — none of those suites is in the diff, and all pass.
6. **Production image copies what the builder needs** — `COPY --from=build /app/Dockerfile.agent-* ./`; `.dockerignore` excludes none of them; the test fixture lives under `src/test/fixtures/` so the root glob does not ship it.

### Also checked
- Removed exports (`getImageName`, `imageExists`, `buildImage`) have no remaining callers; the docker settings route is the only consumer of `/api/settings/docker` and was updated with the component.
- Deploy prune: `docker image prune -a -f --filter label=co.interlude.agent-dockerfile-sha256 … || true` keeps the unused-only semantics of the old `rmi`; the pre-#216 `interlude-agent:latest` carried the same label so it is cleaned up by the same line; the label is pinned to `DOCKERFILE_HASH_LABEL` by a test so a rename cannot make the prune a silent no-op.
- Review-gate glob `Dockerfile.agent` -> `Dockerfile.agent-*`: necessary (the old name no longer exists) and a widening, not a removal.
- No stale references in `docs/runbook.md`, `scripts/`, or `docker-compose.yml`; the remaining `Dockerfile.agent` mentions are in dated 2026-03 plan/spec records.
- Smell baseline: nothing I can tie to a standard. One **suggestion**, not blocking: the tri-state `HarnessImageState.built` -> chip mapping now exists twice (`docker-status.tsx` and `imageChip` in `execution-lane-settings.tsx`) with different words and tones; if the two panels are meant to say the same thing about an image, a shared helper would keep them from drifting. Reasonable to leave as is since the readout and the lane row have different jobs.

### For the human merging
- Visual change on the settings screen: the Docker readout now renders one `<id> image` row per registered adapter (image name + ready / not built / unknown chip) instead of a single image row — worth a glance in the UI.
- After this deploys, the base image is not referenced by any container, so the prune removes it every deploy and the first pass afterwards does a full base + layer build (same cost as today's forced `rmi`). An adapter image kept alive by a parked container is left alone, as before.
- The implementer's follow-up note stands: the Claude CLI is still installed unpinned in the layer, so a "harness bump" today still arrives via the deploy prune rather than a file change. Out of scope for AC1 ("exactly what the single image contained before"); worth its own ticket.

PR #235, opened 2026-09-05
Codex CLI adapter, with a subscription lane and an API lane

lennons301:
Closes #221

## What this adds

A second harness behind the seam #214 widened: the **Codex CLI adapter** (`src/lib/harness/codex/`), registered with a descriptor, and two lanes in `lanes.yaml` — `codex-subscription` (ChatGPT plan, `billing: subscription`) and `openai-api` (API key, metered, priced). Neither is in `primary`, so nothing routes onto them until their credential is provisioned.

- **Capability statement** (descriptor): no quota telemetry, no cost reporting, session resume, user-invoked skills **off** until #224.
- **Command**: `codex exec --json` headless, approvals and sandbox bypassed (the container is the sandbox), `--skip-git-repo-check`, `-c features.plugins=false`, model pinned with `-m`, effort mapped onto `model_reasoning_effort` (`max` → `xhigh`; unmappable → omitted), `codex exec resume <id>` for a follow-up, prompt piped to stdin.
- **Credentials, exec-scoped**: an API key passes through as `CODEX_API_KEY` (the variable the CLI reads — `OPENAI_API_KEY` alone sends no bearer, measured on 0.149.1 and 0.153.4). The ChatGPT `auth.json` is materialised from `CODEX_AUTH_JSON` into a per-exec Codex home and removed with it in an EXIT trap; the script also sweeps any home a killed turn left behind. The cleanup is part of the command the adapter builds, pinned by test.
- **Base URL** → a per-exec `config.toml` `model_providers` entry (Responses wire).
- **Sessions** outlive the per-exec home through a symlinked persistent sessions directory; the cleanup normalises each rollout onto a deterministic canonical name, which is the adapter's one session artefact (the CLI finds a rollout by that name shape and refuses any other — measured).
- **Parser**: items → the transcript's stored shapes (agent text, `Shell`/`Patch`/MCP/`WebSearch`/`Plan` tool rows with results), session id, token usage in the fleet's shape, normalised outcome. `turn.completed.usage` is the **thread's running total** (measured: a resumed turn spending 100 tokens reported 4200), so the parser charges the difference from the largest total recorded on a turn-complete feed note of any task sharing the session id.
- **Exit classifier**: `turn.failed`'s one sentence → `refused { quota }` (wall wording / 429), `auth` (401/403), `other` (any other provider status), else `failed`.
- `Dockerfile.agent-codex`: `npm install -g @openai/codex` plus the persistent sessions directory, on the shared base.
[...]

Issue #225, opened 2026-09-05
OpenCode lane proven end to end


lennons301:
## Parent

Spec: #213

## What to build

Prove the OpenCode lane end to end the way #175 proved OpenRouter, on the same GLM models the skin lane already runs, and record the comparison.

On a target repo, one real ticket runs its implement pass on `opencode-openrouter-glm` to a PR, its review pass, and lands; cost is booked from the lane's declared prices and checked against OpenRouter's billed truth for the same generation ids (the check script from #193 is the prior art); lane and harness are stamped and appear on the dashboard and in the digest. One generation session runs on the lane and loads a session skill through OpenCode's skill tool from the adapter's instruction, reaching the publish step. A side-by-side note records the same ticket shape on the skin lane and the native lane: tokens, cost, cache reads, attempts. Every deviation from the spec is recorded on this issue and the lane file and adapter are tuned in the same PR. When the generation session is proven, the adapter's user-invoked-skills capability is switched on.

No new credential is needed: the OpenRouter key is already provisioned.

## Acceptance criteria

- [ ] An implement pass on the OpenCode lane produces a PR, a review pass returns a verdict the orchestrator posts, and the PR lands or parks exactly as a Claude-lane run would
- [ ] Spend is booked from the lane's prices, appears in metered spend under the cap, and agrees with OpenRouter's billed figure for the run's generations within the rate-card tolerance
- [ ] A generation session on the lane loads its skill via the adapter's instruction and reaches the publish step
- [ ] A skin-vs-native comparison for one ticket shape is recorded on this issue (tokens, cost, cache reads, attempts)
- [ ] Findings are recorded on this issue with the CLI version observed; the adapter's user-invoked-skills capability is switched on only if the session proof passed
- [ ] The spec's OpenCode facts that proved wrong are corrected on #213 by comment

## Workflow

model: standard

## Blocked by

- #220
- #222
- #223

lennons301:
🤖 Attempt 1/3 starting.

lennons301:
## Verification findings (2026-09-06) — OpenCode lane end to end

[...]

Issue #220, opened 2026-09-05
Refusals and turn bounds on a harness with no rate-limit event and no CLI ceiling


lennons301:
## Parent

Spec: #213

## What to build

A pass on a harness that emits no rate-limit event and has no turn or budget flag is still bounded, and a provider refusal there still pauses or fails over through the existing wall ordering.

Today a refusal is recognised only through Claude Code's rate-limit event and rejection shape, and a turn is bounded by the CLI's own `--max-turns` and `--max-budget-usd`. Neither exists on Codex or OpenCode.

After this ticket the reducer acts on the normalised `refused { kind: quota }` outcome from **any** adapter: tier degrade when the refusal names a tier, lane failover when another lane can serve the run, otherwise park — and when the provider gave no reset time, the park uses a default backoff constant named in the budgets module and the same resume bound. A refusal of kind `auth` is a lane-availability failure (report, do not retry on that lane); `other` fails the turn as today. Separately, the orchestrator enforces an adapter-agnostic **wall-clock ceiling per exec**, generous by default and named in the budgets module, ending an over-long turn as `turn-limit` so the attempt exhausts the way a turn-limited Claude turn does today — which also bounds the hung-process case seen on suspended hosts. Claude Code keeps its flags; the ceiling is a second bound, not a replacement.

## Acceptance criteria

- [ ] A `refused { quota }` from the fake adapter follows the wall ordering in the reducer: degrade when a tier is named, else failover when a lane can serve, else park
- [ ] A refusal carrying no reset time parks on the default backoff constant and counts against the existing resume bound; at the bound the run exhausts to a human
- [ ] A `refused { auth }` marks the lane unavailable for the run and reports it; it never retries the same lane
- [ ] A turn exceeding the wall-clock ceiling ends as `turn-limit`, the container exec is stopped, and the attempt accounts for it exactly as a turn-limited Claude turn does today
- [ ] The Claude Code adapter's existing rate-limit pause, degrade and failover tests pass unchanged
- [ ] The ceiling and backoff constants are documented in the runbook beside the other budgets

## Workflow

model: heavy

## Blocked by

- #214
- #219

lennons301:
**Upstream landed (2026-09-05):** #214 → PR #227 and #215 → PR #228 are merged on `main`. Build on what they shipped rather than the spec's wording:

[...]

PR #236, opened 2026-09-05
OpenCode adapter, with an OpenRouter open-weights lane

### LINKED ISSUE #222: OpenCode adapter, with an OpenRouter open-weights lane
## Parent

Spec: #213

## What to build

An OpenCode adapter, registered and unit-tested, with one lane declared: `opencode-openrouter-glm` — the same OpenRouter credential and GLM models the fleet already runs through the Anthropic-compatible skin, now on a harness that speaks to OpenRouter natively, so the two paths can be compared.

The adapter runs `opencode run` headless with JSON output, continues a session with the session flag, pins the lane's model in OpenCode's `provider/model` form, maps the fleet effort level onto the variant flag where the provider exposes one (otherwise omitted with a note), and auto-approves permissions through a per-exec config inside the container. Its output handler parses the JSON event stream into the stored message shapes, the normalised outcome, token usage and the session id. Credentials pass through as variables the harness already reads; a per-exec provider config carries the lane's base URL when declared. An OpenCode image layer installs the CLI. The adapter declares no quota telemetry, no cost reporting, session resume, and user-invoked skills **off** until the proof ticket flips it — OpenCode loads skills through a model-called tool, so its skill invocation is an instruction to load the named skill and follow it.

## Acceptance criteria

- [ ] The adapter is registered with a descriptor; the lane file parses with the OpenCode lane, priced for all three tiers, and it is unavailable when the OpenRouter key is absent
- [ ] Env and command builders are table-tested: model pin in provider/model form, variant mapping, session continuation, provider config, and no credential on the command line
- [ ] A recorded fixture of a real `opencode run --format json` run is checked in and parses to agent text, tool events, tool results, token usage and a session id; a recorded provider-refusal fixture parses to `refused { quota }`
- [ ] The OpenCode image layer builds from the shared base, and a container on the OpenCode lane runs it
- [ ] The adapter's skill invocation text names the skill and the agenda and is what seed composition emits for this adapter
- [ ] The `BLOCKED:` marker, review verdict and triage exit parsers read the adapter's final message unchanged
- [ ] AGENTS.md gains the adapter under the harness section with its capability statement

## Workflow

model: heavy

## Blocked by

- #216
- #217
- #218
- #219

lennons301:
Closes #222

## What this adds

A second harness behind the seam #214 widened: the **OpenCode adapter** (`src/lib/harness/opencode/`), registered with a descriptor, and one lane in `lanes.yaml` — `opencode-openrouter-glm`, the same OpenRouter credential and the same GLM models as `openrouter-glm`, on a harness that speaks to OpenRouter natively. Not in `primary`; nothing routes there until an operator picks it or the ranking does.

- **Capability statement** (descriptor): no quota telemetry, no cost reporting, session resume, user-invoked skills **off** until #225.
- **Command**: `opencode run --format json --pure`, prompt piped from `OPENCODE_PROMPT` on stdin, `--model` in the CLI's `provider/model` form (the lane's `models` carry it: `openrouter/z-ai/glm-5.3-flash`), `--session <id>` to continue, effort as `--variant`.
- **Per-exec config** at a fixed path named in `OPENCODE_CONFIG`: `permission: { "*": "allow" }`, plus a `provider.<id>.options.baseURL` override only when the lane declares a `base_url` (built with jq). No secret in it.
- **Credentials** pass through as the variables the CLI reads (`OPENROUTER_API_KEY` as-is), exec-scoped; nothing on the command line.
- **Session** = the CLI's one SQLite file: pinned with `OPENCODE_DB`, WAL checkpointed by the turn script after the CLI exits, and that file is the adapter's artefact.
- **Terminal event of the adapter's own**: `run --format json` ends with no event, so the script prints `{"type":"interlude.turn_exit","exitCode":N}` after the checkpoint; without it the outcome is null (container death, or #220's stop).
- **Exit classifier**: `error.data.statusCode` 429/402 → `refused { quota }` (no reset, no window), 401/403 → `refused { auth }`, other status → `refused { other }`, status-less error → `failed`; never `turn-limit` (no `--max-turns`; #220's wall clock is the in-turn bound).
- **Parser**: `text` parts → agent text; `tool_use` parts → tool rows under the transcript's verbs with camel-case arguments renamed to the keys the view reads (an OpenCode edit gets the same line diff); usage summed over `step_finish` (reasoning counted as output); unknown events → recorder; the last `error` rides on the terminal result as `lastError`.
- `Dockerfile.agent-opencode`: `npm install -g opencode-ai` on the shared base, plus the database's directory.

## Verification

- Full suite passes (`--maxWorkers=2`); `tsc --noEmit` clean; eslint clean on every changed file.
- **Measured on 1.18.29 in the built image, against OpenRouter (2026-09-05, ~$0.30 of credit):** a turn on GLM 5.3 Flash ran two shell commands and finished (exit 0, 20s, no stderr); `--session <id>` resumed it; the same database copied alone into a fresh container resumed the session — with the same repository root commit and a different one; an invalid key surfaced as a 401 `APIError`; `--variant max` on GLM ran as if unset (exit 0); `OPENCODE_DB` is honoured; `opencode db 'PRAGMA wal_checkpoint(TRUNCATE)'` folded a 1.5 MB WAL into the main file in ~1s; the OpenRouter provider enables from `OPENROUTER_API_KEY` with no `/connect`.
- **The adapter's built command, verbatim, ran in the built image** (env from `buildOpenCodeTurnEnv`, key from Doppler): fresh turn → work done, sentinel last, config written, WAL 0 bytes; `--session` resume → recalled the earlier turn; a declared base URL → the provider override landed in the config and the turn ran.
- **Fixtures** are the CLI's stdout verbatim (the test appends the terminal event as the script does): a real GLM turn, its resume, a live OpenRouter 401, and — because the account had credits, so OpenRouter would not refuse — a 429 whose body was captured live from OpenRouter's free tier and a 402 with OpenRouter's documented wording, both replayed to the **real CLI** by a local stub (the CLI retried the 429 nine times over ~80s before surfacing it).
- `opencode-lane-turn.test.ts` drives the real `startTask` on the shipped lane over the real `lanes.yaml`: OpenCode image, the adapter's exec, the recorded turn on the feed, charged at the lane's prices ($0.00069376, which equals the CLI's own estimate since it is the same price table); and a recorded 429 fails over to `openrouter-glm` as a cross-adapter move.

## Self-review (mattpocock-skills:code-review, fixed point origin/main)

Acted on: a stale "only layer so far" in AGENTS.md's #216 bullet; a dangling reference to a Codex parser this tree does not have; a malformed `error` event forwarded by type would have been dropped by the recorder's Claude-vocabulary allowlist — it now goes as an unparseable line, pinned by test; the cost note now says what a *subscription* OpenCode lane would book; `requireHarnessDescriptor` shared with the Claude Code adapter instead of duplicated; the image test derives the data directory from `OPENCODE_DB_PATH`; two stale "the one adapter that ships" phrases; the lane-turn test above, so "a container on the OpenCode lane runs it" is pinned rather than a prose claim.

Kept, with the reason:
- **Effort → variant: identity mapping, flag always passed, no feed note.** The ticket says "maps the fleet effort level onto the variant flag where the provider exposes one (otherwise omitted with a note)". Whether a variant is exposed is a *per-model* fact inside the CLI (its OpenRouter provider defines `low`/`medium`/`high` for a reasoning model; Anthropic 4.7+ defines all five), and the CLI drops a variant the model does not define silently (measured). So the adapter passes the fleet's level — which *is* one of the CLI's names — and the CLI applies it where the model exposes it. The adapter cannot see which, so it cannot honestly note "omitted"; the only omitted case is a level outside the fleet's vocabulary, unreachable from either entry point. Consequence, stated in AGENTS.md: an `xhigh`/`max` ticket on GLM runs at the model's default with no trace. A per-provider variant table copied from the CLI's internals would be the alternative, and it would drift.
- **402 → `refused { quota }`.** "Insufficient credits" is the account's allowance spent; the wall ordering (fail over, else park on #220's backoff and re-resume to the bound) is the right response, and it spends no attempt. A judgement the spec did not make explicitly.
- **`Retry-After` ignored** (`resumeAfter: null` on a 429): it sits inside the provider's response body, not the CLI's fields; honouring it would park for 5s and burn the resume bound on a flapping upstream. #220's default backoff applies; a refinement for #225 if it matters.
- **`--pure`**: no external plugins, so a fresh container fetches nothing from a plugin registry at every turn start.
- **Per-adapter env builder, `isRecord`/`readString` copies, byte-identical `mapEffort`**: the seam keeps adapters independent by design; the non-vendor parts are three lines.
- **Provider override with no shipped lane using it**: the ticket asks for it ("a per-exec provider config carries the lane's base URL when declared"; AC2 "provider config"), and it ran against OpenRouter in the image.
- **Unpinned `opencode-ai` install**: the Claude and (in #235) Codex layers' convention; the deploy's prune picks up new versions. The JSON event schema is undocumented, so a release could change the wire — flagged for #225.
- **AGENTS.md bullet length**: in line with its neighbours; the file's size is its baseline, not this PR's to fix.

## For #225 (proof ticket)

- Whether the skill-invocation text actually makes the model call the `skill` tool and load the SKILL.md (unguessable sentinel), then flip `userInvokedSkills`.
- Whether GLM honours `reasoning.effort` from `--variant low|medium|high` on OpenRouter, and whether `xhigh`/`max` should be collapsed rather than dropped.
- The real wording of an OpenRouter 402 through the CLI; `Retry-After` handling.

## Gates

Touches `Dockerfile.agent-*` and `src/lib/docker/**` (test only) → `human-signoff` per `docs/agents/review-gates.yaml`. Pending #235 (Codex) touches the same shared files (`descriptors.ts`, `registry.ts`, `adapter.ts`, `session-carry.ts`, `stream-recorder.ts`, the descriptor/lane/image tests); whichever lands second will need a trivial conflict resolution in those.


lennons301:

| gate | glob | path |
|---|---|---|
| agent-sandbox | `Dockerfile.agent-*` | `Dockerfile.agent-opencode` |

lennons301-reviewer (approved):
**Approve** — gated PR: `human-signoff` stays, a human merges. Reviewed at `28ac8bd` against issue #222 and AGENTS.md.

## Objective signals (no CI on the branch, so run locally)
- `pnpm test --maxWorkers=2`: 114 files, 2185 tests pass, no flakes first run. `tsc --noEmit` clean. `eslint` on every changed `.ts`: clean.
- Built `Dockerfile.agent-opencode` from stdin on the local `interlude-agent-base:latest`: builds; both layers came back `CACHED`, so this exact layer had been built on this base before (the tag had since been removed — the image the PR measured in was absent from the box, the cache was not).
- In that image (1.18.29): `run --help` has every flag the command uses (`--format json`, `--pure`, `--model`, `--variant`, `--session`); `opencode db [query]` runs a PRAGMA and `db path` honours `OPENCODE_DB`; `jq`/`timeout` present. Ran the turn script **keyless** (no credential, no spend): the config landed exactly as documented (allow-all plus `provider.openrouter.options.baseURL`), the prompt was read from stdin, one `error` event then exit 1, the checkpoint left a 0-byte WAL, and the terminal event printed last.

## Acceptance criteria — all seven met, each pinned by a test
1. Descriptor + registry (`descriptors.test.ts`); lane parses, priced for all three tiers, unavailable without `OPENROUTER_API_KEY`, not in `primary` (`lane-config.test.ts`).
2. Env/command table-tested: provider/model pin, `--variant`, `--session` with a shape check before `bash -c`, per-exec provider config via jq, no credential or prompt on the command line (`opencode.test.ts`).
3. Real fixtures: GLM turn → text, two `Bash` rows, summed usage, session id; live 401 → `refused { auth }`; 429 and 402 → `refused { quota }` (`opencode-stream.test.ts`, `opencode-outcome.test.ts`).
4. Layer builds from the base (verified above); `opencode-lane-turn.test.ts` drives the real `startTask` on the shipped lane over the real `lanes.yaml`: OpenCode image, adapter exec, feed rows, charged from lane prices ($0.00069376), 429 → failover to `openrouter-glm` with no attempt spent.
5. Skill invocation text is what `composeSeed` emits, framing intact (`seed.test.ts`).
6. `BLOCKED:`, `parseReviewVerdict`, `parseTriageExit` read the final message unchanged.
7. AGENTS.md bullet checked claim by claim against the code — command, env, config, `OPENCODE_DB` + checkpoint, terminal event, classifier table, parser mapping, fixtures, image, lane-turn test. It matches.

## Standards
Capabilities read from the descriptor table, never restated; `createOutputHandler` ignores the lane and reports `rateLimit: null`, consistent with `quotaTelemetry: false`; the recorder-allowlist collision on `error` is handled (malformed → unparseable line, pinned); `src/lib/docker` has test-only changes and `container-manager.ts` stays harness-agnostic; the artefact copy is binary-safe (base64 out, Buffer in), so the SQLite artefact survives a pause. Shared-module edits are comment updates plus one shared `requireHarnessDescriptor` — in scope.

## For the human signing off (agent-sandbox gate)
- **Sandbox posture:** `permission: { "*": "allow" }` via a per-exec config — OpenCode's equivalent of `--dangerously-skip-permissions`; the container remains the boundary. No new `Bind`, network, alias or resource change. Credentials exec-scoped; prompt via env → stdin; session id regex-checked before reaching `bash -c`; `--pure` fetches no plugins. The config file holds no secret (checked in the image).
- **Unpinned `npm install -g opencode-ai`** matches the Claude layer's convention, but this CLI's JSON event schema is undocumented, so a release could change the wire under the parser silently. Flagged for #225 in the PR body; a pin sooner is a fair call.
- **Accepted deviation from the ticket's "otherwise omitted with a note":** `mapEffort` is the identity and `--variant` is always passed; the CLI drops a variant the model does not define silently (measured), so an `xhigh`/`max` ticket on GLM runs at the model default with no trace. Stated in AGENTS.md and defensible — the adapter cannot see per-model variants. `adapter.ts`'s `mapEffort` doc still says #222 brings the note; stale line, nit.
- On the shipped configuration a 429/402 fails over to `openrouter-glm`, which is the **same key**, so the first move will probably meet the same wall on the next lane; bounded by the resume bound, then a pause. The ranking's business (#176), not this adapter's — worth knowing when reading the first real failover.

## Suggestions (not tied to the ticket or a standard; non-blocking)
- `classifyOpenCodeExit` lets any `error` event outrank exit 0. Measured behaviour is that an error is terminal, so fine today; if a release ever emits a recoverable `error` mid-turn, a finished pass would read as refused/failed. Requiring a non-zero exit for the refusal reading, or a fixture pinning "error ⇒ exit 1", would close that.
- Keyless run: `ProviderAuthError` carries no `statusCode` and so classifies as `failed`, not `refused { auth }`. Unreachable via the resolver's availability check, but worth adding to `outcome.ts`'s status-less list.
- OpenCode also reads a project-level `opencode.json` from the workspace; a target repo shipping one could override the per-exec allow-all. The wall clock bounds a hang; `--auto` would be belt-and-braces. A #225 question.
- Baseline smell, acknowledged in the PR: `isRecord`/`readString` copied between `outcome.ts` and `stream-parser.ts` inside the one adapter directory — a small `opencode/wire.ts` would remove the intra-adapter copy without coupling adapters.

PR #246, opened 2026-09-07
Autonomy: route handlers nudge the sweep instead of running it, so one module graph sweeps (#163)

### LINKED ISSUE #163: Webhook-triggered sweeps run on the app-router graph — an inbound webhook can silence a standing fleet-health card, and two sweeps can claim one ticket
A watchdog that an inbound webhook can silence, and a single-flight guard that doesn't hold across module graphs.

Filed at the reviewer's request on PR #162 (finding 8): *"That is a watchdog that can be silenced by an inbound webhook, and it will be lost if its only record is a sentence in a CLAUDE.md paragraph. Please open the issue rather than merging on the note."* Discovered while fixing #159 and deliberately out of scope there.

## The mechanism

Next compiles `instrumentation.ts` — where the queue poll loop and the interval autonomy sweep run — and the app-router route handlers into **separate module graphs**, each with its own module-level state. #159's fix moved `activeTasks` to a `globalThis`-backed `processSingleton` for exactly this reason, but the sweep's own state was left per-graph.

`POST /api/webhooks/github` calls `runAutonomySweep()` directly, so it runs on the app-router graph against that graph's freshly-empty copies of `fleetHealthState`, `sweeping`, `inFlightClaims`, `saturationAnnounced`, `dailyCapAnnouncedDay` and every announce-dedup set — while `recordFleetHealth` **is** `globalThis`-backed.

## What that causes

1. **A standing needs-you card can be silenced by an inbound webhook.** The webhook sweep evaluates the #126 signals against debounce clocks that started that instant, computes all-clear, and overwrites the shared store — **dedup memory included**, so the card does not simply reappear on the next interval sweep.
2. **Two sweeps can claim the same ticket.** `sweeping` is per-graph, so a webhook sweep can run concurrently with the interval sweep — precisely what its single-flight comment promises cannot happen.
3. **`claimableSlots` overcounts** on that graph: it reads an empty `slotReservations`.
4. **Queue health reads false**: `isQueueRunning()` is `false` and `getQueueLastProgress()` is `null` on the app-router graph, because that graph never started the loop.

## Suggested shape (from the PR body, and the reviewer agrees it looks right)

Rather than move eleven more fields into `processSingleton`: have the webhook route **record "something changed"** and let the orchestrator's loop pick it up on the next tick. That removes the class of problem by construction instead of field by field, and it costs ≤30s of latency the spec already calls acceptable — *"the webhook is only latency on top of this — the sweep is the backbone."*

Moving the fields is the alternative. It keeps webhook latency at zero but needs all eleven done together, plus tests for concurrent sweeps, and it leaves the next per-graph field one commit away from reintroducing this.

## Acceptance

- A webhook-triggered sweep cannot clear or overwrite a fleet-health card that the interval sweep is standing behind, and cannot reset its dedup memory.
- Two sweeps cannot claim the same ticket concurrently — with a test that exercises the webhook path and the interval path against one ticket, not one path twice.
- Whatever shape is chosen, a **new** per-graph field cannot silently reintroduce the bug: either the state has one home, or the route no longer sweeps at all.

## Notes

- Same root class as #159 (one process, two module graphs), which is why #159's `src/lib/process-singleton.ts` doc comment and the module-graph convention in `CLAUDE.md` are the right places to read first.
- The `CLAUDE.md` bullet added by PR #162 describes this defect; it should cite this issue number once #162 lands (its current wording says both "needs filing" and "Filed, not fixed here").

lennons301:
Closes #163.

## What was wrong

`POST /api/webhooks/github` called `runAutonomySweep()` directly. Route handlers are compiled into a separate module graph from `instrumentation.ts`, so that sweep ran against the app-router graph's own, freshly-empty copies of `sweeping`, `inFlightClaims`, `fleetHealthState` and every announce-dedup set. Reproduced in production on 2026-09-06: moontide#123 claimed twice 250 ms apart, both attempts ran, two review passes. The same shape lets a webhook sweep compute all-clear health signals against an empty debounce clock and overwrite a standing needs-you card through the globalThis-backed health store.

## What changed

- **Route handlers no longer sweep.** New `src/lib/orchestrator/autonomy/sweep-nudge.ts` holds the one globalThis-backed slot (`processSingleton`, as #159 established). The webhook route and Discord arming call `requestAutonomySweep(reason)`; nothing outside the orchestrator's loop calls `runAutonomySweep` any more.
- **`startAutonomySweeps` is a 1 s ticker** that runs a sweep when the 30 s interval is due *or* a nudge is pending — never while one is in flight. A nudge is a level, not an edge: one landing mid-sweep is kept, so the tick after the sweep ends runs the follow-up rather than dropping the event. Webhook latency is at most ~1 s.
- **ESLint `no-restricted-imports`** forbids `src/app/**` from importing the sweep module, so the class of bug cannot come back through a new route (verified the rule fires on a probe file).
- `AGENTS.md`'s module-graph bullet now records the fix instead of "needs filing".

## Acceptance

- A webhook-triggered sweep cannot clear a standing fleet-health card or reset its dedup memory: there is no webhook-triggered sweep.
- Two sweeps cannot claim one ticket concurrently: `sweep-nudge.test.ts` drives the tick decision with the interval sweep in flight while the arming nudge lands (declines, nudge kept, next tick runs); `github-webhook-nudge.test.ts` drives the real route with the sweep module mocked and asserts it is never called and the nudge is recorded, for `issues.labeled ready-for-agent` and `issues.opened`, including two webhooks for one ticket collapsing to one nudge.
- A new per-graph field cannot reintroduce it: the route no longer sweeps at all, and the lint rule keeps it that way.

## Verification

`npx vitest run --maxWorkers=2`: 123 files, 2303 tests passing. `tsc --noEmit` clean. ESLint clean on changed files.


lennons301 (commented):
**Verdict: approve** — with one should-fix a human may want before merging (finding 1). Gated PR (`src/lib/orchestrator/autonomy/**` is `autonomy-control` in `docs/agents/review-gates.yaml`), so this is a comment-type review; a human merges.

The fix is the shape the ticket recommended and the reviewer on #162 agreed with: route handlers record a nudge on a `processSingleton` slot and only the orchestrator's loop ever sweeps. It holds by construction, and the single caller that already lived on the orchestrator graph (`armIssueFromDiscord`) is folded into the same path, which is a small improvement over what was there — a Discord arm that landed mid-sweep used to be silently dropped by the single-flight early return and wait for the next 30 s tick; now it is kept as a level and runs on the tick after.

## Acceptance criteria (issue #163)

- [x] **A webhook-triggered sweep cannot clear/overwrite a standing fleet-health card or reset its dedup memory.** Met by construction — there is no webhook-triggered sweep. `src/app/api/webhooks/github/route.ts:67,82` call `requestAutonomySweep` only; the `runAutonomySweep` import is gone. `github-webhook-nudge.test.ts` mocks the whole sweep module and asserts it is never called on `issues.opened` and `issues.labeled ready-for-agent`. Grepped `src/` for remaining callers: `runAutonomySweep` is now called only from `startAutonomySweeps` (boot + ticker) in `sweep.ts`; the only other importer of `sweep.ts` is `src/lib/discord/client.ts:247` (orchestrator graph, dynamic import of `armIssueFromDiscord`).
- [x] **Two sweeps cannot claim one ticket concurrently.** The ticker (`sweep.ts:261-273`) reads the module-level `sweeping` on the same graph that owns it, and `shouldSweepNow` refuses while `sweeping` is true regardless of `nudged`/due (`sweep-nudge.test.ts` "never starts a second sweep while one is in flight"). The nudge is consumed only inside `runAutonomySweep` after `sweeping = true` (`sweep.ts:298-300`), so one landing mid-sweep survives to the next tick (tested: "a nudge that landed mid-sweep is still pending"). There is no interleaving window: the ticker's `isSweepRequested()` read, the call, and `sweeping = true` are one synchronous frame. `sweeping` is reset in the `finally` at `sweep.ts:344`. See finding 2 on the letter of "a test that exercises the webhook path and the interval path against one ticket".
- [x] **A new per-graph field cannot silently reintroduce it.** The route no longer sweeps at all, which is the second of the two shapes the ticket allowed. The lint rule is a belt on top of that; see finding 1 for a hole in the belt.
- [x] **AGENTS.md bullet cites the issue and records the fix** instead of "needs filing" (`AGENTS.md:10` in the diff). `CLAUDE.md` is `@AGENTS.md`, so this is the right file.
- [x] **Only the ticket.** No unrelated changes. The `quotaGateAnnouncement` comment rewrite in `sweep.ts` is in scope (it described this defect).

## Findings, by severity

**1. Medium — the `no-restricted-imports` rule does not catch a dynamic `import()`, so "the class of bug cannot come back through a new route" (PR body, AGENTS.md, and the comment at `eslint.config.mjs:23-31`) is overstated.** I probed the rule via `eslint --stdin --stdin-filename src/app/api/probe/route.ts` without touching the tree:
   - static `@/lib/orchestrator/autonomy/sweep` → **error** (good)
   - static `../../../lib/orchestrator/autonomy/sweep` → **error** (good)
   - `const { runAutonomySweep } = await import("@/lib/orchestrator/autonomy/sweep")` → **clean, exit 0**
   - `@/lib/orchestrator/autonomy/sweep-nudge` → clean (good, the allowed path)

   `no-restricted-imports` only inspects `ImportDeclaration`/`Export*` nodes. The dynamic shape is not hypothetical here — `src/lib/discord/client.ts:247` reaches `sweep.ts` exactly that way, so it is the first thing someone copying an existing pattern would write in a route. The acceptance criterion itself is still met (the route does not sweep), so this is not a defect in the fix; it is a hole in the guardrail plus an inaccurate claim about it. Fix is one more rule in the same `files` block, e.g. `no-restricted-syntax` with a selector on `ImportExpression > Literal[value=/orchestrator\/autonomy\/sweep$/]` and the same message — or, if the belt is deliberately static-only, soften the three places that say the class of bug cannot come back.

**2. Low — the ticket asked for "a test that exercises the webhook path and the interval path against one ticket, not one path twice"; what landed is the two halves tested separately.** `github-webhook-nudge.test.ts` proves the route records and never sweeps; `sweep-nudge.test.ts` proves the pure `shouldSweepNow` declines while in flight. The seam where they meet — the ticker body in `startAutonomySweeps` (`sweep.ts:261-273`) reading `sweeping`/`lastSweepStartedAtMs`/`isSweepRequested()` and calling `runAutonomySweep` — is not exercised by any test (no test in the repo touches `startAutonomySweeps`, before or after this PR). By inspection it is correct, and extracting the decision into a pure function is a defensible way to make it testable given `sweep.ts`'s dependency surface. Noting it because the ticket was explicit and the PR body should say this was a deliberate substitution; a human accepting the PR should know the glue is unpinned.

**3. Low — `docs/runbook.md:118-119` is now stale.** "the `issues.labeled` webhook … kicks a sweep immediately" — it nudges, and the sweep starts on the next 1 s tick (or after the in-flight sweep ends). One sentence. The runbook's log-line grep at line 90 still matches, since the new boot line keeps the `Reconciliation sweep every 30s` prefix.

Nothing else. I checked and am satisfied on: the 1 Hz ticker is only started when `autonomyEnabled && isGitHubConfigured()` (`init.ts:537`), so `lastSweepStartedAtMs = 0` cannot make it spin `runAutonomySweep` at 1 Hz on a disabled fleet; `processSingleton` key `autonomy.sweepNudge` follows the documented owner-prefix convention; `MAX_REASONS`/dedup keep the log line bounded under a webhook burst; the route test's `vi.mock` of the sweep module is a string, so it does not trip the new lint rule on `src/app/**` test files. Smell baseline: nothing to raise — `sweep-nudge.ts` is a small module with an honest name and no speculative surface.

## What I ran

- `npx vitest run --maxWorkers=2` in the PR worktree (fresh `local.db`): **123 files, 2303 tests passed**, exit 0, 70.7 s.
- `npx tsc --noEmit`: clean.
- `npx eslint` on the seven changed files: 0 errors (1 expected "file ignored" warning for `AGENTS.md`).
- Four `eslint --stdin` probes against the new rule, results above.
- `gh pr checks 246`: no CI checks reported on this branch, hence the local runs.

lennons301:
Addressed the review's should-fix: the lint guard now also catches a dynamic `import()` of the sweep module (a `no-restricted-syntax` selector on `ImportExpression`, alias and relative forms; probed both, and `sweep-nudge` stays importable). Runbook §1 no longer says the webhook "kicks a sweep immediately".

On the low finding about the ticker glue in `startAutonomySweeps` being untested: left as is. Driving it needs the sweep module's config/GitHub guards mocked past the early returns and then the whole gather stubbed; the two halves it joins (`shouldSweepNow`, the nudge slot) are each tested, and the glue is five lines.

PR #244, opened 2026-09-06
Pin a single task or run to a named lane without moving the fleet's primary

lennons301:
Closes #241

## What this does

An operator can pin **one task or one run** to a named execution lane without moving the fleet's primary or its cost routing for anything else — the lever the milestone-6 proofs lacked, when the only way to run one ticket on `openai-api` was to pin production and watch every other ticket follow it onto the paid lane.

- **A pin is the operator's explicit lane, scoped down — not a new mechanism.** The fleet already has exactly one notion of "the operator chose this lane": the `primaryLane` override (or `AGENT_LANE`), which #172 honours even when the lane is unavailable and #176 treats as a pin that turns cost routing off. `src/lib/lanes/lane-pin.ts` (`settingsPinnedTo` / `overridesPinnedTo`) hands the crossing, ranking, money guards and resolver settings whose `primaryLane` is the pin, for that pass only. Everything that held for a fleet pin therefore holds for a task pin by construction: a walled pinned lane still fails over (a pin is a preference, not a refusal to move), a metered pin still answers to the confirm-once press and the cap by billing kind, a missing credential is refused rather than routed around, and #218's skill-capability filter runs against the pin.
- **Where the pin is read.** The turn manager reads `task.lanePin ?? run.lanePin` at pass start and on every follow-up turn (`readLaneCrossing`'s new `lanePin` parameter, `requirePassLane`'s pinned overrides); the queue judges an attended pickup against the task's pin; the claim spends a stored pin onto the run and says so in the claim comment. A pass pinned to a lane the file no longer declares is **refused naming the pin** (entry validates against the file, so this is reachable only when `lanes.yaml` changed under a stored pin) and takes the bounded interruption path an unavailable lane in force takes (#220) — never quietly run elsewhere.
- **Surfaces.** `POST /api/tasks` takes `lane`, judged at entry as the resolver judges the primary: undeclared → 400 naming the declared lanes; declared but unavailable → 409 naming the variable; a generation session pinned to a lane whose harness cannot invoke its skill → 409 `no-skill-capable-lane` naming that lane. Nothing is created on a refusal. `PUT/GET/DELETE /api/projects/:id/issues/:n/lane` is the operator surface for an unclaimed ticket (`lane_pins`, one row per ticket, spent once by the claim). The new-task form gains a lane picker fed by the settings screen's own lane read (absent when that read fails — an unpinned request is byte-identical to before). The run card says `pinned to <lane>` beside the harness.
[...]

Issue #241, opened 2026-09-06
Pin a single task or run to a named lane without moving the fleet's primary


lennons301:
## What to build

An operator can pin **one task or one run** to a named execution lane at creation or arming time, without changing the fleet's primary lane or cost routing for anything else.

Today the only ways to get work onto a non-default lane are fleet-wide: set the primary lane (Settings → Execution lane, or `AGENT_LANE`), or wait for cost routing / failover to choose it. Both move *every* pass. The one per-run control that exists (#202, "move to paid lane") applies only to a run already parked on a wall. So proving or exercising a lane that spends real money — the case that prompted this: running the #224 Codex proof on `openai-api` with a very small credit balance — currently means deviating the whole fleet onto the paid lane for the duration.

Requested by the owner on 2026-09-06 (fleet Discord channel) during the milestone #6 proofs:

> it would be good if a task can run on a specified lane that deviates from the default … you could specify the openai lane for your task only, not deviate the running fleet to a paid lane, when we have a very small credit balance

### Shape

- **Operator-chosen, never ticket-chosen.** The lane comes from a trusted surface — the session-entry form / `POST /api/tasks` for interactive and generation sessions, and an operator action (UI or API) when arming a ticket-loop run. It is *not* a `## Workflow` directive: the ticket contract (#197) and spec #213 keep the lane out of semi-trusted issue bodies, and that stays true.
- **A pin for that task/run only.** Recorded on the task (and run) rows the way `lane` already is; the lane resolver honours it ahead of the fleet's primary/cost routing for that pass and its follow-up turns; every other pass routes as before.
- **Same guards as any other crossing.** A metered pin still answers to the confirm-once gate and the real-money cap (#174) and to the generation-session capability filter (#218); an unavailable pinned lane is refused with the resolver's own reason before a container is provisioned, exactly as an unavailable primary is.
- **Visible.** The run card and lane rows show the pin (the harness stamp from #223 already names the adapter), so a pinned run is never mistaken for a routing decision.

## Acceptance criteria

- [ ] A task created with an explicit lane runs its passes on that lane while the fleet's primary and cost routing are unchanged for every other task
- [ ] A ticket-loop run can be armed with an explicit lane by an operator surface; the ticket body cannot set one (existing directive parser behaviour, pinned by a test)
- [ ] A pinned metered lane still requires the day's real-money confirmation and stays under the cap; a pinned lane the session's harness cannot host is refused at entry with the reason
[...]

PR #247, opened 2026-09-07
Operator gate: HTTP Basic auth on every page and API route, dopplerToken never served (#242)

### LINKED ISSUE #242: The orchestrator's API is unauthenticated: a local pass changed production's lane, and anyone can
## What happened

During the milestone-6 proofs (#224, #225) a *local* ticket-loop pass changed **production's** execution lane by calling `PATCH https://interludes.co.uk/api/settings/overrides` — the same public route the settings screen uses — and the owner's own tickets then ran on a paid lane until the override was put back by hand (2026-09-06, LPS #266 spent ~$0.92 on `openai-api`). The pass did nothing clever: every API route on the deployed orchestrator answers to anyone who can reach the host, because none of them authenticate (`auth: none` is the product's declared choice, and Caddy fronts the routes without a gate).

That means, today, from any machine on the internet:

- `PATCH /api/settings/overrides` can pin the fleet to a metered lane, raise the real-money cap up to the code ceiling, or change tiers.
- `POST /api/settings/metered-spend` (the confirm-once press) can authorise the day's real-money spend.
- `POST /api/runs/:id/lane-move` (#202) can move a parked run onto a paid lane.
- `POST /api/tasks` can create tasks — and, after #241, pin them to a paid lane.
- `GET /api/projects` serves each project's stored Doppler token in cleartext (recorded earlier in the prod findings; still open).

The money guards bound the damage (a cap, a daily confirmation) but they are operated through the same unauthenticated surface.

## What to build

Authentication on the orchestrator's HTTP surface, so that a mutating route is reachable only by the operator, and a local development process cannot reach production by accident.

Shape to decide (grill first; this is decision-shaped):

- **Where the gate lives.** Simplest: Caddy `basic_auth` (or forward-auth) on `/api/*` and the app pages, exempting only what must stay open — `POST /api/webhooks/github` (verified by its own signature), `/api/internal/validate-subdomain` (Caddy's on-demand TLS ask), the `task-*` preview subdomains, and the fleet SSE stream if the dashboard is to stay a glanceable page. Alternative: an app-level bearer token from Doppler checked in one place (a Next middleware / `custom-server.js`).
- **Read vs write.** At minimum every mutating route; probably reads too, given the token leak.
- **The dopplerToken leak** on `/api/projects` is fixed regardless — a secret is never served, only "set / not set".
- **Local runs never target production.** The proof tickets' habit of pinning production to run a real ticket should be replaced by #241's per-task pin (a local dev-mode instance, or a pin scoped to one run), and the runbook should say so.

## Acceptance criteria

- [ ] A request to any mutating `/api/*` route without the operator credential is refused (401/403) on the deployed stack; the settings screen, task entry and lane-move controls still work for the operator
- [ ] The GitHub webhook, the subdomain validator and preview subdomains remain reachable as they are today
- [ ] `GET /api/projects` (and any other read) never returns a stored secret's value
- [ ] The runbook documents the credential and how a local session is expected to reach production (or not)
- [ ] A test pins that the exempt paths are exactly the ones listed and nothing else

## Workflow

model: heavy

## Blocked by

- None — can start immediately (a design conversation first: where the gate lives)

lennons301:
Closes #242.

> **Before merging: set `OPERATOR_PASSWORD` in Doppler `interlude/prd`.** The deploy reads it and refuses to run without it (the stack is already down at that point, exactly like a missing `DOMAIN`), and the app fails closed — every gated route answers 503 until it is set. A long random string; user is fixed as `operator`. `products/interlude.yaml` in the platform repo still says `auth: none` and wants updating separately.

## What was wrong

Every page and API route on the deployed orchestrator answered to anyone who could reach the host. On 2026-09-06 a *local* ticket-loop pass changed **production's** execution lane through the public `PATCH /api/settings/overrides`, and the owner's tickets ran on a paid lane until it was put back by hand. The metered cap, the confirm-once press, the lane move and task creation were all operated through the same open surface, and `GET /api/projects` served each project's Doppler token in cleartext.

## Decision: where the gate lives

App-level, in Next's `src/proxy.ts`, not Caddy — one credential, HTTP Basic (`operator` / `OPERATOR_PASSWORD`). Reasons: the exempt list has to be pinned by a test, which a Caddyfile cannot be; Basic gives the browser flow for free (one prompt, then every same-origin fetch and the dashboard's EventSource carry it) and `curl -u` is all a script needs; and a local process holds no credential for production, which is the "local never targets production" property the ticket asks for, by construction. Reads are gated too, given the token leak.

## What changed

- `src/lib/operator-auth.ts` — the pure decision: exempt set, Basic parsing, constant-time compare (both halves, no short-circuit on the username), fail-closed rule.
- `src/proxy.ts` — applies it to every request. No `matcher`, deliberately: a second list the test cannot see.
- **Exempt, exactly:** `POST /api/webhooks/github` (its own HMAC), `/api/internal/validate-subdomain` (Caddy's on-demand TLS ask), `/manifest.webmanifest` (fetched credential-less by spec; not secret). Preview subdomains never reach Next — `custom-server.js` routes them by Host header — so they stay reachable as before.
- **Fails closed:** production build with no password → 503 on every gated route. Development with it unset → open, so `pnpm dev` is unchanged; set it locally to try the gate.
- `GET /api/projects` masks `dopplerToken` (`••••••••` / `null`) exactly as `GET /api/projects/[id]` always did; the settings UI only tests it against null.
- `deploy.yml` reads `OPERATOR_PASSWORD` from Doppler beside `DOMAIN`, fails without it, and the health check now proves both halves: the operator's request gets 200 **and** an anonymous request gets 401 (a 200 there would mean production is open). The verbose failure dump redacts the header.
- Runbook: new "The operator credential" reference section (browser, scripts, exemptions, fail-closed, rotation), the local-never-reaches-production rule, env-table row, and `-u` on the production `curl` examples. `AGENTS.md` convention bullet.

## Acceptance

- Unauthenticated request to any mutating (or reading) `/api/*` route is refused 401 with a Basic challenge; the settings screen, task entry and lane-move controls work for the operator because the browser carries the credential after one prompt.
- Webhook, subdomain validator and preview subdomains unchanged.
- No read returns a stored secret's value (`projects-redaction.test.ts`).
- Runbook documents the credential and how a local session reaches production (it does not).
- `operator-auth.test.ts` pins the exempt set entry by entry and asserts every money/state route is gated.

## Verification

`npx vitest run --maxWorkers=2`: 124 files, 2309 tests passing. `tsc --noEmit` clean. ESLint clean on changed files. `src/proxy.ts` is the Next 16 convention (`isProxyFile` accepts `/src/proxy`); it runs on the Node runtime.


lennons301 (commented):
**Verdict: approve** — sound, complete against every acceptance criterion, verified with a production build and live smoke run. Human merges (this PR touches `.github/workflows/**` and is the auth gate itself; auto-merge is off). **Precondition before merging: `OPERATOR_PASSWORD` must exist in Doppler `interlude/prd`.** The deploy runs `compose down` and `build` *before* it reads the secret, so a merge without it takes production down until the value is set and the workflow is re-run.

## Acceptance criteria

- [x] **Unauthenticated mutating (and reading) `/api/*` refused; operator controls still work.** Production smoke (`next start`, `OPERATOR_PASSWORD` set): `PATCH /api/settings/overrides`, `POST /api/settings/metered-spend`, `POST /api/tasks`, `POST /api/runs/x/lane-move`, `GET /api/projects`, `GET /api/fleet/stream`, `GET /`, `GET /_next/static/chunks/*.js` all 401 with `www-authenticate: Basic realm="interlude", charset="UTF-8"` and `cache-control: no-store`; same requests with `-u operator:<pw>` 200 (fleet stream 200 `text/event-stream` with a `fleet` event). Wrong password, wrong user, `Bearer`, and malformed Basic all 401. No `fetch` in `src/` sets `credentials:` and both `EventSource`s are same-origin (`src/components/fleet/fleet-dashboard.tsx:26`, `src/components/task-stream.tsx:98`), so the browser carries the cached credential. Caveat on installed PWAs in finding 2.
- [x] **Webhook, validator, preview subdomains unchanged.** Smoke: `POST /api/webhooks/github` anon → 404 "not configured" (the route's own answer, no `WWW-Authenticate`); `GET /api/internal/validate-subdomain?domain=x` anon → 403 (route's own). Preview subdomains: Caddy's wildcard block proxies to `app:3000`, `custom-server.js` matches `SUBDOMAIN_RE` on the Host header and proxies to the container before Next is involved — confirmed by reading; nothing in the diff touches that path. Caddy's `ask` URL is the exempt validator.
- [x] **No read serves a stored secret.** `src/app/api/projects/route.ts:11-17` masks like `[id]` always did; `projects-redaction.test.ts` asserts the literal never appears. Grepped every `dopplerToken` reference in `src/` outside tests: the only serialisations are the list route and `[id]` GET/PATCH (all masked); `container-manager.ts` only injects it into the agent env, `turn-manager.ts` only tests truthiness.
- [x] **Runbook documents the credential and the local→production rule.** `docs/runbook.md` new "The operator credential" section, env-table row, `-u` on the production `curl` examples (the only two in the file), and the explicit "a local session never reaches production" rule pointing at #241's per-task pin.
- [x] **A test pins the exempt set.** `src/lib/__tests__/operator-auth.test.ts:31-39` `toEqual`s the sorted `EXEMPT_PATHS` entry by entry, and there is no `matcher` export in `src/proxy.ts` — the built `functions-config-manifest.json` registers `/_middleware` with matcher `^.*$`, so there is genuinely one list.

## Findings (by severity)

1. **Medium — merge-order precondition, operational not code.** `.github/workflows/deploy.yml:105-109` exits after `compose down`/`build` if Doppler has no `OPERATOR_PASSWORD`. The PR body warns; repeating it here because the human merging is the one who has to set the secret first. Also worth knowing: the health check now needs a `-u` 200 *and* an anonymous 401, so if the gate were ever not in the built image the deploy fails loudly rather than reporting green — good.

2. **Medium — browser flow on an installed PWA is unverified.** `src/app/layout.tsx:9` sets `appleWebApp.capable` and `manifest.ts` is `display: standalone`. iOS home-screen web apps historically cannot show an HTTP auth prompt (the standalone WebView has no UI for it) and just fail to load; Chrome/Android installed PWAs share the profile's HTTP auth cache but that cache is session-scoped, so the prompt returns on each cold start. The PR body's "one prompt, then everything carries it" holds for a browser tab; it does not address the standalone case, and this app is used from a phone. Not provable from here — please verify on the device after deploy. If standalone fails, opening in Safari/Chrome works today; a cookie set from the Basic prompt would be the follow-up. Related, low: on a lapsed credential the fleet dashboard's `onerror` (`fleet-dashboard.tsx:31`) reads as "offline", not "sign in again".

3. **Low — `.env.example` not updated.** Every other orchestrator variable is documented there (`DOMAIN`, `AUTONOMY_ENABLED`, `REVIEWER_GH_TOKEN`, ...); `OPERATOR_PASSWORD` is only in the runbook and AGENTS.md. Add a commented entry so the file stays the exhaustive reference it currently is.

4. **Low (suggestion) — `constantTimeEqual` leaks password length.** `src/lib/operator-auth.ts:121-128` loops `max(a.length, b.length)`, so response time reveals whether the presented string is longer than the real one; an attacker ramping lengths learns the password length. Negligible for a long random string, but the standard fix is trivial and available: hash both sides with SHA-256 and `crypto.timingSafeEqual` (fixed 32 bytes, no length signal). The "does not depend on `node:crypto` so the decision stays runtime-agnostic" rationale is Speculative Generality — Next 16 hard-codes "Proxy always runs on Node.js runtime" (`node_modules/next/dist/build/analysis/get-page-static-info.js:552`).

5. **Low (informational) — the webhook exemption is path-only, not `POST`-only.** AGENTS.md, the runbook and the module docs say `POST /api/webhooks/github`; `isExemptPath` ignores method, so `GET /api/webhooks/github` also bypasses the gate (Next answers 405 today because the route exports only `POST`, so nothing leaks). Either say "any method" in the docs or check the method — otherwise a future `GET` export on that route is silently public.

6. **Observation, pre-existing and out of scope.** `custom-server.js` handles WebSocket upgrades for the path-based preview (`/api/tasks/:id/preview/*`, `PREVIEW_PATH_RE`) before Next sees them, so that upgrade path is ungated while its HTTP half now is. It is equivalent to the subdomain preview the ticket explicitly keeps open and needs a task ULID; noting so it is not mistaken for something this PR missed.

7. **Platform follow-up, not this PR.** The estate `standards/review-gates.yaml` auth category gates `**/middleware.*` only; Next 16's rename to `proxy.*` is not covered, so a future PR touching only `src/proxy.ts` would arm auto-merge. This PR is gated anyway via `.github/workflows/**`. Recommend adding `**/proxy.*` to the platform gates (or interlude's extension). `products/interlude.yaml` `choices.auth: none` also needs updating, as the PR body says.

Bypass probes that were fine: `/api/webhooks/github/../projects` (`--path-as-is`) → 401; `//api/projects` and `/api/webhooks/github/` → 308 to the normalised path (then gated); `/API/webhooks/github` → 401; query-string decoys → 401. The built chunk reads `process.env.OPERATOR_PASSWORD??null` at request time (not inlined at build), so a production image cannot be stuck at 503 by a build-time value. Smells: none beyond finding 4. Branch is behind `main` by #244 but GitHub reports mergeable; the route #244 added (`/api/projects/[id]/issues/[number]/lane`) is gated by default.

## What I ran

- `npx vitest run --maxWorkers=2` — 123 files, 2303 tests, all pass. The three new files run verbosely: 17/17.
- `npx eslint` on the six changed `.ts` files — clean. `npx tsc --noEmit` — clean.
- `NODE_OPTIONS=--max-old-space-size=2048 npx next build` — success, `ƒ Proxy (Middleware)` listed; `.next/server/functions-config-manifest.json` registers `/_middleware`, `runtime: nodejs`, matcher `^.*$`; `.next/standalone/.next/server/middleware.js` present and wired to the chunk containing the gate.
- Production smoke, `next start` on a scratch DB, three runs: password set (table above), unset → 503 on `/`, `/api/projects`, and even with a credential; exempt `/manifest.webmanifest` 200 and validator 403 still served; empty-string password → 503.
- `gh pr checks 247` — no CI runs on PRs in this repo (deploy is push-to-main only), so the above is the objective signal.

lennons301:
Taken from the review:

- **Finding 4** — `constantTimeEqual` now hashes both sides (SHA-256) and compares the digests with `crypto.timingSafeEqual`, so nothing about the password's length is observable. `proxy.ts` is Node-runtime, as you confirmed from the build.
- **Finding 3** — `OPERATOR_PASSWORD` documented in `.env.example` beside the other orchestrator secrets.

Left for the operator, as the review says: **finding 1** (set `OPERATOR_PASSWORD` in Doppler `interlude/prd` before merging — the PR body leads with it), **finding 2** (verify the installed PWA on the phone after deploy; if iOS standalone cannot show the Basic prompt, the fallback is opening the site in Safari once so the credential is stored), and **finding 7** (platform repo: gate `**/proxy.*` in the estate `review-gates.yaml`, and `products/interlude.yaml` `auth: none`). Finding 5 is as described — the exemption is by path and the route only serves POST, so a GET there is a 405 with no body worth protecting.

Issue #189, opened 2026-09-02
Deploy reports success while deploying nothing: an anonymous `git pull` 401s and the script carries on


lennons301:
## Problem

`deploy.yml` deployed nothing twice in a row and reported **success** both times. Prod ran the 08:03 image for 4.5 hours while `main` moved two commits ahead (#187, #188), and nothing anywhere said so.

From the run log for `22cf0d9` (run `33630264127`, conclusion **success**):

```
fatal: could not read Username for 'https://github.com': No such device or address
fatal: expected flush after ref listing
 Container interlude-app-1 Removed
 Image interlude-app Building
#7 CACHED … #25 CACHED
```

Three faults compound:

1. **The pull is anonymous.** `/opt/interlude`'s `origin` is `https://github.com/lennons301/interlude` with no credential helper, no `~/.git-credentials` and no `~/.gitconfig`. The repo is public so this normally works — and GitHub 401s it under unauthenticated rate limiting. That box generates a lot of anonymous git traffic (every pass clones the plugin marketplace over HTTPS: `SSH not configured, cloning via HTTPS`), so recurrence is likely, not hypothetical.
2. **`git pull` failing is not fatal.** The `script:` block has no `set -e`, so every later step runs against the *old* checkout.
3. **A cached build is indistinguishable from a real one.** `docker compose build app` on unchanged source is a 100% cache hit that reproduces the *identical* image — same ID, same `CreatedAt` (08:03:54). `up -d` then recreates the containers and the TLS health check passes, because the old app is perfectly healthy. Every signal the job has says "deployed".

Evidence on the box: `git log -1` = `cd626b2`; reflog's last pull = `08:02:48`; `docker images interlude-app` `CreatedAt` = `08:03:54` while the app container was `Created` at `12:30:15`.

## What to build

- `set -euo pipefail` in the deploy script, so a failed pull stops the deploy instead of decorating it.
- **Assert the deployed SHA.** Pass `github.sha` in via the action's `envs:` and fail if `git rev-parse HEAD` doesn't equal it after the pull. This is the check that makes the whole class loud: a cached build, a stale checkout and a partial pull all fail it.
- Retry the pull a few times with backoff before giving up, since the failure mode is transient rate limiting.
- Consider authenticating the pull (a deploy token passed through `envs:`, or an SSH remote) so the box isn't competing with its own agent containers for the anonymous rate limit.

## How to tell it's fixed

[...]

PR #248, opened 2026-09-07
Discord inbound: watch the gateway, surface handler failures, and collect blocked answers over REST (#135)

### LINKED ISSUE #135: Discord inbound can go deaf silently — a blocked run's answer is lost with no trace
## Problem

Everything that reaches Interlude **from** Discord — a blocked run's answer, an arming "yes", a ✅ completion, a new-task message — arrives on exactly one path: the discord.js gateway wired up in `startDiscordBot` (`src/lib/discord/client.ts:18`). Everything Interlude **sends** to Discord goes out over REST through the cached client. The two halves fail independently, and only the outbound half is observable — embeds keep landing in the fleet channel while inbound is stone deaf, so the fleet looks healthy while every human reply falls on the floor.

Nothing watches the gateway. There is no handler for `shardDisconnect`, `shardError`, `shardResume`, `error` or `invalidated`; `clientReady` (`client.ts:34`) is the only lifecycle event handled, and all it does is cache the client for outbound use. A session that drops and re-IDENTIFYs does not replay what it missed, and discord.js does not recover an `invalidated` session at all. Both are silent.

The handler's own failures are equally silent: `client.on("messageCreate", …)` (`client.ts:39`) swallows everything into `console.error`. A human whose reply throws mid-handler gets no reaction, no answer, and no trace they will ever see.

A lost reply has no second route in. `parkBlockedRun` (`turn-manager.ts:1266`) stops the container, posts the question, stores the embed id as `tasks.discordMessageId` and leaves the run `blocked` — a member of `ACTIVE_RUN_STATUSES` (`sweep.ts:132`), so the sweep never re-claims it, the reaper never touches its container, and the reducer skips it by design ("A run escalated as blocked is driven by exactly one thing — its question"). If the answer never becomes a `messages` row, the run waits forever behind an amber needs-you card that looks identical to one nobody has answered yet.

## Observed instance

`lennons301/last-person-standing` #189, 2026-08-13 — run `01KZXN62GXB6WXX8GKHNZQZK39`, attempt 2/3, task `01KZXN66QX2ZK1YVMWHHNW6MHM`, PR #198:

- **13:45:32Z** — implement pass parked `blocked`; question posted to the fleet channel (LPS has no linked channel of its own), embed `1537457104976085155` stored as the task's `discordMessageId` at 13:45:38.
- **14:06:38Z** — the owner replied to that embed. Verified afterwards over the Discord REST API: `message_reference.message_id` = `1537457104976085155`, non-bot author, content intact. A correct reply to the right message.
- The reply **never became a `messages` row**, and the bot placed **no reaction** on it — neither the 👍 of a successful `handleReply` nor the ❌ of a terminal task. Every silent-return path in `handleReply` (`client.ts:182`) is excluded: the task was found by `discordMessageId`, its `kind` is `implement` not `triage`, and `blocked` is not in the terminal list.
- The process was **up and healthy** at that moment: the moontide #50 repair task held `status: running` from 13:36 to 14:34, and any restart in that window would have force-failed it through `recoverOrphanedTasks`. Outbound was working throughout — eleven saturation embeds posted to that same channel between 14:12 and 17:48.
- The run then sat `blocked` for **6.5 hours** while the owner believed he had answered. Recovery meant posting the same text by hand through `POST /api/tasks/{id}/messages`; delivery then worked on the very next 2s poll (task `running` six seconds later, turn 2 under way). The loss was entirely on the inbound edge — every other seam was healthy.

Cost: an attempt's slot idle all evening and the ticket silently absent from the frontier, with the dashboard showing an amber card indistinguishable from an unanswered one.

## What to build

Four layers. The last is the one that makes the loss impossible rather than merely visible.

1. **Instrument the gateway.** Handle `shardDisconnect`, `shardError`, `shardResume` and `error` in `startDiscordBot`, logging shard id and close code; on `invalidated`, re-login rather than sit deaf (discord.js will not do it). Track the timestamp of the last received gateway event — any event counts — and export it for the watchdog below.

2. **Stop swallowing handler failures.** `messageCreate`'s catch should make a failed reply visible to the human who wrote it: best-effort ⚠️ reaction, and when the message resolves to a task, a system message on that task saying the reply was seen and failed. Silence is the one outcome a human cannot act on.

3. **A gateway-liveness signal in the health watchdog.** Add a fourth signal to `evaluateFleetHealth` (`src/lib/fleet/health.ts`, joining `owedReviewStalls` / `pickupWedged` / `queueStale`), fed by the last-gateway-event timestamp: stale beyond threshold while the client is meant to be connected → one dashboard needs-you card plus one Discord ping. Outbound still works when inbound is dead, which is exactly why the ping lands.

4. **Reconcile blocked answers over REST — poll, don't trust push.** On each sweep, for every `blocked` run whose task carries a `discordMessageId`, `GET /channels/{channelId}/messages?after={discordMessageId}` and adopt any non-bot reply referencing that id which has no matching `messages` row — inserting exactly as `handleReply` does, then reacting 👍. This is the estate's own principle (the tracker is the coordinator; poll rather than trust delivery) applied to the one push channel the loop still depends on, and it self-heals all three loss modes at once: a dead gateway, a swallowed throw, and a reply that arrived while the process was down.

   Keep it bounded: blocked runs only, messages after the stored question only, replies that reference the stored id only. Nothing is ever adopted from general channel chatter, and an already-adopted reply is never delivered twice.

## Seams under test

- `evaluateFleetHealth` (`src/lib/fleet/health.ts`) — pure; the new inbound-staleness signal, its since-timer and its announce-once behaviour, exactly as the three existing signals are tested.
- A new pure selector for layer 4 — given the fetched channel messages, the stored `discordMessageId` and the task's existing `messages` rows, return the replies to adopt. The Discord fetch and the DB insert stay in a thin executor around it, so idempotency (no double delivery) and scoping (only replies to the question) are testable without a gateway.
- The gateway instrumentation itself is I/O; assert only that a received event advances the exported timestamp, and that `invalidated` triggers a re-login attempt.

## Acceptance criteria

- [ ] `shardDisconnect`, `shardError`, `shardResume`, `error` and `invalidated` are handled and logged; `invalidated` re-logs-in.
- [ ] A received gateway event advances a last-inbound-event timestamp exposed to the health watchdog.
- [ ] Inbound staleness past threshold raises one needs-you card and fires exactly one Discord ping per occurrence (not once per sweep).
- [ ] A `messageCreate` handler error leaves a human-visible trace: ⚠️ reaction, plus a system message on the task when one is identifiable.
- [ ] A human reply to a blocked run's question is delivered even when the gateway never sees it, via the sweep's REST reconciliation.
- [ ] Reconciliation is idempotent — a reply already delivered by the gateway is never delivered a second time — and adopts nothing that is not a reply to the stored question id.
- [ ] `health.test.ts` covers the new signal; the reconciliation selector has its own tests for: fresh reply → adopt; already-present reply → skip; unrelated channel message → skip; bot message → skip.

## Related

- #19 — the BLOCKED escalation and the Discord question loop this protects.
- #126 — the fleet-health watchdog whose fourth signal this becomes; the parent #115 lesson (a wedge hidden behind a healthy-looking dashboard) is the same failure shape.
- Sibling, filed separately: a blocked run's answer is deliverable only while the process that parked it lives — the in-memory half of the same fragility.

lennons301:
Closes #135.

## What was wrong

Everything that reaches Interlude *from* Discord — a blocked run's answer, an arming "yes", a ✅, a new task — arrives over one gateway session, and nothing watched it. Outbound (REST) fails independently and was the only observable half. On 2026-08-13 (LPS #189, run `01KZXN62…`) the fleet posted eleven embeds to the same channel while a correct reply to a blocked run's question never became a `messages` row and got no reaction; the run sat `blocked` for 6.5 hours behind a card indistinguishable from an unanswered one. `blocked` is an active run status, so nothing re-claims it and nothing else can deliver the answer.

## What changed — the four layers the ticket asks for

1. **Gateway instrumented** (`src/lib/discord/client.ts`). `shardReady`, `shardResume`, `shardReconnecting`, `shardDisconnect` (with the close code named), `shardError` and `error` are handled and logged; every delivered event stamps `src/lib/discord/gateway-health.ts` (globalThis-backed via `processSingleton`, because sends happen from both module graphs), and `sendWithRetry` stamps a successful send there too. **Nothing re-logs-in, deliberately.** Checked against discord.js 14.26.4 source: `shardDisconnect` is emitted only for an unrecoverable close code — 4004 bad token, 4010–4014 shard/version/intent errors, "will no longer reconnect" — every one a configuration error that a re-login with the same token and intents would only repeat, IDENTIFYing toward Discord's daily limit (which resets the token). Recoverable closes the library resumes itself (`shardReconnecting`). So a terminal close is **recorded** and the watchdog acts on it at once (layer 3). The ticket's `invalidated` exists in the `Events` enum but is never emitted by this version (`@discordjs/ws` handles INVALID_SESSION itself); it is logged if it ever arrives. `startDiscordBot` takes an injectable client factory so the wiring is testable with a fake client.

2. **Handler failures are visible.** A `messageCreate` handler that throws now leaves a ⚠️ on the human's message (best-effort) and, when the message resolves to a task, a system message on that task saying the reply was seen and failed — instead of a `console.error` nobody reads.

3. **Deaf-gateway health signal** (`evaluateFleetHealth`, sixth signal `discordInboundStale`, two causes). **`closed`**: discord.js gave the shard up (layer 1) — fires on the next sweep with no threshold, names the close code, and the card/ping say "fix the bot's configuration and restart". **`silent`**: the observation is the **echo**: the bot's own message comes back as a `MESSAGE_CREATE` within a second on a healthy session, so "we sent and the gateway has delivered nothing since" for `DISCORD_INBOUND_STALE_MINUTES` (default 5, `budgets.ts`/`config.ts`) is precise. It cannot fire on a quiet fleet (nothing sent, no echo owed), a fresh session starts owed nothing for sends that predate it, and it fires on a zombie session whose heartbeats still ACK — which discord.js cannot detect. One red `discord deaf` needs-you card (`fleet-view.ts`, `needs-you.tsx`), one Discord ping per spell (`notifyDiscordInboundStale`; outbound still works, which is why it lands), cleared the moment an event arrives after the send.

4. **Blocked answers collected over REST** (`src/lib/discord/blocked-replies.ts`, called at the top of every sweep). For every `blocked` task with a stored question id: fetch the channel's messages after that id (project channel, else fleet channel — the same resolution `parkBlockedRun` used), adopt any non-bot reply that **references** the question and is not already a row, react 👍. `selectRepliesToAdopt` is the pure selector; the executor only fetches and inserts, bounded by `DISCORD_REST_TIMEOUT_MS`, never throwing. **Idempotency:** both paths write through one helper, `insertDiscordAnswer`, which puts the Discord message id in the content JSON beside `text` (no schema change — every reader takes `text`) and checks for it before inserting, synchronously, so neither path can deliver what the other already has; an exact-text fallback covers rows the gateway wrote under the previous build. Bounded as the ticket asks: blocked tasks only, after the question only, replies to the stored id only.

## Acceptance

- [x] `shardDisconnect`/`shardError`/`shardResume`/`shardReconnecting`/`error` handled and logged. **Deviation from the ticket's "`invalidated` re-logs-in", on evidence:** this discord.js never emits `invalidated`, and its terminal event (`shardDisconnect`) only ever carries a configuration error a re-login cannot fix and would storm on — so a terminal close raises the deaf card immediately with the fix named instead. `gateway.test.ts` drives a fake client through `shardDisconnect` (4014) and asserts the close is recorded, no second client is created, and a fresh `clientReady` clears it; through `shardReconnecting` (nothing); and through `invalidated` (logged).
- [x] A received event advances the inbound timestamp — `gateway.test.ts`.
- [x] Staleness raises one card and exactly one ping per occurrence — `health.test.ts` ("deaf Discord gateway"): fires, dedups across sweeps, clears on recovery, re-arms.
- [x] Handler error → ⚠️ plus a system message on the task — `gateway.test.ts`, directly and through `messageCreate`.
- [x] A reply the gateway never saw is delivered by the sweep — `blocked-replies.test.ts` over a real migrated DB with a stub channel.
- [x] Idempotent and scoped — fresh → adopt; already-present by id → skip; already-present by text (pre-upgrade row) → skip; unrelated / reply-to-other → skip; bot → skip; a second sweep and a late gateway delivery both insert nothing.
- [x] `health.test.ts` covers the signal; the selector has its own tests.

Runbook §4 explains the ⚠️, the REST collection and the `discord deaf` card; env table gains `DISCORD_INBOUND_STALE_MINUTES`; `AGENTS.md` records the four layers.

## Verification

`npx vitest run --maxWorkers=2`: 123 files, 2316 tests passing. `tsc --noEmit` clean. ESLint clean on changed files. The stub channel in `blocked-replies.test.ts` returns a Map keyed by id, the shape of a discord.js `Collection`, so the executor's `.values()` read is what is exercised.


lennons301 (commented):
**Verdict: request-changes**

Layers 2, 3 and 4 are sound and well-tested; the REST reconciliation in particular is correctly scoped and idempotent. Two things need fixing before this lands: the `invalidated` recovery path is wired to an event discord.js 14.26.4 never emits (and the docs added here assert the opposite), and the new failure trace can tell a human "nothing was delivered — reply again" when the reply *was* delivered. Merge state: not armed (`autoMergeRequest: null`, no labels); the PR touches gated paths (`src/lib/config.ts`, `src/lib/orchestrator/autonomy/**`, `src/components/**`) so a human merges regardless.

## Acceptance criteria

- [~] **`shardDisconnect`/`shardError`/`shardResume`/`error`/`invalidated` handled and logged; `invalidated` re-logs-in.** Handlers exist and the event names/argument shapes match `node_modules/discord.js/typings/index.d.ts` (14.26.4, lines 6053-6142). Met to the letter, not in substance — see finding 1: `invalidated` is never emitted by this discord.js, the re-login path is unreachable, and the `shardDisconnect` log says the opposite of what that event means in v14.
- [x] **A received gateway event advances a last-inbound timestamp exposed to the watchdog.** `recordDiscordInbound()` on `shardReady`/`shardResume`/`messageCreate`/`messageReactionAdd`; `observeDiscordGateway()` read by `gatherFleetHealthInput` in `sweep.ts`. `gateway.test.ts` drives it through a fake client.
- [x] **Staleness raises one card and exactly one ping per occurrence.** `health.ts` (f) block; `discordInboundAnnounced` boolean mirrors `queueStaleAnnounced`; state threaded through `fleetHealthState` in `sweep.ts` like the other signals. `health.test.ts` covers null bot / quiet fleet / answered / below threshold / fire-once / dedup across sweeps / clear / re-arm.
- [~] **Handler error leaves a human-visible trace: ⚠️ plus system message when a task is identifiable.** Implemented and tested (`reportDiscordHandlerFailure`). But the trace can be false — finding 2.
- [x] **A reply the gateway never saw is delivered by the sweep.** `reconcileBlockedReplies` at the top of `runAutonomySweep`; `blocked-replies.test.ts` over a real migrated DB with a Map-shaped stub channel; the adopted row has `deliveredAt` null so the queue's `role='user' AND deliveredAt IS NULL` pickup (`queue.ts:414-424`) delivers it.
- [x] **Idempotent and scoped.** Both paths write through `insertDiscordAnswer`, which checks by Discord message id and inserts synchronously with no await between (single-threaded, so the gateway and REST paths cannot interleave). Scope: `status='blocked'` + `discordMessageId IS NOT NULL` + `referencedMessageId === questionMessageId` + non-bot + non-empty; channel resolution mirrors `parkBlockedRun` (project channel, else fleet). One over-skip in the text fallback — finding 3, minor.
- [x] **`health.test.ts` covers the signal; selector has its own tests** for fresh → adopt, present-by-id → skip, present-by-text → skip, unrelated / reply-to-other → skip, bot → skip, plus empty and dedupe/ordering.

## Findings

**1. [High] `invalidated` is never emitted by discord.js 14.26.4; the re-login path is dead code and the `shardDisconnect` log inverts the event's meaning.**
`grep -rn invalidated node_modules/discord.js/src/` finds only the enum entry in `util/Events.js:150` — no emit site. In v14 the gateway is owned by `@discordjs/ws` (1.2.3), whose `WebSocketShard` handles `INVALID_SESSION` itself (`dist/index.js:1021-1033`): resume if resumable, else destroy with `recover: Reconnect` and re-identify. So the ticket's premise ("discord.js does not recover an `invalidated` session") is a v13-era statement; v14 does recover it, silently. The only case where a shard stays down is `shardDisconnect`, which `WebSocketManager.js:249-257` emits *only* for `UNRECOVERABLE_CLOSE_CODES` (4004 auth failed, 4010/4011 sharding, 4012 API version, 4013/4014 intents) with the JSDoc "disconnects and will no longer reconnect". The handler at `src/lib/discord/client.ts:100-108` logs "discord.js will try to resume" — the opposite — so the one event that means "this bot is permanently deaf" is logged as benign, during exactly the incident this ticket is about. Meanwhile `client.ts:115-117` and `reloginAfterInvalidation` (`client.ts:140-172`), `DiscordBotDeps.reloginDelayMs` (`:35`), `MAX_RELOGIN_DELAY_MS` (`:38`), and `recordDiscordRelogin` are unreachable in production; `gateway.test.ts` "re-logs-in with a fresh client" passes only because the fake emits an event the real client never will. The PR body, `AGENTS.md:108` ("re-logging-in with a fresh client on `invalidated`, which discord.js does not recover") and `docs/runbook.md:325` ("an `invalidated` session re-logs-in by itself") record this as fact.
*Fix:* (a) correct the `shardDisconnect` handler — it is the terminal signal in v14; `console.error` with the close code's name (`GatewayCloseCodes[code]` from discord-api-types) and say the shard will not reconnect; (b) either remove the `invalidated`/re-login machinery, or re-point it at unrecoverable `shardDisconnect` — noting that re-login with the same token and intents cannot fix 4004/4013/4014, so a loud log (the health card will follow within 5 min of the next send, since the REST token survives) is the honest option; (c) fix the three prose claims and the code comments in `client.ts` (`:16-25`, `:96-99`, `:132-139`) and `gateway-health.ts`. If you keep the re-login path for a reason I have missed, say what emits the event.

**2. [Medium] A 👍 that fails now produces a false "Nothing was delivered to the agent — reply again" trace.**
`handleReply` inserts the row (`client.ts:392`) and then awaits `message.react("👍")` (`:393`) with no guard. `insertDiscordAnswer` is a synchronous SQLite insert and almost never throws, so the most likely throw in `handleReply` is the REST reaction — missing `ADD_REACTIONS`, a 429, a network blip. That throw reaches `reportDiscordHandlerFailure`, which puts ⚠️ on the message and writes the system message at `client.ts:211` telling the human nothing was delivered and to reply again — when the row exists and the queue will deliver it within 2 s. A human who follows the instruction sends a second Discord message with a new id, which is inserted as a second answer: two turns for one question. This is the reverse of the ticket's problem — a trace that misleads rather than one that is missing.
*Fix:* make the 👍 in `handleReply` best-effort (own try/catch + log, as the REST path already does at `blocked-replies.ts:236-242`), and/or have `reportDiscordHandlerFailure` check `existingAnswers(task.id)` for `message.id` before claiming non-delivery — the helper is already imported. Add the test: handler throws after the insert → no false system message, one row.

**3. [Low] The text fallback over-skips.** `blocked-replies.ts:73,82` — `knownTexts` includes rows that *have* a Discord id. A task blocked twice whose owner gives the same short answer to both questions ("yes", "option B"), with the gateway missing the second, never has the second adopted; the run stays blocked with no card distinguishing it. The fallback was written for pre-upgrade rows, so restrict it to rows with `discordMessageId == null` (or scope `existingAnswers` to rows created after the question — the snowflake carries the timestamp). Suggestion-level; the deliberate side of the fallback (a manual web-UI re-post suppressing a late REST adoption) is worth keeping and should stay covered by a test.

**4. [Low] Write-only state (possible Speculative Generality).** `gateway-health.ts:38,70` — `lastDisconnect` is stored by `recordDiscordDisconnected` and read by nothing; `relogins` (`:40,74`) is read only by the dead re-login log line. Drop them, or surface `lastDisconnect` (code + when) on the `discord deaf` card where it would actually help an operator.

**5. [Info] Fetch window.** `blocked-replies.ts:173,207` — `fetch({ after, limit: 100 })` returns the 100 messages immediately after the question. A run blocked across a busy day in the fleet channel with more than 100 intervening messages misses its reply permanently and silently. The comment acknowledges the bound; noting it so it is a known limit rather than a surprise. Not blocking.

**6. [Info] Only matters if finding 1 keeps the re-login path:** between `stale.destroy()` (which nulls the REST token, `Client.js:250-257`) and the new session's `clientReady`, `getBotClient()` still returns the destroyed client, so outbound fails in that window. Harmless as written (dead path), but a reason to prefer deleting it.

## Verified and fine

- Event argument shapes: `shardReady(shardId, Set<Snowflake>|undefined)`, `shardResume(shardId, replayedEvents)`, `shardDisconnect(CloseEvent{code, reason}, shardId)`, `shardError(Error, shardId)`, `error(Error)`. `CloseEvent.reason` is deprecated and always empty in v14 — the `event.reason ? ... : ""` guard handles it.
- Echo premise: MESSAGE_CREATE is dispatched for the bot's own messages under `GuildMessages` (the `author.bot` early return in `handleMessage` exists for exactly that reason), and `recordDiscordInbound()` runs before that return.
- `messages.fetch({ after, limit })` returns `Collection<Snowflake, Message>` (typings 5221); `.values()` is correct and ordering is irrelevant because the selector sorts by `createdTimestamp`. `FetchedMessage`/`FailedMessage` are structurally compatible with `Message` (`reference.messageId: Snowflake|undefined`, `author.bot`).
- Content JSON: every reader takes `text` and ignores other keys — `turn-manager.ts:1642`, `chat/chat-view.ts:84-103`, `orchestrator/init.ts:373-378`; no component parses `content` directly. The `role: "system"` failure trace is not picked up as a prompt (queue filters `role='user'`).
- Sweep safety: per-row try/catch, `raceWithTimeout` on channel fetch, message fetch and react, `.catch` at the call site, and `raceWithTimeout` swallows an abandoned promise's later rejection.
- Health false positives: cannot fire with no connected bot, nothing sent, or inbound ≥ outbound; `recordDiscordConnected` resets inbound to connect time so pre-session sends are forgiven; `shardReady` on an internal re-identify also stamps inbound; outbound is recorded only on a successful `channel.send` in `sendWithRetry` (reactions and `message.reply` are not recorded — the safe direction). globalThis via `processSingleton` is right, since `POST /api/tasks/[id]/complete` sends from the route graph. Residual: a guild that is gateway-unavailable while REST still accepts sends would fire — arguably a true positive.
- Scope: nothing beyond the ticket; runbook §4, env table and `AGENTS.md` updated.

## What I ran

- `npx vitest run --maxWorkers=2` (fresh `local.db`): 123 files, 2316 tests, all passing, 80 s.
- `npx tsc --noEmit`: clean.
- `npx eslint` on the 14 changed source/test files: clean.
- `gh pr checks 248`: no checks reported on the branch.

lennons301:
Addressed, with the library checked rather than the ticket's wording trusted:

- **Finding 1 (high)** — confirmed in `node_modules/discord.js/src/client/websocket/WebSocketManager.js`: `Events.Invalidated` is never emitted, and `shardDisconnect` is emitted only when the close code is in `UNRECOVERABLE_CLOSE_CODES` ("will no longer reconnect"); recoverable closes emit `shardReconnecting` and `@discordjs/ws` resumes or re-identifies itself. So `shardDisconnect` is now the terminal signal that destroys the client and logs in a fresh one, logged with the close code's name; `shardReconnecting` is logged and left alone. `invalidated` remains wired to the same path, labelled as not emitted by this version, for a future one that does. Tests drive `shardDisconnect` (4004) → re-login, `shardReconnecting` → no re-login, and `invalidated` → same path. AGENTS.md, the runbook and the PR body are corrected.
- **Finding 2 (medium)** — the 👍 in `handleReply` is best-effort, and `reportDiscordHandlerFailure` looks the row up by Discord message id: if it exists the trace says "recorded and will be delivered, no need to reply again"; only a throw before the insert says "nothing was delivered". Two tests: a rejecting 👍 leaves exactly one row and no ⚠️; a failure after the row exists produces the "recorded" wording.
- **Finding 3 (low)** — the text fallback now applies only to rows with no id; test added for a same-worded answer to a later question.
- **Finding 4 (low)** — `lastDisconnect` removed; `relogins` is read (the re-login count in the log line).
- **Finding 5** — acknowledged; the `limit: 100` window is a known bound, documented at the constant.

lennons301 (commented):
**Verdict: request-changes** (re-review of bbdefff — narrower than the first pass: one finding, on the re-pointed re-login)

Findings 2, 3 and 4 from the first review are fixed, with tests. Finding 1's *diagnosis* is now correct everywhere — `shardDisconnect` is logged as terminal with the close code's name, `shardReconnecting` is logged and left to the library, `invalidated` is labelled as not emitted by 14.26.4, and AGENTS.md, the runbook and the PR body say so. But the *action* now attached to `shardDisconnect` — re-login with a fresh client — turns the one condition it handles into a self-amplifying reconnect storm. Details and evidence below. Merge state unchanged: not armed, no labels, gated paths.

## Previous findings

1. **`invalidated` / `shardDisconnect`** — diagnosis fixed (`client.ts:29-40, 95-138`, `AGENTS.md:108`, `runbook.md:325-326`, PR body). Action introduces a new defect: see finding A.
2. **False "nothing was delivered" trace** — fixed. `handleReply`'s 👍 is best-effort (`src/lib/discord/client.ts:431-438`); `reportDiscordHandlerFailure` looks the row up by Discord message id and words the system message accordingly (`client.ts:235-246`). Tests added: "says the reply was recorded when its row exists" and "a failed 👍 after the row is written is not a handler failure" — the second drives the real `handleReply` through `messageCreate` with a throwing `react` and asserts one row, one 👍 attempt, no ⚠️.
3. **Text fallback over-skip** — fixed. `knownTexts` is built from id-less rows only (`src/lib/discord/blocked-replies.ts:77-79`); test "lets a same-worded reply to a later question through when the earlier row carries its id" added; the pre-upgrade-row case remains covered by "does not re-deliver a reply the gateway already delivered".
4. **Write-only `lastDisconnect`** — removed. `relogins` stays and is now read by a live log line.
5/6. Informational; unchanged.

## Finding A — [High] Re-login on `shardDisconnect` multiplies under the exact condition it handles

**What the library does.** In `@discordjs/ws` 1.2.3 (`node_modules/.pnpm/@discordjs+ws@1.2.3/.../dist/index.js`), `WebSocketShard.onClose` (lines 1075-1160) handles every one of discord.js's `UNRECOVERABLE_CLOSE_CODES` (4004, 4010, 4011, 4012, 4013, 4014) the same way, in this order: `emit("closed", { code })`, then `emit("error", { error })`, then `destroy({ code })` with no recovery. The initial `connect()` (lines 634-652) awaits `Promise.race([once(shard, "ready"), once(shard, "resumed")])`, and Node's `events.once` rejects when the emitter emits `error` — so the fresh client's `login()` rejects (discord.js `Client.login`, `Client.js:228-234`, catches, destroys, rethrows). Meanwhile the `closed` emission has already reached discord.js's `WebSocketManager` `Closed` handler (`WebSocketManager.js:246-258`), which emits `shardDisconnect` on the client.

**What that does to `reloginWithFreshClient`.** Both things happen for one failure. `connect()` sets `client = c` *before* `await c.login(token)` (`client.ts:76-81`), so when the fresh client emits `shardDisconnect`, its handler's `client !== stale` check passes and a *second* `reloginWithFreshClient` loop starts; the first loop's `await connect(token)` then rejects and it *also* backs off and retries. Every failed attempt therefore spawns one more concurrent loop and keeps its parent. New loops start at `reloginDelayMs` (5 s) while parents back off, so the number of IDENTIFYs per minute grows roughly geometrically until the 5-minute cap flattens it — with N loops each attempting every ≤5 min and each attempt spawning another. It also starts from boot: `startDiscordBot()` rejecting on such a code now leaves a loop behind where before it was terminal.

**Why it matters.** The one realistic member of the set is 4014 — the Message Content privileged intent toggled off in the developer portal (a revoked token fails earlier, at `GET /gateway/bot` with 401 → `TokenInvalid`, before any IDENTIFY, and stays single-loop with clean backoff). Each attempt is a real IDENTIFY; Discord's limit is 1000 per 24 h and exceeding it resets the bot token. So an intents misconfiguration that today leaves the bot deaf and loud in the logs would, with this change, get the token reset within the hour — a strictly worse outage that needs a Doppler change to recover from. And none of these codes can be fixed by re-logging in with the same token and intents, so the loop cannot succeed even in principle.

**Missing test.** `gateway.test.ts` "re-logs-in with a fresh client when discord.js gives the shard up" has the fresh client's `login` resolve. There is no case where the fresh client itself emits `shardDisconnect` or its `login` rejects during the loop — the case the path exists for.

**Fix — recommended: remove the re-login.** Every code that produces `shardDisconnect` in 14.26.4 is a configuration error, and the library leaves the REST token intact on an unrecoverable close (the `Closed` handler sets shard status and returns; it does not destroy the client), so outbound keeps working and the deaf card + ping already cover this case within 5 min of the next send. Keep the error-level log with `closeCodeName` (as now); delete `reloginWithFreshClient`, `DiscordBotDeps.reloginDelayMs`, `MAX_RELOGIN_DELAY_MS`, `recordDiscordRelogin`/`relogins`; keep `invalidated` as a logged no-op or drop it; and record in AGENTS.md and the runbook that recovery from an unrecoverable close is "fix the configuration, restart". The ticket's "`invalidated` re-logs-in" rested on a v13 premise; the honest completion is the documented fact, not an equivalent that amplifies.

**If a human wants the path kept**, it needs all of: single-flight (one module-level in-progress promise; a `shardDisconnect` while a re-login is pending is ignored whichever client emits it), one trigger only (either the `shardDisconnect` re-entry or the `login` rejection retry — not both for the same failure), a bound (attempts or total time), and a test in which the fresh client emits `shardDisconnect` and/or its `login` rejects, asserting exactly one loop and a bounded number of clients created. I would still recommend deletion.

## Finding B — [Low, only if the path is kept]

Our own `stale.destroy()` on a still-open connection makes the shard emit `closed` with code 1000 (`index.js:747-756`), which discord.js's `Closed` handler — no `destroyed` guard — turns into `shardReconnecting` on the stale client. The log would read "recoverable close; discord.js resumes or re-identifies by itself" immediately after "re-logging in". Moot if finding A is resolved by deletion.

## Info (unchanged; outside this ticket)

The deaf card is echo-based, so it cannot fire when outbound is dead too (a revoked token): only the logs say so. That is a different failure from the ticket's inbound-deaf-while-outbound-works, noted for the record.

## What I ran

- `npx vitest run --maxWorkers=2` (fresh `local.db`): 123 files, 2321 tests, all passing, 118 s.
- `npx tsc --noEmit`: clean.
- `npx eslint` on the five re-changed source/test files: clean.

lennons301:
Finding A taken as recommended: the re-login path is gone.

- `shardDisconnect` now **records** the terminal close (`recordDiscordGatewayClosed(code)`) and logs it at error level with the code's name and the remedy ("fix the bot's configuration — token, intents — and restart"). No client is destroyed, nothing logs in again, so there is no loop to amplify. `invalidated` is logged only.
- So the deaf card covers the case *immediately*, not after a missed echo: `discordInboundStale` gained a `cause` — `closed` (fires on the next sweep with no threshold, carries `closeCode`; the card and the ping name the code and say config + restart) and `silent` (the echo rule, unchanged; "a restart re-identifies it"). A quiet fleet therefore hears about a 4014 within one sweep even though it has posted nothing.
- Tests: `shardDisconnect` (4014) → close recorded, no second client, one `login`, error log names `DisallowedIntents`, a fresh `clientReady` clears it; `shardReconnecting` → nothing; `invalidated` → logged, no client; `health.test.ts` → the `closed` cause fires with no threshold. Full suite green after the change (2321 tests before the last commit; re-running now).
- PR body, AGENTS.md and the runbook now state the deviation from the ticket's "`invalidated` re-logs-in" as a decision on evidence, not as done.

lennons301 (commented):
**Verdict: approve** (re-review of aad1206)

Finding A is resolved the way I recommended: the re-login machinery is gone, an unrecoverable close is recorded and raised as a card at once with the code and the remedy named, and the deviation from the ticket's "`invalidated` re-logs-in" is stated as a decision on evidence in the PR body, `AGENTS.md` and the runbook. I would be comfortable with this on `main`. The PR touches gated paths (`src/lib/config.ts`, `src/lib/orchestrator/autonomy/**`, `src/components/**`) and is not armed, so a human merges — one thing for that human to ratify consciously is called out below.

## For the merging human

The ticket's first acceptance criterion says `invalidated` re-logs-in. This PR deliberately does not do that, and it is right not to: discord.js 14.26.4 never emits `invalidated` (`@discordjs/ws` 1.2.3 resumes or re-identifies an INVALID_SESSION itself), and the event it *does* emit for a dead shard, `shardDisconnect`, fires only for configuration errors (4004, 4010-4014) that a re-login with the same token and intents can only repeat — and, as the previous round showed, a re-login loop there is a self-amplifying IDENTIFY storm toward Discord's daily limit, which resets the token. The intent of the criterion — never sit deaf silently — is met by the new `closed` cause of the deaf card instead. The PR body's "Deviation from the ticket" paragraph, `AGENTS.md:108` ("Nothing re-logs-in, deliberately") and `docs/runbook.md:321-330` all record this. Approving means agreeing that is the right reading of the ticket; I think it is.

## What I verified in aad1206

- **Re-login deleted.** `connect()`, `reloginWithFreshClient`, the module-level `client`, `DiscordBotDeps.reloginDelayMs`, `MAX_RELOGIN_DELAY_MS`, `recordDiscordRelogin`/`relogins` are all gone (`src/lib/discord/client.ts`, `gateway-health.ts`). Nothing destroys a client or logs in twice; `startDiscordBot` is one `createClient` + `wireGateway` + `login`. tsc clean confirms no dangling references.
- **`shardDisconnect`** (`client.ts:109-121`) records `recordDiscordGatewayClosed(event.code)` and logs at error level with `GatewayCloseCodes[code]` and the remedy (fix token/intents, restart). It cannot false-positive: discord.js emits it only from the `UNRECOVERABLE_CLOSE_CODES` branch (`WebSocketManager.js:249-257`), and nothing in this codebase destroys the client any more (a self-destroy would surface as code 1000 → `shardReconnecting`, not `shardDisconnect`).
- **`invalidated`** (`client.ts:129-137`) logged only, with an honest comment. **`shardReconnecting`** logged and left to the library.
- **`closed` cause** (`src/lib/fleet/health.ts:480-489`): fires on the next sweep with no threshold while `closedSinceMs != null`; takes precedence over `silent`; shares the announce-once flag so it pings once per spell; cleared only by `recordDiscordConnected` (a fresh `clientReady`, i.e. after the restart the card asks for). Card (`fleet-view.ts:1355-1373`), ping (`notifications.ts:410-425`) and the sweep log (`sweep.ts:371-379`) all word the two causes differently and name the close code for `closed`. Since the `closed` card is state-driven rather than echo-driven, it also covers the case I noted last round where outbound dies with inbound (4004): the dashboard card shows even if the ping cannot be sent.
- **Tests.** `gateway.test.ts`: `shardDisconnect(4014)` → observation carries `closedSinceMs`/`closeCode: 4014`, no `destroy`, one client, one `login`, error log names `DisallowedIntents`, a subsequent `clientReady` clears it; `shardReconnecting` → nothing; `invalidated` → logged, one client. `health.test.ts`: `closed` fires at once on a quiet fleet (no outbound) with the exact signal shape and an announce. The `silent` table is unchanged and still passes with the widened observation type.
- **Docs.** `AGENTS.md`, `docs/runbook.md` §4 and the PR body are consistent with the code and with each other; the runbook names both causes and their remedies.

## Remaining, none blocking

- **[Low, suggestion]** `observeDiscordGateway()` (`gateway-health.ts:78-88`) returns null until the first `clientReady`, so a boot-time unrecoverable close (bad intents from the start) records `closedSinceMs` but the `closed` card never shows — only the boot log says so. Returning the observation whenever `closedSinceMs != null`, ready or not, would make it visible on the dashboard at no false-positive cost (`shardDisconnect` is always a true terminal). Outside the ticket's failure mode (a session that goes deaf after connecting), so a follow-up if wanted.
- **[Info]** A `silent` spell that then becomes `closed` updates the card body but does not re-ping (shared announce flag). Reasonable — one ping per spell — noting it so it is a choice, not a surprise.
- **[Info]** `health.test.ts` does not separately assert `closed` dedup across sweeps or its precedence over `silent`; both ride the same code path as the tested `silent` dedup, so low value.
- Earlier informational items stand unchanged: the 100-message fetch window in `blocked-replies.ts:173`, and the `silent` cause being echo-based.

## What I ran

- `npx vitest run --maxWorkers=2` (fresh `local.db`): 123 files passed (123), 2322 tests passed (2322), 119 s, exit 0.
- `npx tsc --noEmit`: clean.
- `npx eslint` on all 14 changed source/test files across the three commits: clean.
- `gh pr checks 248`: no checks reported on the branch.

Issue #160, opened 2026-08-19
Live preview is broken for interactive sessions: dev server detected late, wrongly, or not at all


lennons301:
The original interactive use case — chat to an agent and watch its dev server live in the preview pane ("lovable-style") — no longer produces a working preview. Three independent defects in the dev-server detection path compound; any one of them is enough to leave the pane empty or pointed at nothing.

Diagnosed against two real production sessions on 2026-08-19.

## Evidence

**Session `01M0CVY2551TSV3SVC3GJW4DDY`** (LPS, "Start the dev server - want to have a look at the UI preview page"):

| time | event |
|---|---|
| 11:21:10 | `next dev` up, `Ready in 608ms`, bound to all interfaces (`Network: http://172.19.0.4:3000`); agent curls all seven `/preview` routes → 200 |
| 11:39:47 | owner: *"are you still alive? if so can you take down the dev server"* — agent curls `localhost:3000` → `000`. The server is dead. |
| 11:39:56 | agent restarts it with `setsid nohup node_modules/.bin/next dev … & disown` → survives |
| **11:42:00** | **`Dev server detected on port 3000`** — 21 minutes after the server first came up, and only after the turn ended |
| 11:43:14 | `Dev server on port 44195 stopped` |

**Session `01M0CYWVYXFN3GBY9QXVBBY4RJ`** (lemons, titled `hi`). The agent's entire output was *"Hi! Ready to help with the Lemons codebase. What would you like to work on?"*. No dev server was ever started. The feed nevertheless says:

```
Dev server detected on port 36159
Dev server on port 36159 stopped
```

## Defect 1 — `devPort` is only ever sampled at turn boundaries

`scanForDevServer` runs after the initial turn (`src/lib/orchestrator/turn-manager.ts:414`), after each follow-up turn (`:751`), and on a 30s poll gated to idle tasks (`src/lib/orchestrator/queue.ts:265-276`, `if (entry.state !== "idle") continue`).

A lovable-style session is *one long turn*: the agent starts the server a minute in and then works for twenty. For that entire window `devPort` is null and the pane renders "No dev server running" — the 21-minute gap above.

**Fix:** scan on a short interval (~10s) while an interactive task's turn is *running*, not only when idle. The `isParked` exemption and the autonomous-pass exemption both stay — this is interactive-only.

## Defect 2 — the scanner reports non-dev ports as dev servers

[...]

PR #250, opened 2026-09-07
Preview scanner: read the probe's answer off a non-TTY exec's exit code (#160 live follow-up)

lennons301:
Hotfix for #249's scanner, from the first live run (lemons, 2026-09-07 08:31–08:45 UTC, task `01M1XFWPSTGAG2EFYEQ679AEPH`). The live half of #160 is not done until this is deployed.

## What happened on prod

- `next-server` was listening on `0.0.0.0:3000`; by hand inside the container all three gates passed (`ss` row routable and plausible, `curl http://172.19.0.4:3000/` → 307).
- `docker events` showed the queue scanning every 10s exactly as designed: an `ss -tlnp` exec, then the probe exec, in pairs.
- Docker's records of every probe exec showed the **match exit**: the script found 3000 every time. The orchestrator parsed nothing from it, `devPort` was never written, and the pane sat on "No dev server yet".

## Root cause (replicated on the VPS against the agent image)

The exec was created with `Tty: true` but started with `exec.start({})`. The daemon frames an exec's output from the **start** request's `Tty`, not the create request's, so every scan's stream came back **multiplexed** — an 8-byte stdout/stderr frame header per frame — and was read as plain text. What the orchestrator received for the probe, bytes escaped:

```
\x01\x00\x00\x00\x00\x00\x00\x063000\r\n
```

That line does not parse as a number. `ss` survived the same framing because the header landed on the column-heading row the parser ignores. The original Phase 2c scanner had the identical latent bug; its tolerance of the header row is what hid it.

## Fix

- Both scanner execs run **without a TTY** and are demultiplexed with `modem.demuxStream`, the pattern `execAgentTurn` already uses. Replicated output: `3000\n`, exit 100.
- The probe reports its match on the **exit code** — `PROBE_EXIT_BASE + candidate index` (base 100, clear of bash/curl's own codes) — which has no framing to fall foul of. stdout is only the fallback when the daemon reports no code, with control bytes stripped; the exact production bytes are pinned in a test.
- Exec exit is awaited by stream end or an inspect poll, bounded at 10s, as the container manager's file reads already do.

## Worth a look separately
`readContainerFile` / `writeContainerFile` in `container-manager.ts` use the same `Tty: true` create + `start({})` shape and read the stream as text. They may be getting away with it because Node's base64 decoder skips non-alphabet bytes. Not touched here.

## Verification
- Full suite on this branch: 134 files / 2408 tests pass. `tsc --noEmit` and `eslint` clean on the changed files.
- Old-vs-new capture replicated on the VPS against a throwaway `interlude-agent-claude-code` container with a server on `0.0.0.0:3000` (comment below). No live session was touched; the throwaway container was removed.


lennons301:
Replication on the VPS (dockerode 4.0.9 against the live daemon; target = a throwaway `interlude-agent-claude-code` container running `http.createServer(...).listen(3000, "0.0.0.0")` on the `interlude` network). Same probe script, four runs each. Bytes shown escaped.

**Shipped path** — exec created `Tty: true`, `exec.start({})`, raw stream read as text:
```
OLD tty probe: {"out":"\x01\x00\x00\x00\x00\x00\x00\x063000\r\n","exit":100}   ×4
OLD tty ss:    {"len":275,"head":"\x01\x00\x00\x00\x00\x00\x01\x0bState  Recv-Q Send-Q Local Address:Port ..."}
```
The 8-byte multiplex header (`01 00 00 00 | 00 00 00 06` = stdout, 6 bytes) is glued to the front of the only line, so `parseInt` sees a control byte first. On `ss` it lands on the heading row, which the parser skips — hence `ss` "worked".

**This PR** — exec created `Tty: false`, `demuxStream`, exit code via inspect:
```
NEW notty probe: {"out":"3000\n","exit":100}   ×4
NEW notty ss:    {"len":264,"head":"State  Recv-Q Send-Q Local Address:Port  Peer Address:PortProcess ..."}
```
Throwaway container removed afterwards; no session was touched.

